PR #2, opened 2016-02-25
Added XmlWriter and Element::create API

matt2xu:
This PR adds the possibility to transform (add/change elements) and write XML. Two changes:

  1. add a XmlWriter structure, which just wraps a Write at the moment but could (should?) be extended in the future to support pretty-printing or automatic shortening of empty tags.
  2. add a "create" method to Element, to create an element with a name/content, and possibly attributes (should only be used for start elements, but at the moment this is not enforced). For elements with no attributes, use the iter::empty() function.

I added one test for the writer, and one test for the writer + element create. Also, note that I made the XmlWriter use a Write trait rather than BufWrite because it doesn't need its additional methods at the moment, but of course for best performance one should use a writer implementing BufWrite.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,8 +5,8 @@
 //! ```
 //! use quick_xml::{XmlReader, Event};
 //! 
-//! let xml = r#"<tag1 att1 = \"test\">
-//!                 <tag2><!--Test comment-->Test</b>
+//! let xml = r#"<tag1 att1 = "test">
+//!                 <tag2><!--Test comment-->Test</tag2>
 //!                 <tag2>
 //!                     Test 2
 //!                 </tag2>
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -303,6 +303,25 @@ impl Element {
         Attributes::new(&self.buf[self.start..self.end], self.name_end)
     }
 
+    /// Creates a new Element from the given name and attributes.
+    /// attributes are represented as an iterator over (key, value) tuples where
+    /// key and value are both &str.
+    pub fn create<'a, I: Iterator<Item = (&'a str, &'a str)>>(name: &str, attributes: I) -> Element {
```
tafia: Few remarks:
- you can use `new` instead of `create` and rename the existing (private) `new` function into `from_buffer` for instance (opened for other suggestion)
- `(&'a str, &'a str)` is not symmetric with `Attributes::Item = (&[u8], &str)`. I wanted to avoid utf8 conversion when possible because most of the time we (I ?) compare it with statically known names, thus I could always use `key == b"mykey"`. Having symmetric behavior could help creating Elements based on other Elements (in your test, you had to use `from_utf8`), I think it is a common scenario
  - in your example, we could use `Vec<u8>` and push bytes directly
  - an alternative is to have 2 `Attributes`-like iterators, one pure `&[u8]`, another one `&str`

src/lib.rs
```diff
@@ -395,3 +414,53 @@ fn read_until<R: BufRead>(r: &mut R, byte: u8, buf: &mut Vec<u8>) -> Result<usiz
...
+            Event::Start(e) => self.write_start_tag(e),
+            Event::End(ref e) => self.write_wrapped_str(b"</", e, b">"),
+            Event::Text(ref e) => self.write_bytes(e.as_bytes()),
+            Event::Comment(ref e) => self.write_wrapped_str(b"<!--", e, b"-->"),
+            Event::CData(ref e) => self.write_wrapped_str(b"<![CDATA[", e, b"]]>"),
+            Event::Header(ref e) => self.write_wrapped_str(b"<?", e, b"?>"),
+        }
+    }
+
+	#[inline]
```
tafia: alignment typo

src/lib.rs
```diff
@@ -395,3 +414,53 @@ fn read_until<R: BufRead>(r: &mut R, byte: u8, buf: &mut Vec<u8>) -> Result<usiz
...
+            writer: inner
+        }
+    }
+
+    /// Consumes this Xml Writer, returning the underlying writer.
+    pub fn into_inner(self) -> W { self.writer }
+
+    pub fn write(&mut self, event: Event) -> Result<()> {
+        match event {
+            Event::Start(e) => self.write_start_tag(e),
```
tafia: Why not just using `self.write_wrapped_str(b"<", element, b">")`?

src/test.rs
```diff
@@ -125,3 +126,42 @@ fn test_nested() {
...
+#[test]
+fn test_writer() {
+    let str = r#"<?xml version="1.0" encoding="utf-8"?><manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.github.sample" android:versionName="Lollipop" android:versionCode="5.1"><application android:label="SampleApplication"></application></manifest>"#;
+    let reader = XmlReader::from_str(&str).trim_text(true);
+    let mut writer = XmlWriter::new(Cursor::new(Vec::new()));
+    for event in reader {
+        assert!(writer.write(event.unwrap()).is_ok());
+    }
+
+	let result = writer.into_inner().into_inner();
```
tafia: alignment typo

src/lib.rs
```diff
@@ -395,3 +414,53 @@ fn read_until<R: BufRead>(r: &mut R, byte: u8, buf: &mut Vec<u8>) -> Result<usiz
...
+    pub fn new(inner: W) -> XmlWriter<W> {
+        XmlWriter {
+            writer: inner
+        }
+    }
+
+    /// Consumes this Xml Writer, returning the underlying writer.
+    pub fn into_inner(self) -> W { self.writer }
+
+    pub fn write(&mut self, event: Event) -> Result<()> {
```
tafia: Can you add a comment for the pub function (even simple)?

tafia:
That's awesome!
I've done some comments but it looks really nice, thank you very much!

tafia:
Also might be good to update the readme :+1:

matt2xu:
Thanks for the review! I have made some changes to address these points: typos, comment on the write function, new instead of create.

Rather than change the README (not testable), I have added an example in the lib.rs file, and also taken the liberty to improve the existing example and fix the XML. Ideally, the documentation could be hosted on Github pages from the output of rustdoc, to make sure the examples are always up to date and compile.

Regarding the signature of Iterator<&str, &str> I know it's not symmetric, it just feels more natural in my opinion :-) I find it a bit low-level to do comparisons with byte string literals, but I understand that you want to avoid unnecessary UTF-8 conversion. In the case of the element's name, you have a choice: you can use either as_bytes() or as_str(). But for attributes, since the type is fixed on the Iterator implementation, it's a bit more difficult. At the same time, is that really a bottleneck on attributes? Generally, you match on element names first, so I assume this is where a program would typically spend most of its time checking.

Alternatively, for the Element::new method, maybe I could use AsRef<[u8]> to support both (&[u8], &str) and (&str, &str); I haven't tried, but it seems doable. Then we would implement two iterator methods on Attributes to support both cases.

Just let me know what your preference is :-)

tafia:
AsRef looks good to me!

matt2xu:
Works with AsRef, which I think is the way to go: Into would move its argument, while we only need to borrow it as a reference to a byte slice. In the future (in another contribution) I say we should try to improve Attributes to have the same flexibility as Element regarding the as_bytes()/as_str() duality. But for now it's good as it is I think :-)

tafia:
Thanks a lot!!

Issue #3, opened 2016-02-26
Improve API for creating new elements


matt2xu:
Instead of Element::new(name, attributes), we could do separate the text and attributes part:

  1. Element::new(text) // works for any kind of element: text, comment, start, end...
  2. element.add_attributes(iter) // only useful for start elements

This way, no need to pass an empty iterator for elements where attributes don't make sense (text or comment).

I would also change the Element::new to use an AsRef<[u8]> for the name, to allow faster/easier creation of elements from existing ones.

tafia:
Looks like a good idea.

For the attributes, I'd prefer `extend_attributes(iter)` to consume an iterator (a-la `Vec`). 
A `push_attribute` could be useful too (could be internally used by the extend version) because one does not always have an iterator at hand.

matt2xu:
:+1: I'll look into it.

tafia:
closed by #5

PR #5, opened 2016-02-26
improve Element API with new, push_attribute, extend_from_attributes

matt2xu:
Implemented #3, now using IntoIterator so we don't need to call .into_iter() on collections, and permitted chained calls (the two additional methods return &mut Element)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,14 +48,15 @@
 //!             // collect existing attributes
 //!             let mut attrs = e.attributes().map(|attr| attr.unwrap()).collect::<Vec<_>>();
 //!
-//!             // adds a new my-key="some value" attribute
-//!             attrs.push((b"my-key", "some value"));
+//!             // copy existing attributes, adds a new my-key="some value" attribute
+//!             let mut elem = Element::new("my_elem").with_attributes(attrs);
+//!             elem.push_attribute(b"my-key", "some value");
 //!
 //!             // writes the event to the writer
-//!             assert!(writer.write(Start(Element::new("my_elem", attrs.into_iter()))).is_ok());
+//!             assert!(writer.write(Start(elem)).is_ok());
 //!         },
 //!         Ok(Event::End(ref e)) if e.as_bytes() == b"this_tag" => {
-//!             assert!(writer.write(End(Element::new("my_elem", iter::empty::<(&str, &str)>()))).is_ok());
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -456,12 +445,42 @@ impl Element {
...
+        where K: AsRef<[u8]>, V: AsRef<[u8]>
+    {
+        {
+            let bytes = &mut self.buf;
+            bytes.push(b' ');
+            bytes.extend_from_slice(key.as_ref());
+            bytes.extend_from_slice(b"=\"");
+            bytes.extend_from_slice(value.as_ref());
+            bytes.push(b'"');
+        	self.end = bytes.len();
```
tafia: typo

tafia:
I have mixed feelings for `extend_attributes` and `push_attribute` returning `&mut Element`. 
I understand that it's helpful to chain things together but at the same time it looks too different to what `Vec` is doing.
For chaining, I think what is expected is more a `with_attributes(self, iter)` which takes ownership of self in the meantime and returns it.

```rust
// chaining function ... initialization
impl Element {
    fn with_attributes(self, iter) -> Element;
    fn push_attribute(&mut self, att: (...));
    fn extend_attributes(&mut self, iter: ...);
}
```

src/test.rs
```diff
@@ -153,10 +154,12 @@ fn test_write_attrs() {
                     let (k,v) = e.unwrap();
                     (from_utf8(k).unwrap(), v)
```
tafia: We don't need `from_utf8` anymore here

matt2xu:
The spurious tabs are killing me, I'm using Eclipse and despite being configured for spaces only it still inserts tabs in certain cases. What editor do you recommend for coding in Rust?

Agreed for from_utf8. :+1: for with_attributes consuming self, I like that. For push_attribute and extend_attributes the chaining is less useful, because in the end the element must be put in an Event, and that consumes it. So I'm OK with push_attribute not returning anything. If we have with_attributes and push_attribute, do we still need extend_attributes?

tafia:
I used to write on Atom at home and vim at work. I'm fully vim now.

I agree we can skip `extend_attributes` for the moment.

matt2xu:
I updated the PR to take your feedback into account.

tafia:
Thanks!

PR #6, opened 2016-02-28
Improve Element API and string conversion

matt2xu:
Now using &[u8] everywhere in Element and Attribute, and added an AsStr trait for converting any byte slice to an &str. Renamed Element::as_bytes() to Element::name(), and added an Element::content() method. Removed Element::as_str(), no longer needed since it is handled by the AsStr trait.

On Attribute, made the value field a &[u8], to make it consistent with the type of the key; both can be converted with a call to as_str().

Will rebase if needed when #5 is merged.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,7 +17,7 @@
 //! for r in reader {
 //!     match r {
 //!         Ok(Event::Start(ref e)) => {
-//!             match e.as_bytes() {
+//!             match e.name() {
 //!                 b"tag1" => println!("attributes values: {:?}", 
 //!                                  e.attributes().map(|a| a.unwrap().1).collect::<Vec<_>>()),
 //!                 b"tag2" => count += 1,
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -44,18 +44,19 @@
 //! let mut writer = XmlWriter::new(Cursor::new(Vec::new()));
 //! for r in reader {
 //!     match r {
-//!         Ok(Event::Start(ref e)) if e.as_bytes() == b"this_tag" => {
+//!         Ok(Event::Start(ref e)) if e.name().as_str().unwrap() == "this_tag" => {
```
tafia: for the example I'd prefer `e.name() == b"this_tag"`

tafia:
There is a conflict with the PR #5.
Can you please rebase?

tafia:
Could you also update the readme (just copy lib.rs example).
Thanks

matt2xu:
@tafia rebased, and changed the example to use the raw string. I also updated the readme, but I think the easiest way to keep the readme up to date would be to not show code there, since the examples are already included in the documentation. Personally when I look at a new crate and I want to understand how to use it, I go 1) to the documentation (when it's included) 2) to the "examples" folder in the git repository. As a last resort I start browsing the source repositoy directly :worried:

tafia:
I personally think the readme must show the most important features (like the one we put on each module documentation) because it is easier to read than doc (you can add context, paragraphs etc ...). 
The docs must on the other hand have ALL kind examples if possible
In particular when it is just one copy paste.

Thanks for the PR!

PR #22, opened 2016-03-30
Run xmlrs converted tests

### LINKED ISSUE #20: More tests
Try implementing all xml-rs' for a start ?

tafia:
Closes #20 

For the moment, still work to do

- [x] error in escaping characters
- [x] error in `<!DOCTYPE ...>` nodes
- [x] invalid `--` in comments
- [x] namespaced names
- [x] mismatch between `Event::Text` and `Event::Characters` when no trim (more Text than Character)
- [ ] issue with `>` character in attributes


tafia:
Merging even without `>` character in attribute not handled

PR #23, opened 2016-03-31
Lazily support escaped character

### LINKED ISSUE #12: Escape characters
https://www.w3.org/TR/xml/#syntax
http://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents

Add a fn on `Element` or on `AsStr`, so the escape is done on demand and there is no perf penalty when xml is known to be simple enough.

tafia:
Adds methods for converting a `&[u8]` into `Cow<[u8]>` with un-escaped characters
Closes #12

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -53,3 +53,22 @@ fn bench_quick_xml_namespaced(b: &mut Bencher) {
     });
 }
 
+#[bench]
+fn bench_quick_xml_escaped(b: &mut Bencher) {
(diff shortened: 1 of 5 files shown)


tafia:
To unescape values:
- in `Event::Text(Element)`: use `Element::unescaped_content()`
- for attribute, use `Element::unescaped_attributes()` instead of regular, faster, `Element::attributes()`

PR #32, opened 2016-07-02
Correct typo and remove trailing spaces in code.

AlbertoGP:
Just a correction of "cgaracter" to "character".

Removing the trailing spaces is there just because I have my text editor configured to do that automatically. If you prefer to leave those spaces in place I can do another pull request with only the typo, or you might find it simpler to simply correct it yourself. Just let me know what works best for you.


tafia:
Thanks!

PR #33, opened 2016-07-09
Fix attribute writing on end-element

tmoers:
Rewriting an xml document with an empty element that
has attributes also writes the attributes on the end tag.

This pull request adds a unit test to show the bug and
fixes the problem.


tafia:
Thanks!

tafia:
Published v0.2.3

PR #35, opened 2016-07-18
Add support for empty elements

tmoers:
Empty elements are represented as the combination of a start element with an
end element. This means that reading a file with empty elements:

```
  <some_tag attr='1'/>
```

would result in an empty start element, followed by an end element:

```
  <some_tag attr='1'></some_tag>
```

This merge request changes that behaviour so that a read/write roundtrip
results in an identical file.


tafia:
Great!
It looks much  better indeed!

Just one thing though, benches are not working anymore because they count for the `Start`s events which incidentally changed.

Could you send a commit to fix this or I'll do it myself later (please try to bench with features too, to be 100% sure).

Thanks also for the fixes for windows new lines (didn't know about `.gitattributes`)

tafia:
Another point, I think we need to handle the `Empty` event in [`XmlnsReader::next`](https://github.com/tafia/quick-xml/blob/master/src/namespace.rs#L93) match statement

tmoers:
Ok, I'll see when I have time to look into it further.. Thanks for the
response!

tmoers:
Ok, please recheck the benches... I don't know what you mean by `bench with features too`.. (pretty new to rust, so I'm still getting used to everything.. e..g, I had a hard time getting the benches to run.. Love rustup already though! :) )

tmoers:
Hmm..  About XmlnsReader::next... This probably also misses a test, because I didn't run into it after the changes... I'll have a look at it tomorrow..

tmoers:
I've fixed the benches and prepared `XmlnsReader::next` to make it easy to add the `Empty` case, but I'm not sure what the correct behavior should be, so if you check the rework. Could you pay special attention to changes in `namespace.rs`?

tmoers:
I've fixed the benches and prepared `XmlnsReader::next` to make it easy to add the `Empty` case, but I'm not sure what the correct behavior should be, so if you check the rework. Could you pay special attention to changes in `namespace.rs`?

tafia:
Looks great!
I just have one concern. This is a breaking change (for the better), and I'm worried about people using the library and silently not checking for `Empty` events anymore.
Could we have a property that can trigger whether we use `Empty` events or not ? Per default it won't use it.
What do you think?
Thanks a lot again for this PR, this is very nice!

tmoers:
Yes, I was afraid of the same thing. I'll have a look into that

tmoers:
Hi Tafia, Here is a preview of an opt-in empty event. What do you think. Note that I still have to change the default.. But I wanted some feedback first before changing all unit tests.

tafia:
Sorry for the late answer!
It looks great!!
Thank you very much!

tafia:
Published v0.3.0

tmoers:
Hmm.. I guess I didn't ask my message clear enough :S.. This was just a preview to get your opinion, it isn't completely what we want yet. I.,e. the default is now still set to `use empty elements`, I didn't want to update all the unit tests yet if you didn't like how it was implemented now.

I won't be able to work on this before this evening, so maybe you can revert the merge for now?

tafia:
!
I'll revert as soon as I get close to a computer.
I wrongly took the `expand_empty...` for a  `use_empty...`. My bad and
thanks for the warning.

tafia:
Just published v0.3.1 with default NOT to use empty event.
Also fixed all the tests (it is weird they all passed before ...)

tafia:
... and yanked published versions

tmoers:
great... thanks!

PR #40, opened 2016-09-20
Add Element method to get content as a String with XML entities unescaped

emosenkis:
I'm not sure that I did the error handling correctly and there's probably a more efficient way to do this but I figured that a pull request would be better than just opening an issue.


tafia:
It's been a month since you've done this PR. I am VERY sorry, i totally forgot it.
I have fixed it in #44.
Thanks a lot and again sorry.

emosenkis:
No problem. Thanks for fixing it (and writing and maintaining this library)!

PR #41, opened 2016-10-18
Encourage users to use the latest version

msiemens:
The README tells users to install v0.1.9 instead of the latest v0.4.0.


tafia:
Thanks!
I didn't see this one. You're totally right.

PR #47, opened 2016-12-19
Encodings

### LINKED ISSUE #46: Other charsets are not supported
Looks like it works with UTF-8 only.
At least I can confirm that it doesn't work with KOI8-R.

tafia:
Closes #46 

This PR adds support for decoding all formats.
The decoding itself is handled by the [encoding](https://github.com/lifthrasiir/rust-encoding) crate.

# Example

```rust
use quick_xml::{AsStr, Event, XmlReader};
use quick_xml::Event::*;

let xml = "/path/to/my/custom/encoding.xml";
let reader = XmlReader::from(xml).trim_text(true);
let mut decoder = None; // defaults to utf8
for e in reader {
    match e {
        Ok(Decl(e)) => decoder = e.encoder().expect("Cannot find 'encoding' attribute"),
        Ok(Text(e)) => println!("{}", e.content()
                                       .as_string(decoder.as_ref())
                                       .expect("Cannot decode Text node")),
        _ => (),
    }
}
```

# Pros

- Existing users just need to add `None` decoder in `into_string` methods to maintain compatibility
- Ability to use ANY decoder when the xml doesn't have a Declaration node
- Decoding being expensive (it returns an owned string), tradeoff is developer responsability, he can continue using default `as_str`
- There is no lifetime issue as the decoder is handled by the caller and not by the reader itself

# Cons

The API is not straighforward and one must handle a decoder variable in the code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quick-xml"
-version = "0.4.2"
+version = "0.5.0"
 authors = ["Johann Tuffe <tafia973@gmail.com>"]
 description = "High performance xml reader and writer"
 
@@ -13,6 +13,7 @@ license = "MIT"
 
 [dependencies]
 log = "0.3.6"
+encoding = "0.2.33"
 
 [dev-dependencies]
 xml-rs = "0.3.4"
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,7 +9,7 @@
   - chore: Changes to the build process or auxiliary tools/libraries/documentation
 
 ## master
-
+- feat: support for non utf8 charsets
 
 ## 0.4.2
 - feat: add `into_unescaped_string`
--- a/src/decoded.rs
+++ b/src/decoded.rs
@@ -1,0 +1,60 @@
+use std::io::BufRead;
+use encoding::types::EncodingRef;
+use {XmlReader, Event, AsStr};
+use error::ResultPos;
+
+/// An iterator which decodes `Event::Text`
+///
+/// Behaves exactly the same as `XmlReader` except for
+/// `Event::Text` replaced by `Event::DecodedText`, which provides a
+/// decoded `String`
+pub struct XmlDecoder<'a, B: BufRead + 'a> {
+    reader: &'a mut XmlReader<B>,
+    decoder: Option<EncodingRef>,
+}
+
+impl<'a, B: BufRead + 'a> XmlDecoder<'a, B> {
+    /// Creates a new `XmlDecoder`
+    pub fn new(r: &'a mut XmlReader<B>) -> XmlDecoder<'a, B> {
+        XmlDecoder {
+            reader: r,
+            decoder: None,
+        }
+    }
+
+    /// Sets a custom decoder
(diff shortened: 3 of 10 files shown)


Kilte:
Is it possible to read XML declaration before parsing? I'm not familiar with the code enough, but it seems to me that we can do it better.
```rust
// auto
let reader = XmlReader::from(xml); 

// manually
let mut reader = XmlReader::from(xml).trim_text(true);
reader.set_decoder(decoder);
// or
let reader = XmlReader::from(xml).with_decoder(decoder).trim_text(true);
```

tafia:
This is what I wanted to do at first but it has some impacts.

## Need to change the API

Today the decoding is done for any `&[u8]` (as it implements `AsStr`). The entire API is built with this principle in mind: use bytes as much as possible and at the very end, convert it to string *if needed*. There are many cases when you actually don't need to convert it back at all (if you're just converting the data in general, writing to another file in particular).

Currently, to simplify, here is the flow:
`XmlReader` -> `Event` -> `Element` -> `&[u8]` -> `&str/String`

Let's suppose we have a `XmlReader { ..., decoder: Option<EncodingRef> }` which is updated either with a builder fn as you proposed or when hitting a `Event::Decl`.

```rust
for event in reader { // reader is mutably borrowed here
    match event {
        Ok(Start(ref element)) => match element.name() {
            b"test" => //...,
            _ => (),
        },
        Ok(Text(ref element)) => {
            let content: &[u8] = element.content();
            let content_str = content.as_str(); // for utf8, cheap
            // we cannot do the following due to borrowing issues (reader is mutably borrowed)
            // let content_string = content.as_string(self.decoder.as_ref());
        }
        _ => (),
    }
}
```

## Decoding is expensive

While we can cheaply convert the `&[u8]` into `&str` with utf8 there is no way to do so with other encodings ... so we need to allocate a new `String` everytime (this is what the `decode_to` does in the encoding crate).

I would prefer people to opt-in if possible (there are many scenarios where you actually don't need to decode it at all).

## A possible solution

One solution, apart from the one I did in this PR is to have another iterator which iterates on something like a `DecodedEvent` so:

`XmlReader` -> `XmlDecodedReader` -> `DecodedEvent` -> `DecodedElement` -> `String` (not possible to have `&str` here).

Kilte:
Thanks for explanation. If you need to decode data, you likely won't encode it again, right? So I prefer second solution, because API is more straightforward and looks much better. 

@frewsxcv What do you think?

tafia:
I just pushed a new commit which add a `XmlDecoder` wrapper which decodes all `Event::Text` into `Event::DecodedText`.

`XmlDecoder` only has a reference to `&mut XmlReader` which gives the possibility to revert back to the simpler `XmlReader` if needed.

```rust
use quick_xml::{Event, XmlReader};

let xml = "/path/to/my/custom/encoding.xml";
# let xml = "tests/documents/opennews_all.rss";
let mut reader = XmlReader::from(xml).trim_text(true);
let reader = reader.decoded();
for e in reader {
    match e {
        Ok(Event::DecodedText(_, s)) => println!("{}", s),
        _ => (),
    }
}
```

tafia:
Closing this for #52

PR #48, opened 2017-01-12
Implement handling of default namespaces (xmlns="...")

chklauser:
For my work, I need an XML parser that fully supports namespaces, both with prefixes and as default namespaces. ([W3C Namespaces in XML](https://www.w3.org/TR/xml-names11/#scoping-defaulting) ) 

Previously only prefixed namespace declarations were supported in quick-xml. Resolving default namespaces unfortunately involves a bit
more work for the `XmlnsReader`. To offset this, I have eliminated the copy of the declaring element
name in the namespace declaration stack. The 'nesting level' of declarations is no longer updated
each time an element is encountered, instead, the reader itself maintains a current nesting level and
uses it to pop declarations that have a deeper nesting level encoded.

This patch also includes a fix for a related issue where `Empty` events resolved namespaces in the
parent scope instead of the scope the empty element represents. Popping of that scope gets deferred until the next call to `XmlnsReader::next()` so that the consumer has a chance to call `resolve` on any attributes of the `Empty` elements in the scope of that empty element.

Finally, the semantics of `XmlnsReader::resolve` 'changed', depending on how you look at it.
Previously it wasn't specified what exactly it resolves. With this patch, `resolve` explicitly
applies qualified name resolution rules for *attributes* (XML default namespace don't apply to attributes).
Qualified name resolution for *elements* already happens automatically with the `XmlnsReader`.

In my opinion, we should change the name of `resolve` to something like `resolve_attr` to make this clear, but that doesn't need to happen right now.

I have measured the performance before ("master") and after the change ("dflt") and it seems like the two implementations are about equivalent. But my measurements weren't super scientific (done on my development desktop PC in a Fedora VM) and cargo reported quite a high variance for most of the tests (between +/- 100,000ns and 200,000ns typically).
[quick_xml_dflt_ns_perf.csv.txt](https://github.com/tafia/quick-xml/files/702924/quick_xml_dflt_ns_perf.csv.txt)
[quick_xml_dflt_ns_perf.xlsx](https://github.com/tafia/quick-xml/files/702925/quick_xml_dflt_ns_perf.xlsx)


tafia:
Fantastic!!

I'm really impressed by the level of details you have put in it. I will try to review and probably merge it this we or early next week.

Thanks a lot !

src/namespace.rs
```diff
@@ -4,19 +4,64 @@ use {XmlReader, Event, Element};
...
+    fn matches_qualified_elem(&self, name: &[u8]) -> bool {
+        if let Some(ref prefix) = self.prefix {
+            let len = prefix.len();
+            name.len() > len && name[len] == b':' && &name[..len] == &prefix[..]
+        } else {
+            false
+        }
+    }
+
+    /// A namespace declaration matches unqualified elements iff it is a default namespace
```
tafia: small typo `iif`

src/namespace.rs
```diff
@@ -4,19 +4,64 @@ use {XmlReader, Event, Element};
...
+        self.prefix.is_none()
+    }
+
+    /// Check whether this namespace declaration matches the (potentially prefixed) **attribute**
+    /// name. The default namespace does not apply to attributes. The "namespace name" of an
+    /// unprefixed attribute always has no value.
+    ///
+    /// [W3C Namespaces in XML 1.1 (2006)](http://w3.org/TR/xml-names11/#scoping-defaulting)
+    #[inline]
+    fn matches_qualified_attr(&self, name: &[u8]) -> bool {
```
tafia: I may be wrong but I don't see the difference with the `match_qualified_elem` function.
I got that the unqualified is for elements only, but for qualified we could be DRY and have both function call the same `matches_qualified` function.

src/namespace.rs
```diff
@@ -73,81 +125,132 @@ impl<R: BufRead> XmlnsReader<R> {
...
     fn push_new_namespaces(&mut self, e: &Element) {
-        // adds new namespaces for attributes starting with 'xmlns:'
+        // adds new namespaces for attributes starting with 'xmlns:' and for the 'xmlns'
+        // (default namespace) attribute.
         for a in e.attributes().with_checks(false) {
             if let Ok((k, v)) = a {
-                if k.len() > 6 && &k[..6] == b"xmlns:" {
+                // Check for 'xmlns:any-prefix' and 'xmlns' at the same time:
+                if k.len() >= 5 && &k[..5] == b"xmlns" && (k.len() == 5 || k[5] == b':') {
+                    // We use an empty prefix as the 'name' for the default namespace.
```
tafia: Do you mean `None` prefix here? Empty sounds more like `vec![]`

chklauser:
Thanks for the review, I have updated the patch:
 * wrote out "IFF" as "if and only if"
 * merged the two 'matches_qualified_*' functions into one.
 * rewrote comment about representation of default namespaces with 'None' (an earlier version indeed used `vec![]`)
 * mentioned default namespaces and resolution scope for empty elements in changelog

tafia:
Thanks a lot! I'll update crates.io tomorrow

PR #49, opened 2017-01-12
Handle '>' in attribute values

### LINKED ISSUE #34: Attributes with `>` character are not properly managed
Broken tests:
https://github.com/tafia/quick-xml/blob/master/tests/xmlrs_reader_tests.rs#L45-L80

chklauser:
When looking for the end of a start/empty tag, we use an alternate
version of `read_until` that employs a small state machine to *not*
match on the end byte ('>') when we are currently in an attribute
value.

Fixes #34

I didn't re-enable the tests mentioned in the issue because they would then fail for other reasons. First because of the comment that seems to be parsed correctly but doesn't appear in the test and then because the CData section doesn't match the expected Text event. `quick-xml`s behaviour seems correct.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -386,20 +386,62 @@ impl<B: BufRead> XmlReader<B> {
     /// private function to read until '>' is found
     fn read_until_close(&mut self) -> Option<ResultPos<Event>> {
         self.tag_state = TagState::Closed;
-        let mut buf = Vec::new();
-        match read_until(&mut self.reader, b'>', &mut buf) {
-            Ok(0) => None,
-            Ok(n) => {
-                self.buf_position += n;
-                match buf[0] {
-                    b'/' => Some(self.read_end(buf)),
-                    b'!' => Some(self.read_bang(buf)),
-                    b'?' => Some(self.read_question_mark(buf)),
-                    _ => Some(self.read_start(buf)),
-                }
-            }
-            Err(e) => Some(self.error(e, 0)),
+
+        // need to read 1 character to decide whether pay special attention to attribute values
+        let start;
+        loop {
+            // Need to contain the `self.reader.fill_buf()` in a scope lexically separate from the
+            // `self.error()` call because both require `&mut self`.
+            let start_result = {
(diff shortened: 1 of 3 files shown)


chklauser:
The updated version includes a mention of the fix in the changelog.

src/lib.rs
```diff
@@ -873,6 +915,81 @@ fn read_until<R: BufRead>(r: &mut R, byte: u8, buf: &mut Vec<u8>)
...
+                Err(e) => return Err(Error::Io(e)),
+            };
+
+            let mut bytes = available.iter().enumerate();
+
+            let used: usize;
+            loop {
+                match bytes.next() {
+                    Some((i, &b)) => {
+                        if state == ElemReadState::Elem {
```
tafia: I'd prefer for readability something like
```rust
match (state, b) {
    (ElemReadState::Elem, b) if b == end_byte => { ... },
    (ElemReadState::Elem, b'\'') =>  { ... }, 
    // ...
    (ElemReadState::DoubleQ, b'\"') => { ... }
    _ => (), 
}
```

chklauser:
I have rebased the patch and changed the state machine code to
```rust
state = match (state, b) {
    (ElemReadState::Elem, b) if b == end_byte => {
        // only allowed to match `end_byte` while we are in state `Elem`
        buf.extend_from_slice(&available[..i]);
        done = true;
        used = i + 1;
        break;
    },
    (ElemReadState::Elem,  b'\'') => ElemReadState::SingleQ,
    (ElemReadState::Elem, b'\"') => ElemReadState::DoubleQ,

    // the only end_byte that gets us out of state 'SingleQ' is a single quote
    (ElemReadState::SingleQ, b'\'') => ElemReadState::Elem,

    // the only end_byte that gets us out of state 'DoubleQ' is a double quote
    (ElemReadState::DoubleQ, b'\"') => ElemReadState::Elem,

    // all other bytes: no state change
    _ => state,
};
```
That looks much nicer indeed :)

tafia:
Thanks!
Publishing v0.5.0

PR #50, opened 2017-01-21
Add a constructor to XmlDecl

chklauser:
Unless I'm overlooking something, there currently does not seem to be any way for a consumer of `quick-xml` to construct a new `XmlDecl` struct for _writing_ an XML document (I can't find a constructor and the `element` field is private). One would have to assemble the declaration as a string and then parse it using the `XmlReader`. Not exactly ideal for 'write-only' use cases.

This patch adds a constructor (`XmlDecl::new`) that assembles the `Element` from the supplied attribute values for _version_, _encoding_ and _standalone_ (the latter two are option, of course). According to the [W3C XML 1.1 spec](http://www.w3.org/TR/xml11/#sec-prolog-dtd), those are the only attributes allowed on an XML declaration

(I'm currently referencing the quick-xml crate via github in my project, so a release on crates.io is not super urgent for me)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@
   - chore: Changes to the build process or auxiliary tools/libraries/documentation
 
 ## master
+- feat: constructor for `XmlDecl`
 
 ## 0.5.0
 - feat: apply default namespaces (`xmlns="..."`) to unqualified elements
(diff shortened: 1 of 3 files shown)


tafia:
Thanks!

> I'm currently referencing the quick-xml crate via github in my project, so a release on crates.io is not super urgent for me

I'll take some time then to publish a new version. There are some other changes I started but didn't have time to finish yet ...

PR #51, opened 2017-02-08
One vector for all namespaces and prefixes.

vandenoever:
Allocations are expensive. The iterator over `XmlnsReader` returns a new Vec<u8> at every step. In addition most encountered namespace declarations require two allocations.

This PR changes add one Vec<u8> to the `XmlnsReader`. This vec grows and shrinks as namespaces are added and removed. The struct `Namespace` gets a `NamespaceSlice` instead of `Vec<u8>`.

All tests pass. However, the benchmarks do not show a measurable difference, probably because even ` bench_quick_xml_namespaced` does not use the namespaces.

### DIFF
--- a/src/namespace.rs
+++ b/src/namespace.rs
@@ -3,6 +3,7 @@
 use {XmlReader, Event, Element};
 use error::ResultPos;
 use std::io::BufRead;
+use std::rc::Rc;
 
 /// Start and end index into a vector with prefixes and namespaces
 #[derive(Clone)]
@@ -125,6 +126,8 @@ pub struct XmlnsReader<R: BufRead> {
     /// Buffer that holds concatenated prefixes and namespaces.
     /// The offsets are stored in stack of Namespace objects.
     namespace_buffer: Vec<u8>,
+    /// Buffer for passing to the caller of next().
+    iter_buffer: Rc<Vec<u8>>
 }
 
 impl<R: BufRead> XmlnsReader<R> {
@@ -135,7 +138,8 @@ impl<R: BufRead> XmlnsReader<R> {
             namespaces: Vec::new(),
             nesting_level: 0,
             pending_pop: false,
-            namespace_buffer: Vec::new()
+            namespace_buffer: Vec::new(),
+            iter_buffer: Rc::new(Vec::new())
         }
     }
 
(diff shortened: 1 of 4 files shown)


vandenoever:
`Iterator::Item` has no lifetime linked to `Iterator`, so `fn next()` has to use `Rc` or `clone()` to pass the namespaces in the events.

This PR makes counting the elements on a 1G XML file 30% slower. I cannot explain this.

Pulling XML events via an API different from `Iterator` could be faster because it can avoid `Rc` and `clone()`.

tafia:
Thanks for the PR!
I'm sorry but I don't have much time today. I'll try to review it tomorrow.

tafia:
Just to let you know that I haven't forgotten you. I was very busy this week.

tafia:
> This PR makes counting the elements on a 1G XML file 15% slower. I cannot explain this.

From my tests I couldn't spot any noticeable difference in terms of performance. I guess the `Rc::make_mut` does a deep clone most of the time.

> Pulling XML events via an API different from Iterator could be faster because it can avoid Rc and clone().

Could you add a `next_item` fn so it avoids the clones and also a bench using this function?

> Iterator::Item has no lifetime linked to Iterator, so fn next() has to use Rc or clone() to pass the namespaces in the events.

I have started reshaping the library entirely to avoid this lifetime issue once for all ((not necessarily for namespaces only) ) but I couldn't find the time to finish it.

tafia:
> I have started reshaping the library entirely to avoid this lifetime issue once for all ((not necessarily for namespaces only) ) but I couldn't find the time to finish it.

I was referring to #52. Here, the `XmlBytesReader` doesn't implement `Iterator` at all. Everything turns around [`fn next_event`.](https://github.com/tafia/quick-xml/pull/52/files#diff-a093fac65193a08fd2f330a35c4da0bbR667)

tafia:
I will probably merge the refactoring I'm doing in #52 this week or next week.
It also contains the changes you've done here. I'll remove the *old* `XmlReader` in the process which means this PR will probably never get merged, Is it ok for you?

vandenoever:
Sure. the other branch seems to be moving in a great direction.

tafia:
fine

PR #52, opened 2017-02-17
Major refactoring: non allocating reader

tafia:
This PR adds a faster lower level `XmlBytesReader` reader.
It also rearrange modules to split readers/writer

Add a reader::bytes module with all necessary tools to read at *bytes* level.

A XmlBytesReader is a reader which doesn't own the buffer the data is read into.
It does very little allocation saving lot of time/memory. The tradeof is of course
on end user side to manage the buffer and its lifetime issues.

The performance boost is significant
```
running 5 tests
test bench_quick_xml                       ... bench:     524,770 ns/iter (+/- 193,777)
test bench_quick_xml_bytes                 ... bench:     321,664 ns/iter (+/- 25,172)
test bench_quick_xml_escaped               ... bench:     651,365 ns/iter (+/- 45,542)
test bench_quick_xml_namespaced            ... bench:     781,093 ns/iter (+/- 245,433)
test bench_quick_xml_namespaced_while_loop ... bench:     725,188 ns/iter (+/- 35,516)
```

This is still a WIP as I plan to update other reader to use XmlBytesReader internally.


src/reader/bytes.rs
```diff
@@ -0,0 +1,906 @@
...
+/// General content of an event (aka node)
+///
+/// BytesElement is a wrapper over the bytes representing the node:
+///
+/// E.g. given a node `<name att1="a", att2="b">`, the corresponding `BytesEvent` will be
+///
+/// ```ignore
+/// BytesEvent::Start(BytesElement {
+///     buf:    b"name att1=\"a\", att2=\"b\"",
+///     start:  0,
```
vandenoever: This is out of sync with the code which only members `buf`, `content`, and `name`.

src/reader/bytes.rs
```diff
@@ -0,0 +1,906 @@
...
+/// - no attribute parsing: use lazy `Attributes` iterator only when needed
+/// - no namespace awareness as it requires parsing all `Start` element attributes
+/// - no utf8 conversion: prefer searching statically binary comparisons
+/// then use the `as_str` or `into_string` methods
+#[derive(Clone)]
+pub struct BytesElement<'a> {
+    /// content of the element, before any utf8 conversion
+    buf: Cow<'a, [u8]>,
+    /// content range, excluding text defining BytesEvent type
+    content: Range<usize>,
```
vandenoever: Why not:
```rust
pub struct BytesElement<'a> {
    /// bytes from element tag between '<' and '>'
    buf: Cow<'a, [u8]>,
    /// end of the element name, the name starts at that the start of `buf`
    name_end: usize,
}
```
That's only half the size and bookkeeping.

tafia:
Thanks for your comments. They are both valid. I'll make some changes.

tafia:
@vandenoever I pushed all the logic you've implemented in #51 directly into `BytesReader`

```
running 7 tests
test bench_quick_xml                       ... bench:     499,187 ns/iter (+/- 34,727)
test bench_quick_xml_bytes                 ... bench:     305,979 ns/iter (+/- 16,767)
test bench_quick_xml_bytes_wrapper         ... bench:     304,285 ns/iter (+/- 12,925)
test bench_quick_xml_escaped               ... bench:     635,140 ns/iter (+/- 87,921)
test bench_quick_xml_namespaced            ... bench:     727,392 ns/iter (+/- 53,984)
test bench_quick_xml_namespaced_bytes      ... bench:     465,964 ns/iter (+/- 31,183)
test bench_quick_xml_namespaced_while_loop ... bench:     711,604 ns/iter (+/- 42,858)
```

vandenoever:
This is what I get for 265dffb1247b992fe722e276877ceac86e97f87d with a 22M xml file with namespaces on most elements and attributes.

```
running 8 tests
test bench_quick_xml                             ... bench:  51,986,452 ns/iter (+/- 1,689,679)
test bench_quick_xml_bytes                       ... bench:  31,116,674 ns/iter (+/- 565,364)
test bench_quick_xml_bytes_escaped_bytes_wrapper ... bench:  37,406,926 ns/iter (+/- 988,967)
test bench_quick_xml_bytes_wrapper               ... bench:  32,895,290 ns/iter (+/- 2,757,370)
test bench_quick_xml_escaped                     ... bench:  59,996,850 ns/iter (+/- 10,532,006)
test bench_quick_xml_namespaced                  ... bench: 114,945,458 ns/iter (+/- 32,631,223)
test bench_quick_xml_namespaced_bytes            ... bench: 102,024,259 ns/iter (+/- 26,868,078)
test bench_quick_xml_namespaced_while_loop       ... bench: 113,376,712 ns/iter (+/- 23,759,873)
```

tafia:
Not impressive.
I am a little disappointed, because I thought it would be much faster (from the bytes level plus the better allocation on namespaces that you've implemented).
I'll have a look again.

Thanks for the figures!

tafia:
I've pushed some simplification.
I can't notice any difference though. I suspect in your case working with a HashMap might be faster,
Hard to tell like that. 
Also hard to say which xml is closer to the average xml: the sample or yours

vandenoever:
I've placed the file I use on https://vandenoever.info/content.xml.bz2

tafia:
With your file on my pc at home (very slow):
```
running 2 tests
test bench_quick_xml_namespaced_bytes_content    ... bench: 349,383,842 ns/iter (+/- 40,013,902)
test bench_quick_xml_namespaced_content          ... bench: 427,308,499 ns/iter (+/- 20,505,964)
```

tafia:
And with my computer at work:
```
running 3 tests
test bench_quick_xml_namespaced                  ... bench: 116,580,214 ns/iter (+/- 13,590,367)
test bench_quick_xml_namespaced_bytes            ... bench:  84,376,254 ns/iter (+/- 3,603,920)
```

I still use the `include_bytes!` macro to be certain to remove file reading penalty.
Can you give it another try?
Perhaps the simplification I've done yesterday actually paid off.

tafia:
Ok I'll stop here for today.
I'll polish tomorrow but I think I'm where I wanted to be at the beginning:
- `XmlReader` replaced with a non allocating `Reader` (which takes an external buffer)
  - do not implement `Iterator` anymore but I think it wasn't really helpful anyway and the benefit is too huge
- remove the `XmlnsReader` and replace it with a simple `read_namespaced_event`
- remove the `UnescapedAttribute` and instead provide a new `Attribute` struct with `unescape` methods
- remove the `AsStr` trait and support xml encodings. This pushes the user to use the `unescape_and_decode` fns when necessary. Or to run in two steps `unescape` then `Reader::decode`. People already know too well `from_utf8` fns to get the old behavior
- refactor all module hierarchies
- remove `Element`s and replace them with several `BytesStart` etc... which are more specific to the event they cover.
- do not use `from_utf8` when unescaping ascii codes. Instead use `from-ascii` crate which works at bytes level
- rename `XmlWriter` to `Writer` and provide a way to write `&[u8]` directly (now that the user owns the buffer, it could be faster)
- adds @vandenoever changes to save some namespaces allocations

Overall I believe the API is nicer now, even if it provides a lower level of control.
And of course it is much faster 😄 

```
 name                        master ns/iter  bytes ns/iter  diff ns/iter   diff %
 bench_quick_xml             525,368         314,472            -210,896  -40.14%
 bench_quick_xml_escaped     655,665         404,951            -250,714  -38.24%
 bench_quick_xml_namespaced  742,002         449,032            -292,970  -39.48%
 bench_xml_rs                15,220,356      15,462,258          241,902    1.59%
```

@vandenoever would you mind having a look, I know this is a lot of changes but you've followed it a little.

PR #56, opened 2017-03-07
Experimental: add dom wrapper

tafia:
This PR adds a very basic DOM like wrapper on top of quick-xml pull based reader.

The typical usage is:
```rust
use quick_xml::dom::Node;

let file = File::open(...);
let root_node = Node::root(file).expect("error while converting xml into DOM node");
let special_nodes = root_node.select_all("child1/child2/child3");
for n in special_nodes {
    println!("attributes: {}", n.attributes());
}
for n in root_node.children_mut() {
    *n.name_mut() = "My new Name".to_string();
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quick-xml"
-version = "0.6.2"
+version = "0.7.0"
 authors = ["Johann Tuffe <tafia973@gmail.com>"]
 description = "High performance xml reader and writer"
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,9 @@
   - test: Adding missing tests
   - chore: Changes to the build process or auxiliary tools/libraries/documentation
 
+## 0.7.0
+- feat: add dom module to have a DOM like interface wrapper
+
 ## 0.6.2
 - fix: another overflow bug found with cargo-fuzz
 - refactor: update dependencies
--- a/README.md
+++ b/README.md
@@ -131,6 +131,11 @@ let mut root = Node::open(path).expect("cannot read file");
         println!("node: name: {}, attributes count: {}, children count: {}",
             n.name(), n.attributes().len(), n.children().len());
     }
+
+    // alternatively, in we only need these nodes, there is a faster version
+    // which performs less allocation:
+    //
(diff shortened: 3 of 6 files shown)


blanham:
Do you plan on merging this anytime soon?

tafia:
I don't know. I do not really need it at the moment and I haven't been asked it. It was more a proof of concept.
Would you be interested?

danger89:
I'm interested, but wouldn' it be better to create a separate crate for this (probably also the reason you don't merge it into quick-xml now)? Thus also a separate git repo in this same project.

tafia:
Yes it would definitely be better on another crate. Eventually I'd love having a crate just to declare the traits to have a DOM and several implementation (including one for quick-xml, xml-rs, xml5ever, RustyXML etc ...).

I am still struggling to find some time but if anyone wants to starts I'll be happy to help!

danger89:
I try to use your dom, after some hacking (like escape module is private), I now get stuck at:

> error[E0382]: use of moved value: `start`
>    --> src/parse/dom.rs:340:41
>     |
> 339 |         start.with_attributes(self.attributes.iter().map(|&(ref k, ref v)| (&**k, &**v)));
>     |         ----- value moved here
> 340 |         writer.write_event(Event::Start(start))?;
>     |                                         ^^^^^ value used here after move
>     |
>     = note: move occurs because `start` has type `quick_xml::events::BytesStart<'_>`, which does not implement the `Copy` trait
> 
> error: aborting due to previous error
>

tafia:
Sorry I haven't answered that!
This dom is really very experimental for the moment. There are definitely some pain points to fix.
For instance, reading at the error, I am not sure the writer *needs* to move the value. A borrow should be enough.

tafia:
I am closing this for now as [minidom crate](https://crates.io/crates/minidom) has apparently [switched](https://gitlab.com/lumi/minidom-rs/merge_requests/11) to using quick-xml.

Let me know if you think otherwise.

ColinKinloch:
I updated this code to work with the current quick-xml [here](https://github.com/ColinKinloch/quick-xml/blob/dom_update/src/dom.rs).
I don't intend to continue working on it.

Issue #59, opened 2017-03-17
`read_event` vs. `read_namespaced_event`


pfernie:
Interleaving calls to `read_event` and `read_namespaced_event` could lead to surprising behavior (internal stack of namespaces would not be properly maintained?). It seems the `Reader` should always be either tracking or not tracking the namespaces, and as such should expose some configuration `fn parse_namespaces(self, parse: bool) -> Self`. That, or perhaps there should be separate implementations, `EventReader`, `NamespacedEventReader`, with only the latter exposing fns such as `read_namespaced_event`.

pfernie:
I added a [gist](https://gist.github.com/pfernie/c2b1176fc57fa902ed11351a16e89a70) showing an (admittedly pathological) usage of `read_event` & `read_namespaced_event` to illustrate.

Mingun:
Yes, the current design of reading namespaces are not good. I'd prefer that namespace buffer would be in the `Reader`, because, unlike buffer for events, I cannot imagine why the user may want to clear it -- if he/she would do that, he/she lose all namespaces information and probably even breaks quick-xml.

PR for moving namespaces buffer inside `Reader` and always be aware of namespaces is welcome.

dralley:
@Mingun Is there any work left here or can it be closed?

Mingun:
Done in #437, can be closed

PR #60, opened 2017-03-17
Add `local_name`

### LINKED ISSUE #57: Strip namespace from node name
This could be either a `BytesStart` method or an update of `read_namespaced_event`.

Linked to https://github.com/tafia/calamine/pull/73

pfernie:
An attempt to resolve #57. This adds a simple convenience fn `local_name()` to `Bytes{Start,End}`. I opted not to take the naive approach of simply returning any content after a ':', but rather implemented in terms of `resolve_namespace` after reviewing `read_namespaced_event()` again. As such, `local_name()` takes a `&Reader` via which it does the actual resolution. Documentation notes that a call to `read_namespaced_event` is needed in order to use `local_name()`.

This raises the question if it makes sense to have both `read_event` and `read_namespaced_event`, which I have opened as #59.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ loop {
             let mut elem = BytesStart::owned(b"my_elem".to_vec(), "my_elem".len());
 
             // collect existing attributes
-            elem.with_attributes(e.attributes().map(|attr| attr.unwrap()));
+            elem.extend_attributes(e.attributes().map(|attr| attr.unwrap()));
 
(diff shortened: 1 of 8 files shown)


tafia:
First thanks again for the PR and sorry for the late answer.

While this is correct I believe this is not efficient as we do not need to resolve anything (ie find the corresponding namespace in list of namespaces) here, we just need to subslice as you've done.

pfernie:
Ok, closing this in favor of #61

PR #61, opened 2017-03-21
impl simple local_name()

pfernie:
Implements `local_name()` naively, returning the name with any initial prefix up to and including the first ':' character removed. Does not concern itself with correct xml conformance:

* does not validate that the namespace is known
* `<foo: attr='bar'/>` will have an empty `local_name()`
* `<:foo attr='bar'/>` (empty namespace) will return `local_name()` of "foo"
* `<foo:bus:baz attr='bar'/>` will return `local_name()="bus:baz"` (name after __first__ ':' encountered)

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -68,6 +68,18 @@ impl<'a> BytesStart<'a> {
         &self.buf[..self.name_len]
     }
 
+    /// local name (excluding namespace) as &[u8] (without eventual attributes)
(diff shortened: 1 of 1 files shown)


tafia:
Thanks!

tafia:
> `<foo:bus:baz attr='bar'/>` will return local_name()="bus:baz" (name after first ':' encountered)

I believe it is ok and `bus:baz` is indeed the local name in that situation.

PR #69, opened 2017-06-12
Have read_namespace_event borrow the ns buffer

tafia:
The current situation, where the reader owns the namespace buffer leads to an inconvenient api.

In the inner loop of the `read_namespaced_event`, the reader, mutably borrowed cannot be used, notably for decoding the event. The workaround, having temporary variables is not satisfactory (more boilerplate and additional checks).

Making the caller own the buffer fixes the situation, while still avoiding extra allocation.

See #68

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quick-xml"
-version = "0.7.3"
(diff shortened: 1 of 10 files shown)


tafia:
Published v0.8.0

PR #71, opened 2017-06-19
use license identifier instead of file in Metadata

pacman82:
This will help the cargo `license` subcommand to correctly identify this as an MIT license, instead of non-standard. Great job with the library by the way. Helped me speed up parsing xml files by factor 30.


tafia:
Thanks!
I think clippy was warning against this long ago, but if it can help other tools, I'm fine changing it!

PR #76, opened 2017-07-06
Allow writer to borrow Event, using Deref

tafia:
This pull request strictly improves the current situation (shouldn't be a breaking change).  

Will probably fix situations like https://github.com/tafia/quick-xml/pull/56#issuecomment-310488378


tafia:
cc @danger89

PR #77, opened 2017-07-06
Allow writer to borrow event

tafia:
should fix https://github.com/tafia/quick-xml/pull/56#issuecomment-310488378


tafia:
cc @danger89

PR #78, opened 2017-07-25
Escape Text::Event on write

huntiep:
Per rust-syndication/rss#32. What did you mean by "special characters represented by their ascii code"?

### DIFF
--- a/src/escape.rs
+++ b/src/escape.rs
@@ -18,6 +18,24 @@ enum ByteOrChar {
     Char(u32),
 }
 
+/// helper function to escape a `&[u8]` and replace all
+/// xml special characters (<, >, &, ', ") with their corresponding
+/// xml escaped value.
+pub fn escape(raw: &[u8]) -> Vec<u8> {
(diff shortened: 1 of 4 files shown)


tafia:
Sorry for the very late answer.

Most of the time escaping is a no-op. I would then expect `escape` to return a `Cow`. Could you try replicating the logic we have in unescape fn for instance to postpone operations as much as possible?

tafia:
Thanks for the change!
Could you please add a test to be sure we won't regress in the future?

huntiep:
Sorry this took so long. I was on vacation and it slipped my mind when I got back.

tafia:
Thanks!!

tafia:
Fyi, I just published version 0.9.0 including your PR.

huntiep:
Awesome!

andy128k:
The same issue is present in Attribute. I suppose From<(&str, &str)> for Attribute should call escape too.

PR #91, opened 2017-09-14
Escape attributes

andy128k:


src/events/attributes.rs
```diff
@@ -59,13 +59,13 @@ pub struct Attribute<'a> {
     /// the key to uniquely define the attribute
     pub key: &'a [u8],
     /// the value
```
tafia: Can you mention is the doc that the `value` is the `unescaped` value?

And that converting from a `(&[u8], &[u8])` doesn't try to escape the value while converting from a `(&str, &str)` does?

andy128k:
@tafia Of course, I can make changes in documentation too. But the fact that types of methods "version", "encoding", and "standalone" of BytesDecl are changed disturbs me.

tafia:
> But the fact that types of methods "version", "encoding", and "standalone" of BytesDecl are changed disturbs me.

I agree it is not ideal. On the other hand it make it all more consistent. I don't mind having a `Cow`. I don't really mind having a `&[u8]` either to be honest. One is transparency, the other one is usability (with no performance impact).

andy128k:
Any updates on this PR?

tafia:
Sorry! It completely got under the radar.
I'll have a lot asap!

tafia:
Thanks again for the PR and sorry for the time it took me to merge it!

andy128k:
Thanks!

PR #96, opened 2017-11-15
Fix double escape

andy128k:


tests/unit_tests.rs
```diff
@@ -615,10 +642,45 @@ fn test_read_write_roundtrip_escape() {
...
+
+    let mut reader = Reader::from_str(input);
+    reader.trim_text(false).expand_empty_elements(false);
+    let mut writer = Writer::new(Cursor::new(Vec::new()));
+    let mut buf = Vec::new();
+    loop {
+        match reader.read_event(&mut buf) {
+            Ok(Eof) => break,
+            Ok(Text(e)) => {
+                let t = String::from_utf8(e.unescaped().unwrap().into_owned()).unwrap();
```
tafia: I'd prefer using a `unescape_and_decode` or `unescape` then `decode` separately instead of `String::from_utf8`. Could you change it?

tests/unit_tests.rs
```diff
@@ -615,10 +642,45 @@ fn test_read_write_roundtrip_escape() {
...
                 assert!(
                     writer
-                        .write_event(Event::Text(BytesText::borrowed(&t)))
+                        .write_event(Event::Text(BytesText::from_str(t)))
```
tafia: Can you add a comment here to insist on the use of `from_str`?

andy128k: @tafia 
`t` is a String in this context.

BytesText::from_str is a complementary to BytesText::unescape_and_decode.
In the same way as BytesText::borrowed and BytesText::owned are to BytesText::escaped.

andy128k:
@tafia gentle bump

tafia:
Thanks!

tafia:
Publishing v0.10.0.
Thanks again!

CryZe:
Looks like this PR caused a breaking change for me. It seems like neither the borrowed nor owned constructors document whether they want the data pre-escaped. Before you needed to give it unescaped data, which is fine, but now you need to escape it yourself, which is a breaking change. I feel like this should be documented, so other people don't run into trouble here.

CryZe:
Additionally, I don't even know how to even escape here, as there doesn't seem to be an actual escape functionality exposed anymore.

CryZe:
I'll create a PR fixing this.

PR #100, opened 2017-12-07
Migrate from 'error-chain' to 'failure'.

andy128k:
**Do not merge**

I've introduced enums for different errors. There are still few string-based errors.
I'm not sure if such granularity of errors is needed. Docs are also missing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ travis-ci = { repository = "tafia/quick-xml" }
 
 [dependencies]
 encoding_rs = "0.7.1"
-error-chain = "0.11.0"
+failure = "0.1.1"
 memchr = "2.0.0"
 
 [dev-dependencies]
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -2,22 +2,46 @@
 
 #![allow(missing_docs)]
 
-error_chain! {
-    foreign_links {
-        Io(::std::io::Error);
-    }
-    errors {
-        EndEventMismatch(expected: String, found: String) {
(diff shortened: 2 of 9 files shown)


tafia:
This is a breaking change that is very expected, thank you for all your work!

Do you see any drawback in using failure over error-chain so far? 
How was the transition (I want to use it on some other libs because of the no_std).

andy128k:
@tafia 
`error-chain`'s macro is a custom syntax and gives cryptic errors if you make a typo.
In opposite, syntax of `failure` is just a rust with annotation. (Frankly speaking, `failure` also could be cryptic if you make error in `#[fail()]` annotation).

`error-chain` gives a single error type for your module with finite list of possible errors (ErrorKind enum). But `failure::Error` is an umbrella which can hold ANY error. So, maybe `failure::Error` is not a good choice for a library and custom error type should be defined (with derived Fail).

As soon as this error type is a part of library's API, it must be well defined. It could be enum (to mimic `error-chain`) or just a string-based struct (for simplicity). This is the question to be answered before any migrations.

andy128k:
@tafia I've updated code. Now it defines two error enums, `errors::Error` and `escape::EscapeError`. All string errors are replaced by corresponding cases of Error enum.

tafia:
Thanks.
It looks ready to be merged and I have tried using Failure in some other projects and it is nice!
Do you want me to merge it now?

andy128k:
Please merge.

tafia:
Thanks!

PR #102, opened 2017-12-20
Issue101

### LINKED ISSUE #101: Do not use unchecked multiply
```
thread '<unnamed>' panicked at 'attempt to multiply with overflow', /home/linkmauve/data/cargo/registry/src/github.com-1ecc6299db9ec823/quick-xml-0.10.0/src/escape.rs:166:9
```

Here is the testcase generated by cargo-fuzz:
```
\x00\x00<\x00\x00\x0a>&#44444444401?#\x0a413518#\x0a\x0a\x0a;<:<)(<:\x0a\x0a\x0a\x0a;<:\x0a\x0a<:\x0a\x0a\x0a\x0a\x0a<\x00*\x00\x00\x00\x00
```

The `&#44444444401` part seems to be the issue, multiplying that by ten does not fit in an `u32`, causing the panic.

This bug has been found while fuzzing [xmpp-parsers](https://hg.linkmauve.fr/xmpp-parsers) which uses [minidom](https://gitlab.com/lumi/minidom-rs), you would benefit a lot from fuzzing quick-xml itself.

tafia:
Closes #101


tafia:
There is an issue with downloading rust nightly on CI servers. I'll try again later.

PR #104, opened 2018-01-03
Basic indentation support in Writer

mxj4:
Thanks for the great library!
Not sure why there is no indentation support in Writer, but I need indentation when producing human-readable XML files, so I added some basic indentation support there.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -63,12 +63,21 @@ use events::Event;
 pub struct Writer<W: Write> {
     /// underlying writer
     writer: W,
+    indent: Option<Indentation>,
 }
 
 impl<W: Write> Writer<W> {
     /// Creates a Writer from a generic Write
     pub fn new(inner: W) -> Writer<W> {
-        Writer { writer: inner }
+        Writer { writer: inner, indent: None }
+    }
+
+    /// Creates a Writer with configured whitespace indents from a generic Write
+    pub fn new_with_indent(inner: W, indent_char: u8, indent_size: usize) -> Writer<W> {
+        Writer {
+            writer: inner,
+            indent: Some(Indentation::new(indent_char, indent_size)),
+        }
     }
 
     /// Consumes this `Writer`, returning the underlying writer.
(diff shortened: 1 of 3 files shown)


src/writer.rs
```diff
@@ -74,7 +95,16 @@ impl<W: Write> Writer<W> {
 
     /// Writes the given event to the underlying writer.
     pub fn write_event<'a, E: AsRef<Event<'a>>>(&mut self, event: E) -> Result<usize> {
-        match *event.as_ref() {
+        if self.indent_size > 0 {
+            match *event.as_ref() {
+                Event::End(_) => {
+                    let new_len = self.indents.len() - self.indent_size;
+                    self.indents.truncate(new_len);
```
tafia: Not sure it is a good idea to truncate it because it will very likely grow for the next node.
I guess it could be better to use write `&indent[..indent_len]` directly and reduce the amount of `extend`/`truncate`.

src/writer.rs
```diff
@@ -74,7 +95,16 @@ impl<W: Write> Writer<W> {
 
     /// Writes the given event to the underlying writer.
     pub fn write_event<'a, E: AsRef<Event<'a>>>(&mut self, event: E) -> Result<usize> {
-        match *event.as_ref() {
+        if self.indent_size > 0 {
```
tafia: why don't you make the tests in the match?
It should be simpler.

```rust
match event.as_ref() {
    Event::End(ref e) => {
         if let Some(indent) = self.indent.as_mut() {
               // your code
         }
         self.write_wrapped(b"</", e, b">")
    }
    Event ...
    Event::Start(ref e) => {
         self.write_wrapped(b"<", e, b">"),
         if let Some(indent) = self.indent.as_mut() {
               // your code
         }
    }
}
```

src/writer.rs
```diff
@@ -59,12 +59,33 @@ use events::Event;
 pub struct Writer<W: Write> {
     /// underlying writer
     writer: W,
+    should_line_break: bool,
+    indent_char: u8,
+    indent_size: usize,
+    indents: Vec<u8>,
```
tafia: I think it would be better to have split indentation logic out of reader logic:
```rust
struct Indentation {
    should_line_break: bool,
    indent_char: u8,
    indent_size: usize,
    indents: Vec<u8>,
}

pub struct Writer<W:Write> {
    writer: W,
    indent: Option<Indentation>,
}
```
  
and put as much logic as possible directly in impl of the struct.

Also (see my other comments) it might be a good idea to directly initialize indent with `vec![indent_char; 128]` or so.

src/writer.rs
```diff
@@ -96,6 +138,14 @@ impl<W: Write> Writer<W> {
 
     #[inline]
     fn write_wrapped(&mut self, before: &[u8], value: &[u8], after: &[u8]) -> Result<usize> {
-        Ok(self.writer.write(before)? + self.writer.write(value)? + self.writer.write(after)?)
+        Ok(
+            if self.should_line_break {
+                self.writer.write(b"\n")? + self.writer.write(self.indents.as_slice())?
```
tafia: Better to use `writeln!` macro to use the platform new line.

src/writer.rs
```diff
@@ -85,7 +115,19 @@ impl<W: Write> Writer<W> {
             Event::PI(ref e) => self.write_wrapped(b"<?", e, b"?>"),
             Event::DocType(ref e) => self.write_wrapped(b"<!DOCTYPE", e, b">"),
             Event::Eof => Ok(0),
+        };
+        if self.indent_size > 0 {
+            match *event.as_ref() {
+                Event::Start(_) => {
+                    let new_len = self.indents.len() + self.indent_size;
+                    self.indents.resize(new_len, self.indent_char);
```
tafia: As explained before, the `resize` can be done only if the indent length is not sufficient.

src/writer.rs
mxj4: According to `writeln!` doc, "`On all platforms, the newline is the LINE FEED character (\n/U+000A) alone (no additional CARRIAGE RETURN (\r/U+000D).`". According to https://stackoverflow.com/questions/33867076/is-there-a-format-spec-that-will-output-a-os-specific-newline/33867420#33867420 , most Windows applications should already handle `"\n"`.

mxj4:
The latest commit addressed most of your concerns, can you take a look again? Thx!

tafia:
Thanks for the changes!
I have just merged the previous PR which migrates to Failure and unfortunately it conflicts with your PR.
Can you please rebase?
Sorry for that.

tafia:
Rebased

mxj4:
Rebased.

mxj4:
The failure seems to be a travis config issue.

tafia:
Thanks

PR #105, opened 2018-01-06
Html attributes

tafia:
This PR allows iterating on html syntax attributes.

Html attributes are more permissive than xml, in particular:
- attribute values do not need quotes if there is no whitespace
- for empty attribute values, we don't even need the `=` character

The simplest way to use it is though the new `BytesStart::html_attributes` function.

Note that depending on your use case, it may have a small performance penalty (within 2-3%) but most of my tests perform similarly. The main reason is that I do not use `memchar` in attributes iterator but a direct `Iter`. The code is simpler and less error prone.

It also add several new tests for attributes.


tafia:
Rebased

PR #110, opened 2018-01-24
Travis: Fix typo (nighlty -> nightly)

jonas-schievink:
Let's see if something broke :D


jonas-schievink:
```
$ travis-cargo --only nightly fmt -- --write-mode=diff
Unrecognized option: 'write-mode'
```

Now *that's* some bit rot

PR #114, opened 2018-02-01
Clarify and extend the API Surface of BytesText

CryZe:
Neither of the 3 constructors `borrowed`, `owned` and `from_str` documented if they expect pre-escaped data. This behavior also silently changed completely in #96 causing a few bugs here and there. The API was therefore completely revamped in order to allow constructing text events from either escaped or unescaped text.

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -298,34 +298,48 @@ impl<'a> BytesEnd<'a> {
 /// Data from various events (most notably, `Event::Text`).
 #[derive(Clone, Debug)]
 pub struct BytesText<'a> {
+    // Invariant: The content is always escaped.
     content: Cow<'a, [u8]>,
 }
 
 impl<'a> BytesText<'a> {
-    /// Creates a new `BytesText` borrowing a slice
+    /// Creates a new text event from an escaped byte sequence.
     #[inline]
-    pub fn borrowed(content: &'a [u8]) -> BytesText<'a> {
+    pub fn from_escaped<C: Into<Cow<'a, [u8]>>>(content: C) -> BytesText<'a> {
         BytesText {
-            content: Cow::Borrowed(content),
+            content: content.into(),
         }
     }
 
-    /// Creates a new `BytesText` owning its name
+    /// Creates a new text event from an unescaped byte sequence.
     #[inline]
-    pub fn owned(content: Vec<u8>) -> BytesText<'static> {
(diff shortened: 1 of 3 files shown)


tafia:
I agree it makes it more explicit.

Before merging, let's try to have @andy128k 's opinion. If he doesn't answer before this week end I'll merge it.

Thanks!

src/events/mod.rs
```diff
@@ -298,34 +298,48 @@ impl<'a> BytesEnd<'a> {
 /// Data from various events (most notably, `Event::Text`).
 #[derive(Clone, Debug)]
 pub struct BytesText<'a> {
+    // Invariant: The content is always escaped.
     content: Cow<'a, [u8]>,
 }
 
 impl<'a> BytesText<'a> {
-    /// Creates a new `BytesText` borrowing a slice
+    /// Creates a new text event from an escaped byte sequence.
```
andy128k: It's still `BytesText`, not an event.

andy128k: It's still `BytesText`, not an event.

andy128k:
I personally like explicit names. `from_escaped_*` is much better. The only thing which disturbs me is double negation in `from_unescaped_*`. Unescaped is just a plain text.

src/events/mod.rs
CryZe: Whoops, I actually didn't look at the original doc comment and just rewrote it from scratch, so it only somewhat resembles what is used to look like ^^'

Will fix

CryZe:
I can use `from_plain` instead if that's better.

CryZe:
I used `from_plain` / `from_plain_str` now. I made the documentation a bit more explicit for those 2, to clarify that they expect content that is not escaped.

andy128k:
escaped/unescaped make sense if you use `Reader`. But if the crate is used to write XML documents, these words are just a noise. You just want to create events with text. Escaping should not bother you at all. That's why I've chosen `from_str` to mimic standard crates.

Yeah, naming things is hard 😉

CryZe:
So are you okay with `from_plain_str` or should I switch it to `from_str`? I feel like the distinction should at least be somewhat visible in the naming here, otherwise that's an easy pitfall. I'm fine with changing it tho, if you want me to.

CryZe:
Just a friendly ping, since it's been 2 weeks now :)

PR #117, opened 2018-03-02
Added into_owned() for Bytes{End, Decl} and for Event

mark-blum:
`BytesStart` and `BytesText` both already have an `.into_owned()` method, while `BytesEnd` and `BytesDecl` don't have similar methods for some reason. Additionally, you can now `.into_owned()` on an `Event` as a convenience.


tafia:
Thanks!

PR #119, opened 2018-03-03
Unescaped now only returns the tag name

### LINKED ISSUE #118: BytesStart unescaped method returns the attributes in the same result
If you had for example the following xml snippet `<namespace:hello&gt; greeting='wave'>` and the corresponding BytesStart for that, (let's called the `element`), `element.unescaped()` would return `namespace:hello> greeting='wave'`, while `.name()` and `local_name()` only include the actual tag part (`namespace:hello&gt;` and `hello&gt;` respectively), and not the attributes.

This goes against the documentation which says it should just return the unescaped tag name.

mark-blum:
Fixes #118

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -94,7 +94,7 @@ impl<'a> BytesStart<'a> {
     /// XML escape sequences like "`&lt;`" will be replaced by their unescaped characters like
     /// "`<`".
     pub fn unescaped(&self) -> Result<Cow<[u8]>> {
-        unescape(&*self.buf).map_err(Error::EscapeError)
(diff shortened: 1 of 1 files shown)


mark-blum:
Doc examples added. I'm actually working on adding doc examples for most of the methods in this library so expect that some time in the future

tafia:
Great! Should be merged very soon once I get on my PC

tafia:
I have make a [small PR](https://github.com/mark-blum/quick-xml/pull/1) on top on your request in your fork. Let me know what you think.

tafia:
Hi @mark-blum do you plan on merging my PR so I can merge yours?

PR #135, opened 2018-09-13
Fix minor bug for parsing comment tag

yawara:
The following code terminates with panic `thread 'main' panicked at 'slice index starts at 6 but ends at 4', libcore/slice/mod.rs:1971:5`.

```rust
extern crate quick_xml;

use quick_xml::events::Event;
use quick_xml::Reader;

fn main() {
    let xml = "<a></a><!-->-->";
    let mut reader = Reader::from_str(&xml);
    reader.trim_text(true);
    let mut buf = Vec::new();
    loop {
            Ok(Event::Comment(ref e)) => {
                println!("Comment: \"{}\"", reader.decode(e));
            }
            Ok(Event::Eof) => break,
            Err(e) => println!("Error: {:?}", e),
            _ => (),
        }
    }
}
```

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -26,7 +26,10 @@ pub enum Error {
     #[fail(display = "Cannot read text, expecting Event::Text")]
     TextNotFound,
 
-    #[fail(display = "XmlDecl must start with 'version' attribute, found {:?}", _0)]
+    #[fail(
+        display = "XmlDecl must start with 'version' attribute, found {:?}",
+        _0
+    )]
     XmlDeclWithoutVersion(Option<String>),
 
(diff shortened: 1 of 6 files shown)


tafia:
Thanks for the fix!
Could you add a test to make sure it won't happen again?
Could you also run the latest version of rustfmt-preview on stable?

yawara:
@tafia I added the test.

If I run rustfmt on stable, the codes which I don't write will be changed because stable rust 1.29 was released yesterday.

tafia:
Thanks!

tafia:
Pushed v0.12.4

PR #140, opened 2018-09-28
Add into_underlying_reader method for Reader<BufRead + Seek>

### LINKED ISSUE #139: into_underlying_reader(self) -> B
Could you provide a conversion into underlying B: BufReader? It could be useful at least to compute line:column of the parsing failure.

artemshein:
Resolves #139

It doesn't make much sense to implement it for BufRead only as there is no way to get parsed data.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,9 @@
   - test: Adding missing tests
   - chore: Changes to the build process or auxiliary tools/libraries/documentation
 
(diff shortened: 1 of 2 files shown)


tafia:
As I said I'll publish it in only two weeks.
Thanks again!

tafia:
Published v0.13.1

PR #145, opened 2019-01-09
strip whitespaces at the end of closing tags

### LINKED ISSUE #144: Strip whitespaces at the end of closing markups.
Spaces after the end of the markup name in a closing tag are valid XML (See: https://www.w3.org/TR/REC-xml/#sec-starttags ). Yet the pares considers them part of the markup name and so checking for matching start and close may fail.

pacman82:
fixes #144. `read_end` now also always returns the markup name without trailing whitespaces.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -52,9 +52,7 @@ pub enum Error {
 
     #[fail(
         display = "error while parsing attribute at position {}: Duplicate attribute at position {} and {}",
-        _0,
-        _1,
-        _0
+        _0, _1, _0
     )]
     DuplicatedAttribute(usize, usize),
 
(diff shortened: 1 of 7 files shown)


tafia:
Awesome!
Could you please add  some method on the reader to avoid doing this check (default to stripping whitespace).
Thanks

pacman82:
@tafia I can. I want to note however that the start tag names are always stripped of whitespaces.

pacman82:
@tafia Done

pacman82:
fyi: the build breaks because of nightly rust. I would recommend to set nightly to allow failures. I can make a PR if you are going to accept it

tafia:
Thanks!

tafia:
Published v0.13.2 (also bumped dependencies).
Thanks again!

PR #149, opened 2019-02-15
Add set_name and clear_attributes functions to BytesStart

### LINKED ISSUE #148: Re-using BytesStart byte vector
Currently, `BytesStart` will [allocate a `Vec<u8>`](https://docs.rs/quick-xml/0.13.2/src/quick_xml/events/mod.rs.html#164-172) the first time you push additional attributes for the element. That's unfortunate if you're emitting lots of `BytesStart` in a tight loop. It would be better if it was possible to re-use a `BytesStart` (specifically its internal `Cow::Owned`) across iterations. I _think_ all that'd be needed would be two methods: `set_name` and `clear_attributes`:
```rust
impl BytesStart {
    fn set_name(&mut self, name: &[u8]) {
        let bytes = self.buf.to_mut();
        bytes.clear();
        bytes.extend(name);
        self.name_len = name.len();
    }

    fn clear_attributes(&mut self) {
        self.buf.to_mut().truncate(self.name_len);
    }
}
```

meven:
Fixes #148

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -170,6 +170,22 @@ impl<'a> BytesStart<'a> {
         bytes.extend_from_slice(&*a.value);
         bytes.push(b'"');
     }
+
+    /// Edit the name of the BytesStart in-place
+    pub fn set_name(&mut self, name: &[u8]) -> &mut BytesStart<'a> {
+        {
+            let bytes = self.buf.to_mut();
+            bytes.splice(..self.name_len, name.to_vec());
+        }
+        self.name_len = name.len();
+        self
+    }
+
+    /// Remove all attributes from the ByteStart
+    pub fn clear_attributes(&mut self) -> &mut BytesStart<'a> {
+        self.buf.to_mut().truncate(self.name_len);
+        self
+    }
 }
 
 /// An XML declaration (`Event::Decl`).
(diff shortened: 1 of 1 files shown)


jonhoo:
Those `fn`s should be `pub fn`, that's why you're getting the warnings :)

meven:
> Those `fn`s should be `pub fn`, that's why you're getting the warnings :)

Thanks ;) that's been taking care of.

jonhoo:
@tafia looks like the Travis failure is just `rustfmt-preview` being unavailable for nightly (the component is now just called `rustfmt`).

src/events/mod.rs
```diff
@@ -170,6 +170,23 @@ impl<'a> BytesStart<'a> {
         bytes.extend_from_slice(&*a.value);
         bytes.push(b'"');
     }
+
+    /// Edit the name of the BytesStart in-place
```
tafia: Can you add a warning stating that the content of the `BytesStart` will be cleared?
Someone may not expect attributes to disappear.

tafia:
Thanks for the PR and sorry for the late response.
Can you please just modify the comment on `set_name`?
I'll push changes to travis in the meantime.

src/events/mod.rs
meven: Good catch, I have added a warning.

meven:
Last change made me think that perhaps `set_names` should not edit the attributes and keep them.
It would end up as :
```rust
    /// Edit the name of the BytesStart in-place
    pub fn set_name(&mut self, name: &[u8]) -> &mut BytesStart<'a> {
        {
            let bytes = self.buf.to_mut();
            bytes.splice(..self.name_len, name.to_vec());
        }
        self.name_len = name.len();
        self
    }
```

tafia:
Sounds better indeed!

meven:
> Sounds better indeed!

Done, I feel like this can be merged

tafia:
Thanks!

src/events/mod.rs
```diff
@@ -170,6 +170,22 @@ impl<'a> BytesStart<'a> {
...
+    pub fn set_name(&mut self, name: &[u8]) -> &mut BytesStart<'a> {
+        {
+            let bytes = self.buf.to_mut();
+            bytes.splice(..self.name_len, name.to_vec());
```
jonhoo: @meven  I don't think the `to_vec` call here is needed. It'd be good to avoid the extra allocation!

meven: I did not find a better way I would expect the same but I ran into 
```console
error[E0271]: type mismatch resolving `<&[u8] as std::iter::IntoIterator>::Item == u8`
   --> src/events/mod.rs:178:19
    |
178 |             bytes.splice(..self.name_len, name);
    |                   ^^^^^^ expected reference, found u8
    |
    = note: expected type `&u8`
               found type `u8`
```
It seems I still need some conversion, name `&[u8]` implements `IntoIterator<u8>` but I need an `IntoIterator<&u8>`. Maybe There is some simple trick I am missing.

jonhoo: Ah, I've run into this in the past. It's a little silly. Basically what you want is `name.iter().cloned()` or `name.iter().map(|&b| b)` to tell the compiler that you want to utilize the fact that `u8` is `Copy`.

meven: Thanks, will make another PR

PR #150, opened 2019-02-19
Avoids allocating a Vec when setting the name

meven:
Improves on #149 

cc @jonhoo


jonhoo:
@tafia it would be awesome if we could also get a minor release after this has landed so we can use it in https://github.com/jonhoo/inferno/issues/41 :)

tafia:
Thanks to both of you!

tafia:
> @tafia it would be awesome if we could also get a minor release after this has landed so we can use it in [jonhoo/inferno#41](https://github.com/jonhoo/inferno/issues/41) :)

I've just published v0.13.3.

PR #152, opened 2019-04-01
Make the dependency on `failure` optional; use `derive_more` for deriving `Display`

koute:
I'm not sure if this is acceptable, but it seemed simple and easy enough so I decided to try and create a PR.

A little backstory - at work we have a [CPU profiler written in Rust](https://github.com/nokia/not-perf); recently I've added support so that our profiler can natively generate profiling flamegraphs using the [`inferno`](https://github.com/jonhoo/inferno) crate. The `inferno` crate uses `quick-xml` for its XML needs.

Now here's the problem - `quick-xml` requires the `failure` crate, and the `failure` crate requires `backtrace`, which requires `backtrace-sys`. This is a problem for our CPU profiler as `backtrace-sys` is not a pure Rust dependency - it actually requires a full C toolchain to be installed, so adding an innocent dependency which is only supposed to generate a bit of XML surprisingly broke our cross-compilation, and it even broke `cargo check --target=$arch` (as seen [on Travis here](https://travis-ci.org/nokia/not-perf/jobs/514131389)) since now even that has to invoke a cross-toolchain.

This made me realize that maybe using `failure` in crates which are meant to be used as a library might not be the greatest idea. From what I can see there even seems to be more-or-less a consensus across the community [that it's preferable to avoid `failure`](https://www.reddit.com/r/rust/comments/a9wbs8/eli5_error_handling_errorchain_failure_upcoming/), and the general direction seems to be [to fix the `Error` trait instead of replacing it wholesale](https://github.com/rust-lang/rfcs/blob/master/text/2504-fix-error.md).

Of course I realize that it's silly to demand anything from `quick-xml` here as this isn't really the `quick-xml`'s fault, however I do in general feel like this would nevertheless be an improvement. Please feel free to close this PR if you disagree. (:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,6 @@ before_script:
 script:
   - if [[ $(rustup show active-toolchain) == stable* ]]; then cargo fmt -- --check; fi;
   - cargo build
+  - cargo build --no-default-features
   - cargo test
+  - cargo test --no-default-features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ travis-ci = { repository = "tafia/quick-xml" }
 
 [dependencies]
 encoding_rs = "0.8.14"
-failure = "0.1.5"
+derive_more = "0.14"
 memchr = "2.1.2"
 log = "0.4.6"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ travis-ci = { repository = "tafia/quick-xml" }
 
 [dependencies]
 encoding_rs = "0.8.14"
+failure = { version = "0.1.5", optional = true }
 derive_more = "0.14"
 memchr = "2.1.2"
 log = "0.4.6"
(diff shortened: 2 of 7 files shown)


tafia:
Thanks for the PR!

I am not a big fan of failure anymore too so I definitely like what you've done.

I would like to keep the breaking changes to a minimum though and I was waiting for the error handling story to settle a bit before doing any move.

On the other hand I would be very interested to merge this behind a feature flag. Is it something you would be ok to do?

Thanks again and thanks for using quick-xml!

koute:
Fair enough!

I've added a turned-on-by-default feature flag. It's still using `derive_more` to implement `Display`, but now the error types also implement `Fail` just as they used to, so this shouldn't be a breaking change now, I think. Is this now acceptable?

tafia:
Thanks!
Sorry for the late merge, I didn't see your message.
I've just pushed v0.14.0

CryZe:
Oh wow, this was a big problem for me too. quick-xml was the only crate that forced failure and thus backtrace upon my crate and it was so painful for cross compilation. Thanks so much for switching away from it (Honestly they should‘ve just released a version of failure without backtrace, but it was soon abandoned before fixing the problems)

PR #159, opened 2019-05-17
Add a UTF-16 encoded test case

cmyr:
I've been trying to clarify my understanding of how this crate handles non-ascii compatible encodings, and so I added this test case, which attempts to parse a utf-16 document.

This is currently failing, presumably because we do all of our scans looking for bytes, and (e.g.) the `?` character is not represented by `0x3F` in utf-16.

Very possible I'm missing something here, just trying to figure out what it is. 🤓


tafia:
Thanks for the PR and sorry for the late answer.
Do you mind fixing the tests and fmt?

cmyr:
Sorry, I should've clarified: this is a PR of a _failing_ test case, intended for discussion but not suitable for merging.

My current understanding is that this test case demonstrates a fundamental problem with the current handling of non-ascii-compatible encodings, although I could also be misunderstanding something 😁

dralley:
Currently working on encoding support - new tests will be added in the process

PR #160, opened 2019-05-22
Rerun rustfmt

cmyr:
There were rustfmt errors in #159 that were in files I hadn't changed, presumably because of a rustfmt update.


tafia:
Thanks!

PR #162, opened 2019-06-25
Remove the `Seek` bound from `Reader::into_underlying_reader`

koute:
As far as I can see this bound is currently pointless while it prevents the user from getting back the underlying reader if it doesn't implement `Seek`.


tafia:
Thanks!

tafia:
Publishing v0.15.0

PR #166, opened 2019-08-23
Allow less bloat plus some speedup

tafia:
Following [Raph Levien's post](https://raphlinus.github.io/rust/2019/08/21/rust-bloat.html) regarding bloat, allow quick-xml user to opt in for bloats..
* allow NOT using encoding_rs when we know input xml to be UTF8
* disable both failure and encoding_rs per default, use features `use-failure` and `encoding` to revert them back (**BREAKING CHANGE**)
* speed up reader even further leveraging on memchr3 iterator (about 18% better on quick-xml benches)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ encoding_rs = { version = "0.8.17", optional = true }
 failure = { version = "0.1.5", optional = true }
 derive_more = "0.15"
 memchr = "2.2.1"
-log = "0.4.8"
 
 [dev-dependencies]
 xml-rs = "0.8.0"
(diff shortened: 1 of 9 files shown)


tafia:
speedups:
```
 name                        old ns/iter  new ns/iter  diff ns/iter   diff %  speedup
 bench_quick_xml             244,066      200,020            -44,046  -18.05%   x 1.22
 bench_quick_xml_escaped     331,202      285,513            -45,689  -13.79%   x 1.16
 bench_quick_xml_namespaced  439,700      367,280            -72,420  -16.47%   x 1.20
```

PR #170, opened 2019-09-07
implement std::error::Error and remove failure dependency

andy128k:
Failure derives `Fail` from `std::error::Error` anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ travis-ci = { repository = "tafia/quick-xml" }
 
 [dependencies]
 encoding_rs = { version = "0.8.17", optional = true }
-failure = { version = "0.1.5", optional = true }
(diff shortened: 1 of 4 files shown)


linkmauve:
Note that this is blocking the release of [minidom](https://gitlab.com/xmpp-rs/minidom-rs) 0.15.2 due to our removal of failure [having broken existing code relying on std::error::Error](https://gitlab.com/xmpp-rs/minidom-rs/issues/18).

tafia:
This looks like a very light `Error` impl.  Would you mind adding a static description for all these errors?

andy128k:
@tafia Do you mean to add implementation of `description()` method (which is deprecated)?

tafia:
`description()` is deprecated! Nobody told me!!

Alright then thanks!

Issue #165, opened 2019-08-21
Selectively deserializing XML elements with serde


andreivasiliu:
Hi!

I want to parse a 3MB XML file with serde. I used serde-xml-rs, and it is painfully slow; I hacked my way through serde-xml-rs's code to make it work with quick-xml instead (the APIs are very similar after all), and that sped it up tremendously (from 1.5s to 0.1s).

But! I don't want to deserialize an entire 3MB XML into a giant struct (which has loads of small heap-allocated vecs inside it), when all I want is to scan for a specific element inside it, and deserialize just that one.

I thought of using quick-xml's events to reach the element I want, then read_to_end() to get the whole element as a big text block, and then use serde-xml-rs to parse the text block as xml; except this approach loses all namespacing/encoding info.

I also thought of implementing some sort of from_element(xmlreader, start_element), which would give a partial Deserializer object, which is my current favorite option.

Thoughts? Any better ways to do this?

tafia:
> I hacked my way through serde-xml-rs's code to make it work with quick-xml instead (the APIs are very similar after all), and that sped it up tremendously (from 1.5s to 0.1s).

Awesome!

As for your question I usually deserialize things manually. Usually it is not too complicated to do (I tend to use internal `States` to know in which "level" i'm in). It IS much more boilerplate than using an automatic deserializer. On the other hand it is usually much faster/uses less memory.

Doing automatic serde-like deserialization is very nice too! But getting all the use cases right is complex. If I were you I'd probably open a PR on serde-xml to add a feature to use quick-xml instead of xml-rs. Not sure if they'd want to merge it but you'll probably get some idea about improving your code from there.

Anyway, let me know what you finally decide!

andreivasiliu:
Doesn't look like they'd agree: https://github.com/RReverser/serde-xml-rs/issues/61

I've asked again though: https://github.com/RReverser/serde-xml-rs/issues/108

andreivasiliu:
Meanwhile, just flushing some ideas I had, you can ignore me...

My first thought was to make a read_to_end()-like method that returns an object that implements Deserializer. That meant putting serde code inside quick-xml though, which is not nice. Or maybe it is? Not sure.

My second thought was making read_to_end() return the string it read; looks like inside of it, it keeps clearing the given buffer, but with a flag that could be made optional. Then I could parse that buffer with something else (e.g. serde-xml-rs). Problem is that it's basically being parsed twice, and the buffer might become huge.

[...]

PR #172, opened 2019-11-25
WIP Serde

tafia:
Add a `serialize` feature to allow (de)serializing xmls.

This is hugely inspired by [serde-xml](https://github.com/RReverser/serde-xml-rs) so most of the credit goes to them.

For the moment only Deserialization has been worked on. Serialization should be simpler.

In terms of performance, there is probably some optimization to do in the future. In particular we should be able to avoid the `'static` lifetime in `Deserializer::next` (thus avoiding some unecessary allocations/copy).


tafia:
close #165

jottabyte:
Latest updated worked for me, swapped out all the xml-rs for quick-xml and everything working properly.

jottabyte:
Not sure your opinion on this, but for http://gd2.mlb.com/components/game/mlb/year_2008/month_02/day_26/gid_2008_02_26_fanbbc_phimlb_1/players.xml I'm using this struct:

```rust
struct Team {
    #[serde(rename="type")]
    home_away: HomeAway,
    id: String,
    name: String,
    #[serde(rename="player")]
    players: Vec<Player>,
    #[serde(rename="coach")]
    coaches: Vec<Coach>,
}
```
This returns an Error (missing field 'coach'), since one of the teams doesn't have any coaches in the file. 

Would it be better (or even possible?) to return an empty vec, rather than having to wrap the field in an option.

tafia:
Have you tried adding a `default` on this field?
```rust
    #[serde(rename="coach", default)]
    coaches: Vec<Coach>,
```

jottabyte:
Always something to learn with SerDe! Yes, that is definitely the better way to do it. QuickMXL with SerDe has been working flawlessly on a pretty large sample size.

tafia:
Thank you for the feedbacks, it helps a lot!

Issue #175, opened 2019-12-11
Deserialization Behavior on Unit Struct


davidbarsky:
First of all, congrats on getting the 0.17.1 out with support for serde! I was going to write my own, but I'm glad you there first :). 

Second, I'm trying out this library for (de)serializing data from S3, but I've noticed that S3 tends to have some _odd_ representations of XML response.  For instance, this is the representation of the response to [the `GetBucketTagging` API call](https://docs.aws.amazon.com/en_pv/AmazonS3/latest/API/API_GetBucketTagging.html):

```xml
<GetBucketTaggingOutput>
   <TagSet>
      <Tag>
         <Key>string</Key>
         <Value>string</Value>
      </Tag>
   </TagSet>
</GetBucketTaggingOutput>
```

I wrote the following Rust code to _serialize_ to XML:

```rust
#[derive(Serialize, Deserialize, Debug, PartialEq)]
struct GetBucketTaggingOutput {
    #[serde(rename = "TagSet")]
    pub tag_set: Vec<Tag>,
}

#[derive(Serialize, Deserialize, Debug, PartialEq)]
pub struct Tag {
    /// Name of the tag.
    #[serde(rename = "Key")]
    pub key: String,
    /// Value of the tag.
    #[serde(rename = "Value")]
    pub value: String,
}
```

With the sequence changes I made [in my fork](https://github.com/davidbarsky/quick-xml/), this is able to serialize correctly to:

```xml
<GetBucketTaggingOutput>
    <TagSet>
        <Tag>
             <Key>string</Key>
             <Value>string</Value>
        </Tag>
    </TagSet>
</GetBucketTaggingOutput>
```

However, _deserialization_ fails with the following error:

```
thread 'test_read_tagging' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("missing field `Key`")', src/libcore/result.rs:1165:5.
```

How _should_ this data be represented in Rust, and if it's not representable in Rust with quick-xml, I'd be happy to send a PR that _does_ make it representable!

tafia:
Thanks for trying it out!

First I'd be happy to add the sequence, would you mind opening a PR?

For the deserialization issue, I made it work using an intermediary `TagSet` struct:

```rust
#[test]
fn get_bucket_tagging() {
    let src = " 
<GetBucketTaggingOutput>
   <TagSet>
      <Tag>
         <Key>string</Key>
         <Value>string</Value>
      </Tag>
   </TagSet>
[...]

PR #178, opened 2019-12-15
Add sequence to serializer; add S3 test.

davidbarsky:
A few days late, but this PR is a followup from: https://github.com/tafia/quick-xml/issues/175. This PR introduces:
- a (possibly incorrect!) sequence serializer.
- The test case covered in https://github.com/tafia/quick-xml/issues/175.

### DIFF
--- a/src/se/mod.rs
+++ b/src/se/mod.rs
@@ -2,7 +2,7 @@
 
 mod var;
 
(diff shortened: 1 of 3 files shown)


tafia:
Awesome! Thanks!

PR #183, opened 2020-01-05
Nested Struct and Enum Serialization

dbcfd:
This PR fixes issues with nested struct and enum serialization, where the struct name will be serialized multiple times, or enum name will serialized. Unit variant enums are supported as well.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -121,7 +121,7 @@ use serde::de::{self, DeserializeOwned};
 use serde::forward_to_deserialize_any;
 use std::io::BufRead;
 
-const INNER_VALUE: &str = "$value";
+pub(crate) const INNER_VALUE: &str = "$value";
 
 /// An xml deserializer
 pub struct Deserializer<R: BufRead> {
--- a/src/se/mod.rs
+++ b/src/se/mod.rs
@@ -28,6 +28,7 @@ pub fn to_string<S: Serialize>(value: &S) -> Result<String, DeError> {
 /// A Serializer
 pub struct Serializer<W: Write> {
     writer: Writer<W>,
+    nesting: usize,
 }
 
 impl<W: Write> Serializer<W> {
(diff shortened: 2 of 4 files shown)


dbcfd:
One test is still failing, and it seems to be an issue in the deserializer.

```shell
Serialized: <Nested><A><Nm>Banana</Nm><Src>Store</Src></A><B><Cnt>2</Cnt><Nodes><Boolean>false</Boolean><EOF/></Nodes></B></Nested>

called `Result::unwrap()` on an `Err` value: Xml(EndEventMismatch { expected: "B", found: "Nested" })
thread 'nested_struct' panicked at 'called `Result::unwrap()` on an `Err` value: Xml(EndEventMismatch { expected: "B", found: "Nested" })'```

dbcfd:
Found a couple more issues with added test cases, will be adjusting this branch more.

dave-andersen:
This PR fixes the bug I'm seeing with nested struct names getting duplicated, but it leaves in several debugging println! statements that render it unusable until fixed. :-)

```Attempting to activate pending: Pending=Some("annotation")   Fields=[]
(annotation) Start: Fields=[]
(folder) Pending: Fields=[Some("annotation")]```

Ten0:
Does this fix `flatten` and untagged enums?

dbcfd:
See the test cases added as part of this PR. Without this PR, those test cases will fail. 

I no longer am working on the project with the dataset that led to this PR, so would recommend at least bringing in the test cases from this PR, and improving the implementation to make those tests pass. This PR is one possible implementation, but there's probably a better implementation.

Ten0:
Should we maybe still try to merge it? I'm sure I wouldn't be alone to benefit from the serde support working better :)

Mingun:
Since merging #490 both problems are fixed:
- struct name no more serialized multiple times
- enum name is not serialized. Type names are no more serialized at all

Issue #191, opened 2020-01-25
quick-xml captures UTF BOM as Event::Text


TakaakiFuruse:
Hi, thank you for creating a cool crate, I'm using your crate in my project!

# Problem

It looks quick-xml captures [UTF BOM](https://en.wikipedia.org/wiki/BOM) as an Event::Text.

Normally, If you parse following XML with [the Reader code](https://github.com/tafia/quick-xml#reader),

```xml
<tag1 att1 = "test">HELLO</tag1>
```
you would capture 

1. "test" as Event::Start, 
2. "HELLO" as Event::Text

However, if BOM was embedded in the file,  the Reader code captures

1. "test" as Event::Start, 
2. "\u{feff}" as Event::Text and
3. "HELLO" as Event::Text

The second one should be BOM code.

I think it's better not to capture such code and I happy to make a pull request if you think that should be fixed. 

(Please also have a look at my [branch](https://github.com/TakaakiFuruse/quick-xml/tree/bug_text_with_BOM), I have added a test to parse an XML with BOM. It might be able to reproduce the problem.)

tafia:
Thanks for the issue and the PR.

Out of curiosity, do you have 2 events or just one with the bom unicode before the hello?

At first look, I believe we should not hide the bom when working with events (they are just bytes), like what utf8 advises. The reason being that if you want to write back that xml. you may not want to modify its content if possible.

What should be done instead is probably work on the decoder part and ignore the bom there only.

TakaakiFuruse:
Thanks for your response. 

> Out of curiosity, do you have 2 events or just one with the bom unicode before the hello?

I just tried to make some code to make sure what's going on. If you try following code with Rust 1.41 and quick-xml 0.17.2...

```rust
use quick_xml::events::Event;
use quick_xml::Reader;

fn main() {
    let xml: &str = std::str::from_utf8(b"\xEF\xBB\xBF<tag1 att1 = \"test\">HELLO</tag1>").unwrap();

    let mut reader = Reader::from_str(&xml);
    reader.trim_text(true);

    let mut txt = Vec::new();
    let mut buf = Vec::new();

    loop {
        match reader.read_event(&mut buf) {
            Ok(Event::Start(ref e)) => match e.name() {
                b"tag1" => println!(
                    "Event Start: {:?}",
                    e.attributes().map(|a| a.unwrap().value).collect::<Vec<_>>()
                ),
                _ => (),
            },
[...]

PR #192, opened 2020-01-26
Added procedure to ignore BOM

TakaakiFuruse:
Hi, I just made a Pull Request for #191.

So far, I only added procedure to ignore UTF8 BOM. 
Do you think we need remove UTF16 BOM?

ref: https://docs.rs/encoding_rs/0.8.22/src/encoding_rs/lib.rs.html#2814-2824


tafia:
Thank you for the PR and sorry for the late answer!

I believe checking for bom at each read is probably not efficient. We should check for it only at the beginning of the file.

Also as you have done I don't think this should be the default.

What I propose instead is to have a something like:

```
enum Bom {
    Utf8,
    Utf16LE,
    // ...
}

fn read_bom(buf: &mut Vec<u8>) -> Result<Option<Bom>, Error> {
    // ...
}
```

This way:
- people using xmls with bom will just read the bom once before actually starting processing the xml
- they can use bom information, if needed
- most people won't have any impact


What do you think?

tafia:
Thank you for the PR, please see my general comment on the associated issue.
Open for discussion!

TakaakiFuruse:
@tafia 

So far...

1. I could remove UTF-8 BOM easily. So I would like to discuss whether I am doing right or not.
2. I am having hard time for dealing with UTF-16 (I just decided to start from UTF-16, no reason).

What should I do?

# UTF-8

```rust
    pub fn decode_without_bom<'c>(&self, bytes: &'c [u8]) -> Result<&'c str> {
        if bytes.starts_with(b"\xEF\xBB\xBF") {
            from_utf8(&bytes[3..]).map_err(Error::Utf8)
        } else {
            from_utf8(bytes).map_err(Error::Utf8)
        }
    }
```
I copied the logic from [encoding_rs](https://docs.rs/encoding_rs/0.8.22/src/encoding_rs/lib.rs.html#2967-2978).
This simply removes BOM code. If you see my test code, you might know that it returns empty string. I am not sure this is what you intended. Would it be better to return None or something rather than empty string?

# UTF-16
This is where I am struggling. 

For UTF-16, I decided to usr encoding_rs's ```decode_with_bom_removal```. This is mostly the same as ```decode```.

If you just run ``` cargo test --features encoding``` on my branch, you would see UTF-16 related tests would fail.

This is because ```Reader```'s encoding option was hard-coded as UTF-8. (If I don't pass features option, quick-xml would encode everything as UTF-8 and it returns encoding error for UTF-16 docs.)

```rust
impl<B: BufRead> Reader<B> {
    /// Creates a `Reader` that reads from a reader implementing `BufRead`.
    pub fn from_reader(reader: B) -> Reader<B> {
        Reader {
            reader: reader,
            opened_buffer: Vec::new(),
            opened_starts: Vec::new(),
            tag_state: TagState::Closed,
            expand_empty_elements: false,
            trim_text: false,
            trim_markup_names_in_closing_tags: true,
            check_end_names: true,
            buf_position: 0,
            check_comments: false,
            ns_buffer: NamespaceBufferIndex::default(),
            #[cfg(feature = "encoding")]
            encoding: ::encoding_rs::UTF_8,
        }
    }
```

It's not a good thing, but if you temporary change UTF_8 to UTF_16LE,  it removes UTF-16LE BOM. However, it also regards new line character. (Also, other tests fail, seems like creating new problems...)

src/reader.rs
```diff
@@ -88,6 +88,9 @@ pub struct Reader<B: BufRead> {
     #[cfg(feature = "encoding")]
     /// the encoding specified in the xml, defaults to utf8
     encoding: &'static Encoding,
+    #[cfg(feature = "encoding")]
+    /// check if quick-rs could find out the encoding
```
TakaakiFuruse: Is my documentation correct? If you have any suggestion, I am happy with it.

tests/unit_tests.rs
```diff
@@ -809,12 +809,12 @@ fn test_unescape_and_decode_without_bom_removes_utf16be_bom() {
...
     loop {
         match reader.read_event(&mut buf) {
-            Ok(Event::Text(e)) => txt.push(e.unescape_and_decode_without_bom(&reader).unwrap()),
+            Ok(Event::Text(e)) => txt.push(e.unescape_and_decode_without_bom(&mut reader).unwrap()),
             Ok(Event::Eof) => break,
             _ => (),
         }
     }
-    assert_eq!(txt, vec![""]);
+    assert_eq!(txt[0], "");
```
TakaakiFuruse: Since BOM code will be the first text event, I thought this would be easier to understand.

src/reader.rs
```diff
@@ -715,10 +720,24 @@ impl<B: BufRead> Reader<B> {
...
+            self.is_encoding_set = true;
+            bytes = &bytes[3..];
+        } else if bytes.starts_with(b"\xFF\xFE") {
+            self.is_encoding_set = true;
+            self.encoding = UTF_16LE;
+            bytes = &bytes[2..];
+        } else if bytes.starts_with(b"\xFE\xFF") {
+            self.is_encoding_set = true;
+            self.encoding = UTF_16BE;
+            bytes = &bytes[3..];
```
TakaakiFuruse: Here, bytes will be
```
[  254,    255,    0 ]
```
so I had to make ```&bytes[2..]``` to ```&bytes[3..]```.
(This is strange but I don't know why I had 0 in the end of slice.)

TakaakiFuruse:
@tafia 

I have applied your suggestion and added some code.

I could just add processes for UTF-16LE, UTF-16BE and UTF-8 and not for other encodings like UTF-7, UTF-1 and so on. It is just because encoding-rs [has only few types for UTFs](https://docs.rs/encoding_rs/0.8.22/encoding_rs/index.html#statics).

Although it has [GB18030](https://docs.rs/encoding_rs/0.8.22/encoding_rs/static.GB18030.html) type, I gave up adding it, since I have no idea how I can generate GB18030 file with its BOM code.

src/reader.rs
```diff
@@ -689,6 +694,50 @@ impl<B: BufRead> Reader<B> {
         self.encoding.decode(bytes).0
     }
 
+    /// Decodes a UTF8 slice without BOM (Byte order mark) regarless of XML declaration.
```
efx:
```suggestion
    /// Decodes a UTF8 slice without BOM (Byte order mark) regardless of XML declaration.
```

TakaakiFuruse:
@efx 
Thanks I fixed typo.

tafia:
Thank you!

PR #202, opened 2020-03-16
doc: example of nested parsing

efx:
After some digging I thought it might be helpful to provide an example of how to parse nested XML structures. I created this example for traversing Microsoft Word / ECMA-376 tables.
I totally mimicked rust-syndication/[rss](https://docs.rs/rss/1.9.0/rss/) here.


tafia:
Awesome!

PR #205, opened 2020-03-27
Fix serialization of structs

crusty-dave:
Attributes were being created as child elements.

This works correctly if the simple types that should be attributes are
kept at the beginning of the struct and the more complex types that
should be children are after all of the simple types.

To solve this correctly, there should probably be two Vecs, one for the
attributes and one for the children.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quick-xml"
-version = "0.17.2"
+version = "0.17.3"
 authors = ["Johann Tuffe <tafia973@gmail.com>"]
 description = "High performance xml reader and writer"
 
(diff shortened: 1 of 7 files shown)


tafia:
Thanks for the PR!!
I'm reviewing it, there are some points I don't really like but I don't have better alternative so far. Thank you for the time you spent on that I know serde related code is not the simplest one to look at.

crusty-dave:
Note, to solve this correctly, there should probably be two Vecs, one for the attributes and one for the children.

serialize_field() would need to collect the Vecs, then end() would serialize the result.

The problem with that is Serialize cannot be used to create a trait object.

If someone wanted to integrate the erased-serde crate, they could probably make that work.

spikespaz:
I wish I had the knowledge to merge this. It's too old.

Mingun:
Since merging #490 the attributes should be marked explicitly in structs, by prepending `@` character to name of field (this is also work for internally and adjacently tagged enum representations).

PR #210, opened 2020-05-03
Make escape module public apart from EscapeError

frankier:
I think these functions should be public. My own usage of unescape is:

1. Search for particular two markers in a text node
2. If they are present, use them as delimiters to make a slice which is passed to unescape
3. Otherwise, don't use unescape.

This way a user can only unescape the text they need.


tafia:
Thanks

PR #213, opened 2020-05-24
Added Clone for Reader

Dgame:
I think it might be useful if that were possible.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -15,6 +15,7 @@ use events::{attributes::Attribute, BytesDecl, BytesEnd, BytesStart, BytesText, 
 
 use memchr;
 
+#[derive(Clone)]
 enum TagState {
     Opened,
     Closed,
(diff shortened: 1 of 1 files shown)


tafia:
Why would you need to clone a reader? It requires the underlying `BufReader` to be clone too.
I don't mind merging it but I'm curious about the motivation, it may not be the best thing to do.

Dgame:
I'm trying to deserialize a `Result<T, E>`. So I first try to deserialize **T**, but if that doesn't work I try to deserialize **E**, but to do so I have to reproduce the state before I tried to deserialize **T**. An example would be `Result<f32, String>`, where the xml can look like one of the following XML snippets: `<approx>3.14</approx>` and `<approx>That is not a float</approx>`.

tafia:
I believe you should probably implement your own deserializer in that case.
See https://serde.rs/custom-date-format.html for instance.

Something like 
```rust
    pub fn deserialize_approx<'de, D>(deserializer: D) -> Result<f64, D::Error>
    where D: Deserializer<'de>,
    {
        let s = String::deserialize(deserializer)?;
        s.parse().map_err(serde::de::Error::custom)
    }

    #[derive(Deserialize)]
    struct S {
         #[serde(deserialize_with="deserialize_approx")]
         approx: f64,
    }
```

PR #219, opened 2020-06-22
Fix underflow on shrink

lilianmoraru:
Would like to also create a UT for this but I am very busy for the next 3 weeks.
I might reduce the bug later and make another PR with the UT that reproduces the bug.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -197,6 +197,9 @@ impl Indentation {
     }
 
     fn shrink(&mut self) {
-        self.indents_len = self.indents_len - self.indent_size;
+        self.indents_len = match self.indents_len.checked_sub(self.indent_size) {
+            Some(result) => result,
+            None => 0,
+        };
     }
 }


lilianmoraru:
Without displaying to much info, here is a reduced stack backtrace:
```
thread '<unnamed>' panicked at 'attempt to subtract with overflow'
stack backtrace:
...
  12:     0x7ff6e38cfe7c - core::panicking::panic
                               at /rustc/49cae55760da0a43428eba73abcb659bb70cf2e4/src/libcore/panicking.rs:54
  13:     0x7ff6e388107d - quick_xml::writer::Indentation::shrink
                               at .../github.com-1ecc6299db9ec823/quick-xml-0.18.1/src/writer.rs:178
  14:     0x7ff6e37f641a - quick_xml::writer::Writer<std::io::cursor::Cursor<alloc::vec::Vec<u8>>>::write_event<std::io::cursor::Cursor<alloc::vec::Vec<u8>>,quick_xml::events::Event>
                               at .../github.com-1ecc6299db9ec823/quick-xml-0.18.1/src/writer.rs:106
```

lilianmoraru:
Introduces regression, not the correct fix.

lilianmoraru:
Sorry, nvm, the fix is ok, the parsing is incorrect in `master` also, when using `ident_size: 1`.

lilianmoraru:
`ident_size: 1` doesn't play well with `reader.trim_text(true);` but that's a separate issue.

tafia:
Thanks!

PR #220, opened 2020-06-25
use write_all instead of write

pacman82:
Hello, great library! Recently I had an issue, though. I've used the library together with a writer compressing into `.gz`. The output had bytes randomly dropped. Tracked the issue down to the use of `write` instead of `write_all`. Hope you are ok with the implied interface change of the writer.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -91,7 +91,7 @@ impl<W: Write> Writer<W> {
     }
 
     /// Writes the given event to the underlying writer.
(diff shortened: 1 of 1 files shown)


markus-klein-by:
@tafia  Can I do something to move this forward?

tafia:
Sorry for the late answer.
While I don't really like the fact that the api is changing I think this is indeed a bug currently.
Thank you!

Issue #196, opened 2020-02-15
Loosing attributes during serialization


mlevkov:
Hello,

I have an XML with some elements that have attributes. It deserializes into a struct ok, but when I serialize to output, the attributes are no longer attributes, but rather individual elements. How do I correctly indicate that attributes need to stay as attributes and nothing else? 

For example, I have the following element with attributes:
```xml
<Representation audioSamplingRate="48000" bandwidth="63700" codecs="mp4a.40.2" id="519678732693145249_AO-00-00-00-71">
```

When I read the element into the struct:
```rust
#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
struct Representation{
    #[serde(rename = "audioSamplingRate")]
    a_audio_sampling_rate: Option<String>,
    #[serde(rename = "bandwidth")]
    a_bandwidth: String,
    #[serde(rename = "codecs")]
    a_codecs: String,
    #[serde(rename = "id")]
    a_id: String,
}
```
It looks and feels right. However, during the output, the serialization step, I get the output that looks like this:
```xml
<Representation>
   <audioSamplingRate>48000</audioSamplingRate>
   <bandwidth>63700</bandwidth>
   <codecs>mp4a.40.2</codecs>
   <id>519678732693145249_AO-00-00-00-71</id>
</Representation>
```
What am I not doing right? Or, what am I missing in this process?
I'm reading the element indicated above, as a string. Also, the output is a string as well.

Thank you,
Max

tafia:
Hmm you're right. I see no obvious solution right now, apart from serializing it manually.

mlevkov:
@tafia If it was to be serialized manually, how would one go about doing so? Could a sample be constructed that demonstrates such?

crusty-dave:
I am having the same issue.

The solution would seem to be to fix serialize_field() in following code:

```
impl<'w, W> ser::SerializeStruct for Struct<'w, W>
where
    W: 'w + Write,
{
    type Ok = ();
    type Error = DeError;

    fn serialize_field<T: ?Sized + Serialize>(
        &mut self,
        key: &'static str,
        value: &T,
    ) -> Result<(), DeError> {
        let key = key.as_bytes();
        self.parent
            .writer
            .write_event(Event::Start(BytesStart::borrowed_name(key)))?;
        value.serialize(&mut *self.parent)?;
        self.parent
            .writer
[...]

PR #221, opened 2020-06-27
Serialize basic type as attributes when using serde

huaoguo:
As [this issue](https://github.com/tafia/quick-xml/issues/196) said, basic type struct fields should be serialized to attributes.

Rather than check type name, I check the serialization result of the field. When the result is a tag(starting with '<'), I treat it as a child node.

In serialize_field(), I don't serialize the field to final result, but store attrs and child nodes in separate temporary location. After all fields have been collected, in end(), I will write them to final result. So the field order is not important.

I got `test framework quit unexpectedly` while runing test in serde_roundtrip.rs, so I write tests in another project. Here is my tests:

```rust
use serde::*;
use quick_xml::se::to_string;
use regex::Regex;
use std::borrow::Cow;

#[derive(Serialize)]
#[serde(rename="classroom")]
struct Classroom {
    pub students: Vec<Person>,
    pub number: String,
    pub adviser: Person
}

#[derive(Serialize)]
#[serde(rename="person")]
struct Person {
    pub name: String,
    pub age: u32
}

#[derive(Serialize)]
#[serde(rename="empty")]
struct Empty {}

#[test]
fn test_nested() {
    let s1 = Person { name: "sherlock".to_string(), age: 20 };
    let s2 = Person { name: "harry".to_string(), age: 19 };
    let t = Person { name: "albus".to_string(), age: 88 };
    let doc = Classroom { students: vec![s1, s2], number: "3-1".to_string(), adviser: t };
    let xml = quick_xml::se::to_string(&doc).unwrap();

    let str = r#"<classroom number="3-1">
                        <students>
                          <person name="sherlock" age="20"></person>
                          <person name="harry" age="19"></person>
                        </students>
                        <adviser>
                          <person name="albus" age="88"></person>
                        </adviser>
                      </classroom>"#;
    assert_eq!(xml, inline(str));
}

fn inline(str: &str) -> Cow<str> {
    let regex = Regex::new(r">\s+<").unwrap();
    regex.replace_all(str, "><")
}

#[test]
fn test_empty() {
    let e = Empty {};
    let xml = to_string(&e).unwrap();
    assert_eq!(xml, "<empty></empty>");
}
```

### DIFF
--- a/src/se/mod.rs
+++ b/src/se/mod.rs
@@ -209,8 +209,9 @@ impl<'w, W: Write> ser::Serializer for &'w mut Serializer<W> {
         name: &'static str,
         _len: usize,
     ) -> Result<Self::SerializeStruct, DeError> {
+        // 这里只写入 '<name', 后面在序列化 field 时写入 '>'
         self.writer
-            .write_event(Event::Start(BytesStart::borrowed_name(name.as_bytes())))?;
+            .write(format!("<{}", name).as_bytes());
         Ok(Struct::new(self, name))
     }
 
--- a/src/se/mod.rs
+++ b/src/se/mod.rs
@@ -211,7 +211,7 @@ impl<'w, W: Write> ser::Serializer for &'w mut Serializer<W> {
     ) -> Result<Self::SerializeStruct, DeError> {
         // 这里只写入 '<name', 后面在序列化 field 时写入 '>'
         self.writer
-            .write(format!("<{}", name).as_bytes());
+            .write(format!("<{}", name).as_bytes())?;
         Ok(Struct::new(self, name))
     }
 
--- a/src/se/mod.rs
+++ b/src/se/mod.rs
@@ -272,7 +272,7 @@ mod tests {
             struct_ser.serialize_field("foo", "bar").unwrap();
             let attrs = std::str::from_utf8(&struct_ser.attrs).unwrap();
             assert_eq!(attrs, " foo=\"bar\"");
-            struct_ser.end().unwrap();
+            let _ = struct_ser.end().unwrap();
         }
 
         let got = String::from_utf8(buffer).unwrap();
--- a/src/se/mod.rs
+++ b/src/se/mod.rs
@@ -209,9 +209,9 @@ impl<'w, W: Write> ser::Serializer for &'w mut Serializer<W> {
         name: &'static str,
         _len: usize,
     ) -> Result<Self::SerializeStruct, DeError> {
-        // 这里只写入 '<name', 后面在序列化 field 时写入 '>'
-        self.writer
-            .write(format!("<{}", name).as_bytes())?;
+        write!(self.writer.inner(), "<{}", name).map_err(|err| {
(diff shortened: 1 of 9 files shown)


src/se/mod.rs
```diff
@@ -209,8 +209,9 @@ impl<'w, W: Write> ser::Serializer for &'w mut Serializer<W> {
         name: &'static str,
         _len: usize,
     ) -> Result<Self::SerializeStruct, DeError> {
+        // 这里只写入 '<name', 后面在序列化 field 时写入 '>'
```
tafia, line 212: Sorry could you please translate in english?

src/se/var.rs
```diff
@@ -96,18 +99,43 @@ where
...
-        value.serialize(&mut *self.parent)?;
-        self.parent
-            .writer
-            .write_event(Event::End(BytesEnd::borrowed(key)))?;
+        // 先序列化到一个独立的地方
+        let mut serializer = Serializer::new(&mut self.buffer);
+        value.serialize(&mut serializer)?;
+
+        // None 序列化后的值长度为 0
+        if self.buffer.len() > 0 {
```
tafia, line 107:
```suggestion
        if !self.is_empty() {
```

src/se/mod.rs
```diff
@@ -209,8 +209,9 @@ impl<'w, W: Write> ser::Serializer for &'w mut Serializer<W> {
...
         self.writer
-            .write_event(Event::Start(BytesStart::borrowed_name(name.as_bytes())))?;
+            .write(format!("<{}", name).as_bytes())?;
```
tafia, line 214:
```rust
write!(self.writer.inner(), "<{}", name)?
```

src/se/var.rs
```diff
@@ -96,18 +99,43 @@ where
...
+            // 如果是复杂类型，必定以 < 打头
+            // 复杂类型当做子节点处理，简单类型当做属性处理
+            if self.buffer[0] == b'<' {
+                self.children.extend(format!("<{}>", key).bytes());
```
tafia, line 111:
```suggestion
                write!(&mut self.children, "<{}>", key)?;
```

src/se/var.rs
```diff
@@ -96,18 +99,43 @@ where
...
+                self.children.extend(&self.buffer);
+                self.children.extend(format!("</{}>", key).bytes());
```
tafia, line 113:
```suggestion
                write!(&mut self.children, "</{}>", key)?;
```

huaoguo:
@tafia resolved

mlevkov:
@tafia Would you be kind to please review this?

tafia:
Thank you!

faldzip:
And what if I don't want this behavior? Or even use it selectively to different fields in struct?

yukihane:
[Unflattening structs into verbose XML](https://github.com/tafia/quick-xml#unflattening-structs-into-verbose-xml)  would be usable to prevent this behavior.

PR #223, opened 2020-07-01
Fix minor formatting issues Travis has been complaining about

AndreasKarg:
Like the title says, Travis builds appear to have been failing for a little while - but only due to `cargo fmt` objecting to some very minor things.

I have taken the liberty of running `cargo fmt` locally and pushing the end result into a branch, making the build go green again.


tafia:
Thanks!

PR #232, opened 2020-09-20
Deserialize bytes

elrnv:
This PR ensures that there is a path for deserializing raw bytes.
There is a real use case for this type of deserialization (see for example the VTK XML file format using raw unencoded binary data).

A new internal method on BytesText is added to potentially make use of an owned byte Vec.

A test is added to demonstrate how this feature can be used.


tafia:
Thanks!

PR #233, opened 2020-09-23
Allow using tokio's AsyncBufRead

endor:
I would like to use `quick-xml` together with `tokio`'s `AsyncRead`, as I am fetching an object from S3 with `rusoto_s3` and that returns a body that implements `AsyncRead`. This is an attempt at making that happen.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ travis-ci = { repository = "tafia/quick-xml" }
 [dependencies]
 async-recursion = { version = "0.3.2", optional = true }
 encoding_rs = { version = "0.8.26", optional = true }
-tokio = { version = "0.2.22", features = ["fs", "io-util"], optional = true }
+tokio = { version = "1.4.0", features = ["fs", "io-util"], optional = true }
 serde = { version = "1.0", optional = true }
 memchr = "2.3.4"
 
@@ -26,7 +26,7 @@ memchr = "2.3.4"
 serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
 regex = "1"
-tokio = { version = "0.2.22", features = ["macros", "rt-threaded"] }
+tokio = { version = "1.4.0", features = ["macros", "rt-multi-thread"] }
 
 [lib]
 bench = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "quick-xml"
 version = "0.22.0"
 authors = ["Johann Tuffe <tafia973@gmail.com>"]
 description = "High performance xml reader and writer"
+edition = "2018"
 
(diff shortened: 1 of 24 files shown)


endor:
The unit tests in `unit_tests.rs` are now working with and without `async`.

endor:
Looks like the `Deserializer` would need to support async as well.

endor:
I realize that this is not the best solution, because there's a lot of duplication. I'm just not sure how else to do it. I don't have a lot of experience with async code in Rust.

* If we were to for example wrap a sync `Reader` in some kind of wrapper to make it async, so they could share more code, we would force the user to use a runtime, I guess.
* If we were to split the async part into a separate crate/library than any changes made to one would need to manually be also made to the other.
* Maybe we could share a bit more code if the methods that are sync in the async `Reader` would be put into a trait or something, but a lot of the methods are async in the async `Reader` and they cannot be shared.

endor:
After some feedback, I've changed the design a bit now and introduced an `AsyncReader` instead of overloading the `Reader`. This also allows me to leave the `Deserializer` as is for now and all the tests pass. There's obviously still duplication but it seems like a cleaner abstraction.

endor:
@tafia Any thoughts on this?

tafia:
Ouch, it's been really long that this issue is opened, really sorry for that.
I love the idea, I need to get into it a little more and update it with (now released) tokio 1.0. Thank you so much and very sorry for the late answer.

endor:
I rebased the PR now on current `master`. Next step is to update `tokio`. It seems like there's a broken test in `master`.

endor:
`tokio` is updated.

endor:
I also fixed the one unit test failing in `master`.

endor:
ping @tafia

agentsim:
@tafia Any update on whether or not this can get merged?

tafia:
Sorry, this is definitely something I consider, I'll have a look in depth today.

agentsim:
That would be great, thanks! For what it's worth, I've been using this branch and haven't run into any problems.

itsgreggreg:
Also FWIW, I'm writing an XMPP server and am interested in this functionality so I can switch to async multi-client handling. Happy to help push this along in any capacity. Fix conflicts, tests, etc. Lemme know if I can help!

tafia:
@itsgreggreg I'd appreciate a rebase, thanks!

endor:
I caught up on some of the changes that were recently made, but the `BufferedInput` change takes a bit more work. I might get to it soon.

erikpols:
any updates on this? or tips how to get this to work?

dralley:
Closing this old PR in favor of https://github.com/tafia/quick-xml/pull/417

PR #237, opened 2020-10-11
Always capture "$value" key when text node is appeared in the map

### LINKED ISSUE #185: Serde support needs to be revisited
PR #183 fixes some of this behavior with the existing interactions of serde and reader, but overall, the implementation needs to be revisited.

Here's the issues I've encountered so far
- Inner enums without rename will fail to deserialize
- Inner enums with `rename` attribute will fail to deserialize unless the rename is `$value`
- `flatten` attribute does not work
- Vector of enums will fail to deserialize, regardless of rename usage
- Nested containers without rename of `$value` will fail to deserialize

I've been trying to make things work "better" in my PR, but the rest of these issues would take a significant overhaul of the code.

### LINKED ISSUE #231: deserializing xml to serde_json::Value doesn't work
I am trying to convert xml to json by serializing xml to serde_json::Value with quick xml but it doesn't work. I have this code:
```rust
let value: serde_json::Value = quick_xml::de::from_str("<root><val>15</val></root>").unwrap();
let json_body = serde_json::to_string(&value).unwrap();
println!("{}", json_body);
```
It prints: `{"val":{}}`.
I would expect it to print `{"val":15}` or `{"val":"15"}`.

code is [here](https://github.com/ozgrakkurt/quick_xml_json/blob/master/src/main.rs).

Mingun:
Fixes #185, fixes #231

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ memchr = "2.3.3"
 
 [dev-dependencies]
 serde = { version = "1.0", features = ["derive"] }
+serde-value = "0.7"
 regex = "1"
 
 [lib]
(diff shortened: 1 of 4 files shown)


tafia:
Thanks!

PR #239, opened 2020-10-28
fix: Unescape all existing HTML entities

### LINKED ISSUE #238: Is this crate aiming to parse HTML?
I've successfully used quick-xml to parse HTML, however I just noticed that quick-xml does not unescape things like `&nbsp;`. So I am not entirely sure if quick-xml is generally supposed to handle HTML or only partially such that we can build our own parsers/unescapers on top.

untitaker:
Fix #238


untitaker:
I can't reliably run rustfmt on this codebase without reformatting everything.

tafia:
That's awesome thanks!
Merging it but I'm wondering if it wouldn't be better to use some `HashMap` now that the list is so big.

tafia:
Following issue #255 this will now be behind a `'escape-html'` feature.

PR #244, opened 2020-12-07
Parse & in cdata correctly

### LINKED ISSUE #184: `escape::unescape` treats any non-entity character sequence `&whatever;` as an error
`escape::unescape` tries to recognize HTML entities so hard that it simply fails on non-entity character sequence that contains `&` and `;`.
Simple case:
```xml
<a>a&b;</a>
```
causes an error `Xml(EscapeError(UnrecognizedSymbol(2..3, Ok("b"))))`.

Should not it ignore such sequences and leave them as is?

seanyoung:
The BytesText buffer should always contain escaped string, so escape
anything in cdata, rather than assume it is already escaped.

This should fix #184 

Signed-off-by: Sean Young <sean@mess.org>

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -371,7 +371,7 @@ impl<B: BufRead> Reader<B> {
                             Err(e) => return Err(e),
                         }
                     }
-                    Ok(Event::CData(BytesText::from_escaped(
+                    Ok(Event::CData(BytesText::from_plain(
                         &buf[buf_start + 8..buf.len() - 2],
                     )))
                 }
--- a/tests/serde_roundtrip.rs
+++ b/tests/serde_roundtrip.rs
@@ -123,3 +123,43 @@ fn no_contiguous_fields() {
     // let serialized = to_string(&xml).unwrap();
     // assert_eq!(serialized, source);
 }
+
+#[test]
+fn escapes_in_cdata() {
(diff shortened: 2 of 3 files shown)


seanyoung:
See https://www.w3.org/TR/xml/#dt-cdsection

```
Within a CDATA section, only the CDEnd string is recognized as markup, so that left angle brackets
and ampersands may occur in their literal form; they need not (and cannot) be escaped using
" &lt; " and " &amp; ". CDATA sections cannot nest.
```

seanyoung:
```
$ echo '<a><![CDATA[&]]></a>' | xmllint -
<?xml version="1.0"?>
<a><![CDATA[&]]></a>
```

seanyoung:
Some more evidence:
```
$ python
Python 3.9.0 (default, Oct  6 2020, 00:00:00) 
[GCC 10.2.1 20200826 (Red Hat 10.2.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import xml.etree.ElementTree as ET
>>> tree = ET.fromstring('<a><![CDATA[&]]></a>')
>>> tree.text
'&'
>>> tree = ET.fromstring('<a><![CDATA[&amp;]]></a>')
>>> tree.text
'&amp;'
```

tafia:
Thanks!

PR #250, opened 2021-01-03
serialize a key as <key/> instead of aborting as an error

yijunyu:
I work on Huawei's Trusted Programming Language project, which was recently asked to convert Rust source code files into XML format. We believe that quick-xml can help for this purpose. However, it turns out that saving key-only is considered Unsupported and the serializer aborts the execution. Hence, I've completed a working prototype for handle this special case.

If someone with more experience developing in `quick-xml` can give this an early look and send feedback, that would be really helpful. Apologies in advance for the current lack of tests and limited documentation. I'll be working on this full time for a while, so it will improve rapidly.

Thanks!

### DIFF
--- a/src/se/var.rs
+++ b/src/se/var.rs
@@ -33,12 +33,14 @@ where
     type Error = DeError;
 
     fn serialize_key<T: ?Sized + Serialize>(&mut self, key: &T) -> Result<(), DeError> {
-        // Err(DeError::Unsupported(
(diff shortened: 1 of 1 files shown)


tafia:
Hi!, Sorry for the late answer, 
Would you mind putting this behind a feature flag as it may break some existing code.

tafia:
Merging it because I don't think it actually breaks any code except if they were actually relying on the error itself!

Issue #249, opened 2021-01-02
$value serialization does not work


amellensatz:
This simple example program

```rust
use serde::{Serialize, Deserialize};
use quick_xml::{de::from_str, se::to_string};

#[derive(Serialize, Deserialize, Debug)]
struct X {
    a: i64,
    #[serde(rename = "$value")]
    b: String,
}

fn main() {
    let d: X = from_str(r#"<X a="3">body</X>"#).unwrap();
    println!("{:?}", d);
    println!("{}", to_string(&d).unwrap());
}
```

outputs

```
X { a: 3, b: "body" }
<X a="3" $value="body"/>
```

but `<X a="3">body</X>` ought to be output instead.

PR #251, opened 2021-01-04
#249 `$value` rename should work the same for serialization too

vitaly-m:
#249 `$value` serialisation should work the same for deserialization and serialization


tafia:
Thanks!

PR #253, opened 2021-02-03
Split text trim into start and end

elrnv:
I am using the wonderful quick-xml crate to read [VTK](https://vtk.org/) files. Some of those files are stored as xmls with data stored in binary format inside an xml element. Since this is not ascii, some trailing whitespace can correspond to valid bytes, which are trimmed by default. This PR allows users to only trim the starting whitespace instead of trimming from both sides.
All behaviour is left as is but an additional api call allows users to disable trimming trailing spaces `trim_text_end(false)`. When combined with a preceding `trim_text(true)` this will just trim the beginning of each text event.

tl;dr: This allows XMLs with binary blobs to work correctly.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -72,8 +72,10 @@ pub struct Reader<B: BufRead> {
     tag_state: TagState,
     /// expand empty element into an opening and closing element
     expand_empty_elements: bool,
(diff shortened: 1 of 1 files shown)


tafia:
Thanks

PR #256, opened 2021-02-04
Put escaping html entities behind a 'escape-html' feature

### LINKED ISSUE #255: Builds of 0.21 are much slower, wasm builds cause massive memory use in rustc
This was probably caused by https://github.com/tafia/quick-xml/pull/239/files#diff-0acc5298e8580ddde57f322c2f6b70406f8d2b13acd0bebc4125428a66afc585 .

Reproduction:
```
git clone https://github.com/tafia/quick-xml.git
cd quick-xml
git checkout 1961d8e
```
```
cargo build --release
```
Builds in 40s (2s before this commit)
```
cargo build --release --target=wasm32-unknown-unknown
```
After 1-2 minutes rustc's memory use starts increasing to 4+ GB, causing my machine to start swapping. (builds in 2s before this commit)

You can also see the effect on GH Actions workers - a PR that bumped quick-xml dependency seems to have significantly increased build times (by 6-7min) on some configurations.

Compare:
https://github.com/ruffle-rs/ruffle/runs/1825148297
https://github.com/ruffle-rs/ruffle/runs/1825150290

tafia:
Close #255

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ bench = false
 default = []
 encoding = ["encoding_rs"]
 serialize = ["serde"]
+escape-html = []
 
 [package.metadata.docs.rs]
 features = ["serialize"]
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,9 @@
 
 ## Unreleased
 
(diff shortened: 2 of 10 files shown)


tafia:
The PR is polluted by rustfmt, you can always check commits by commit.

PR #259, opened 2021-02-06
Fix Windows test cases

ajtribick:
A few of the test cases were failing on Windows because of the line-ending handling. This fixes the tests so they pass, but maybe the `Writer` should be producing `\r\n` on Windows instead?

### DIFF
--- a/tests/documents/html5-windows.txt
+++ b/tests/documents/html5-windows.txt
@@ -1,0 +1,8 @@
+DocType( html)
+Characters(
+)
+StartElement(a, attr-error: error while parsing attribute at position 7: Attribute value must start with a quote.)
+Characters(Hey)
+EndElement(a)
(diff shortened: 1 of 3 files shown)


tafia:
Thanks!
Indeed writing `\r\n` is probably a better default on windows,

Mingun:
@ajtribick , why you expect that on Windows we get an `InvalidUtf8` error? Right now the `html5` test for Linux is passed on my Windows 10:
https://github.com/tafia/quick-xml/blob/9120ec26747300c420153bc2faeccb9e63c1da05/tests/xmlrs_reader_tests.rs#L43-L60

ajtribick:
@Mingun - there have been a fair number of changes to master since this pull request was merged, it may well be that this is no longer necessary.

PR #261, opened 2021-02-08
add `*_with_custom_entities` versions of all `unescape_*\ methods

pchampin:
these versions expect an additional parameter:
a hashmap containing custom entity definitions.
A typical use-case is for the user to parse those entities definitions
from the DOCTYPE.
An example is provided: examples/custom_entities.rs .

### DIFF
--- a/src/escapei.rs
+++ b/src/escapei.rs
@@ -147,7 +147,7 @@ fn do_unescape_with<'a>(
                     b"amp" => unescaped.push(b'&'),
                     b"apos" => unescaped.push(b'\''),
                     b"quot" => unescaped.push(b'\"'),
-                    bytes if bytes[0] == b'#' => {
+                    bytes if bytes.starts_with(b"#") => {
                         let bytes = &bytes[1..];
                         let code = if bytes.starts_with(b"x") {
                             parse_hexadecimal(&bytes[1..])
@@ -5556,7 +5556,7 @@ fn do_unescape_with<'a>(
                         unescaped.push(b'\x1D');
                         unescaped.push(b'\x56');
                     }
-                    bytes if bytes[0] == b'#' => {
+                    bytes if bytes.starts_with(b"#") => {
                         let bytes = &bytes[1..];
                         let code = if bytes.starts_with(b"x") {
                             parse_hexadecimal(&bytes[1..])
--- a/src/escapei.rs
+++ b/src/escapei.rs
@@ -2,6 +2,7 @@
 
 use memchr;
 use std::borrow::Cow;
+use std::collections::HashMap;
 
 #[derive(Debug)]
 pub enum EscapeError {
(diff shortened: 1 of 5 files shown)


src/escapei.rs
```diff
@@ -128,22 +147,27 @@ pub fn unescape(raw: &[u8]) -> Result<Cow<[u8]>, EscapeError> {
                     b"amp" => unescaped.push(b'&'),
                     b"apos" => unescaped.push(b'\''),
                     b"quot" => unescaped.push(b'\"'),
-                    bytes => {
-                        let code = if bytes.starts_with(b"#x") {
-                            parse_hexadecimal(&bytes[2..])
-                        } else if bytes.starts_with(b"#") {
-                            parse_decimal(&bytes[1..])
+                    bytes if bytes[0] == b'#' => {
```
tafia, line 150:
```rs
bytes if bytes.starts_with(b'#')
```
(to accommodate when bytes is empty)

src/escapei.rs
```diff
@@ -5532,22 +5556,27 @@ pub fn unescape(raw: &[u8]) -> Result<Cow<[u8]>, EscapeError> {
                         unescaped.push(b'\x1D');
                         unescaped.push(b'\x56');
                     }
-                    bytes => {
-                        let code = if bytes.starts_with(b"#x") {
-                            parse_hexadecimal(&bytes[2..])
-                        } else if bytes.starts_with(b"#") {
-                            parse_decimal(&bytes[1..])
+                    bytes if bytes[0] == b'#' => {
```
tafia, line 5559: same comment

src/events/attributes.rs
```diff
@@ -130,6 +149,50 @@ impl<'a> Attribute<'a> {
...
+    #[cfg(feature = "encoding")]
+    pub fn unescape_and_decode_value_with_custom_entities<B: BufRead>(
+        &self,
+        reader: &Reader<B>,
+        custom_entities: &HashMap<Vec<u8>, Vec<u8>>,
+    ) -> Result<String> {
+        let decoded = reader.decode(&*self.value);
+        let unescaped =
+            unescape_with(decoded.as_bytes(), custom_entities).map_err(Error::EscapeError)?;
+        String::from_utf8(unescaped.into_owned()).map_err(|e| Error::Utf8(e.utf8_error()))
```
tafia, line 171: Are we sure it'll be utf8? Shouldn't we use the `reader.decode` instead?

tafia (changes requested):
Thanks!
I've made a few comments. Also I feel like there is a lot of code repeat. Not a big deal but if you find a way to make it a little more DRY would be much appreciated!

src/escapei.rs
pchampin: absolutely!

src/events/attributes.rs
pchampin: I lean towards keeping things simple: what about requiring that `custom_entities` contains UTF-8?

pchampin: About making the code more DRY: yes, I would be better, but I wanted to test the idea with you before refactoring... I'll go make the change (as well as your comments above).

pchampin:
About making the code more DRY: yes, I would be better, but I wanted to test the idea with you before refactoring... I'll go make the change (as well as your comments above).

src/events/attributes.rs
tafia: Alright!

tafia (approved):
This is great, thanks!

PR #271, opened 2021-03-02
fix #270

### LINKED ISSUE #270: Unescaping broken for codepoints > 127
The do_unescape function pushes the unicode codepoints in BE order into the utf-8 stream. This is not correct.

s3bk:
write out the correct codepoints


tafia:
Thanks!

I have 2 questions:
- why not using `push_utf8` directly? It looks slightly less expensive than `std::char::encode_utf8` (no need to use an intermediary 4 bytes buffer. (It is very likely that it is optimized away) 
- Ideally we should hard code the utf8 encoded characters in all the match branches.

s3bk:
Yes, ideally the match should have a pre-computed slice each.
I suppose your push_utf8 could be turned into a const fn. Lets see…

s3bk:
I did not turn it into a `const fn`, however …

The majority of the time will be spent looking up the name (which is hopefully faster now).
Encoding one char as utf8 using the method in std is probably not even measurable and it avoids duplicating the functionality.

s3bk:
The alternative would have been to return slices from the match …

s3bk:
@tafia I think this is now a pretty good solution.
`char::encode_utf8` is only called in the case of `&#1234;`.

tafia:
Thanks!!

PR #278, opened 2021-03-20
Basic implementation of the ElementWriter concept

dralley:
@tafia Here's a basic implementation of the idea. See the tests I added for examples of how using it would look.

I can add documentation and improved tests once the general idea is solidified.  There's probably things to be improved :)

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -92,7 +92,7 @@ impl<'a> BytesStart<'a> {
     ///
     /// # Example
     ///
-    /// ```
+    /// ```rust
     /// # use quick_xml::{Error, Writer};
     /// use quick_xml::events::{BytesStart, Event};
     ///
@@ -802,7 +802,7 @@ pub enum Event<'a> {
 
 impl<'a> Event<'a> {
     /// Converts the event to an owned version, untied to the lifetime of
-    /// buffer used when reading but incurring a new, seperate allocation.
+    /// buffer used when reading but incurring a new, separate allocation.
     pub fn into_owned(self) -> Event<'static> {
         match self {
             Event::Start(e) => Event::Start(e.into_owned()),
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -315,10 +315,13 @@ impl<'a> BytesStart<'a> {
     }
 
     /// Try to get an attribute
-    pub fn try_get_attribute(&'a self, attr_name: &[u8]) -> Result<Option<Attribute<'a>>> {
+    pub fn try_get_attribute<N: AsRef<[u8]> + Sized>(
+        &'a self,
+        attr_name: N,
+    ) -> Result<Option<Attribute<'a>>> {
         for a in self.attributes() {
             let a = a?;
-            if a.key == attr_name {
+            if a.key == attr_name.as_ref() {
                 return Ok(Some(a));
             }
         }
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -1,6 +1,7 @@
 //! A module to handle `Writer`
 
-use crate::{errors::Error, errors::Result, events::BytesEnd, events::BytesStart, events::Event};
+use crate::errors::{Error, Result};
+use crate::events::{attributes::Attribute, BytesStart, BytesText, Event};
 use std::io::Write;
 
 /// XML writer.
@@ -10,7 +11,6 @@ use std::io::Write;
 /// # Examples
 ///
 /// ```rust
-/// # extern crate quick_xml;
 /// # fn main() {
 /// use quick_xml::{Reader, Writer};
 /// use quick_xml::events::{Event, BytesEnd, BytesStart};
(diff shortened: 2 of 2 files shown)


src/writer.rs
```diff
@@ -169,19 +170,41 @@ impl<W: Write> Writer<W> {
...
+            writer: self,
+            tag_name: name.as_ref(),
+            start_tag: BytesStart::borrowed_name(name.as_ref()),
+        }
+    }
+}
+
+pub struct NodeWriter<'a, W: Write> {
+    writer: &'a mut Writer<W>,
+    tag_name: &'a [u8],
```
dralley, line 185: As I mentioned, I'm kind of a novice with lifetimes, but I couldn't figure out how to re-use the value from ```start_tag``` without compile errors.

src/writer.rs
```diff
@@ -394,4 +417,39 @@ mod indentation {
...
+        writer
+            .create_node(b"paired")
+            .with_attribute(("attr1", "value1"))
+            .with_attribute(("attr2", "value2"))
+            .write_text(BytesText::from_plain_str("text"))
+            .expect("failure");
+
+        assert_eq!(
+            std::str::from_utf8(&buffer).unwrap(),
+            r#"<paired attr1="value1" attr2="value2">text</paired>"#
```
dralley, line 452: Do these need to use bytes instead?  I notice the other tests do, but it's a lot less convenient to debug when they fail.  I can change them in-line with the others now that they pass.

dralley:
I was also thinking about a design that would allow arbitrary depth, but I wasn't sure if it could be done without allocations.

src/writer.rs
tafia, line 185: You don't need to keep a separate field for `tag_name` as you can get it from the `BytesStart::name`.

src/writer.rs
tafia: `&str` is better indeed!

tafia (changes requested):
Thanks!
I would suggest returning the `&mut Writer<W>` in `write_text`/`write_empty` for people to write more things in the same expression.

src/writer.rs
tafia: Sorry didn't read your comment on lifetimes.

src/writer.rs
```diff
@@ -394,4 +417,39 @@ mod indentation {
...
     }
+    #[test]
+    fn node_writer_empty() {
+        let mut buffer = Vec::new();
+        let mut writer = Writer::new_with_indent(&mut buffer, b' ', 4);
+
+        writer
+            .create_node(b"empty")
+            .with_attribute(("attr1", "value1"))
+            .with_attribute(("attr2", "value2"))
```
dralley, line 428: What do you think about passive ```with_attribute``` vs active ```push_attribute```?

I'm conflicted because then it uses active ```write_empty```.  There's decent arguments for both options

src/writer.rs
dralley: Yup.  It's possible there's a way around it that I don't know about though.

src/writer.rs
```diff
@@ -169,19 +170,41 @@ impl<W: Write> Writer<W> {
...
-        let name = parent_name.as_ref();
-        self.write_event(Event::Start(BytesStart::borrowed(name, name.len())))?;
-        self.write_event(event)?;
-        self.write_event(Event::End(BytesEnd::borrowed(name)))?;
+    /// Write nested
+    pub fn create_node<'a, N: 'a + AsRef<[u8]>>(&'a mut self, name: &'a N) -> NodeWriter<W> {
+        NodeWriter {
+            writer: self,
+            tag_name: name.as_ref(),
+            start_tag: BytesStart::borrowed_name(name.as_ref()),
```
dralley, line 178: Minor question, unrelated to the patch - what is the use case for ```borrowed``` over ```borrowed_name```?  I also noticed that ```borrowed``` on ```BytesEnd``` behaves like ```borrowed_name``` on ```BytesStart```.

src/writer.rs
```diff
@@ -394,4 +417,39 @@ mod indentation {
...
+    #[test]
+    fn node_writer_text() {
+        let mut buffer = Vec::new();
+        let mut writer = Writer::new_with_indent(&mut buffer, b' ', 4);
+
+        writer
+            .create_node(b"paired")
+            .with_attribute(("attr1", "value1"))
+            .with_attribute(("attr2", "value2"))
+            .write_text(BytesText::from_plain_str("text"))
```
dralley, line 447: Should a ```write_cdata``` be added?

dralley:
>I would suggest returning the &mut Writer<W> in write_text/write_empty for people to write more things in the same expression.

Could you give a code example of what this would look like?

src/writer.rs
tafia: I tend to use `with_...` for a builder like pattern (something that returns a `&mut Self`) and `push_...` for something that return `()`. It is more a convention thing but in your case with_attribute is better.

src/writer.rs
tafia: `borrowed` will borrow everything (name and attributes) thus you must specify the name length. `borrowed_name` assumes there is no attribute yet. 
For instance, when you're trying to pipe an xml borrowed is great because you can reuse everything. On the other hand when you want to create it from scratch `borrowed_name` is more natural.

src/writer.rs
tafia: Yes ... ideally all types of events

tafia:
> Could you give a code example of what this would look like? ("writing more things in the same expression")

```rust
 writer
            .create_node(b"paired")
            .with_attribute(("attr1", "value1"))
            .with_attribute(("attr2", "value2"))
            .write_text(BytesText::from_plain_str("text"))?
            .create_node(b"paired")  // another node
            .with_attribute(("attr1", "value1"))
            .with_attribute(("attr2", "value2"))
            .write_text(BytesText::from_plain_str("text"))?
            .write_event(...)? // directly write an event

```

It is not really better. On the other hand we're not losing anything from returning a `&mut self`.

src/writer.rs
dralley: All events, or all events of the type that would make sense to be nested in a single pair of tags?  I don't use XML outside of the one project I'm working on so I don't know - but based on what I could find, I thought that Text, CData, and maybe PI are the only ones you'd typically see in this context.  You probably wouldn't want to write random Start, End, Decl, EOF, etc.

 If you want to allow *any* type of event, we could just add a ```.write_event()```, but in my mind this is more of a simple API for the standard case, and if you want to reproduce weird / invalid XML perfectly then you would just use the existing API.

dralley:
>On the other hand we're not losing anything from returning a &mut self

Well, there is one downside.  It becomes possible to write invalid XML.  That could be fixed with the typestate pattern but it would make the code more complex.

src/writer.rs
```diff
@@ -394,4 +458,61 @@ mod indentation {
...
+        writer
+            .create_element(b"paired")
+            .with_attribute(("attr1", "value1"))
+            .with_attribute(("attr2", "value2"))
+            .write_nested_content(|writer| {
+                writer.create_element(b"inner").write_empty()?;
+                writer.create_element(b"inner2").write_empty()?;
+                Ok(())
+            })
+            .expect("failure");
```
dralley, line 511: I added one new idea.  Maybe this is a step too far?  But it seemed like a neat experiment, and it works.

src/writer.rs
```diff
@@ -394,4 +458,75 @@ mod indentation {
...
+
+        assert_eq!(
+            std::str::from_utf8(&buffer).unwrap(),
+            r#"<paired attr1="value1" attr2="value2">text</paired>"#
+        );
+    }
+
+    #[test]
+    fn element_writer_scope() {
```
dralley, line 498: I wanted to try this experiment, but I'm not strongly attached to it. What do you think?  It seems useful for really simple cases of nesting, but less so more complex ones.

src/writer.rs
tafia: decl probably doesn't make any sense indeed

dralley:
@tafia This is ready for re-review, although I should add more tests before it actually merges. Just let me know if there's anything objectionable about the API.

Here's what a larger piece of code looks like when rewritten to use this strategy: https://github.com/dralley/rpmrepo_rs/commit/6bb4a143240273842469278388868cb9236875fd

src/writer.rs
dralley: Not only was there a way, it was actually documented perfectly.

https://github.com/tafia/quick-xml/blob/4c853840f770cb87cb5482db3476ee262bb2ed55/src/events/mod.rs#L96-L115

tafia (approved):

tafia:
Looks good to me, let me know when you want to merge.
Thanks!

src/writer.rs
```diff
@@ -169,20 +169,134 @@ impl<W: Write> Writer<W> {
...
+        self.writer
+            .write_event(Event::Start(self.start_tag.to_borrowed()))?;
+        self.writer.write_event(Event::CData(text))?;
+        self.writer
+            .write_event(Event::End(self.start_tag.to_end()))?;
+        Ok(self.writer)
+    }
+
+    /// Write a processing instruction `<?...?>` inside the current element.
+    pub fn write_pi_content(self, text: BytesText) -> Result<&'a mut Writer<W>> {
```
dralley, line 274: Name OK?

src/writer.rs
```diff
@@ -169,20 +169,134 @@ impl<W: Write> Writer<W> {
...
+    where
+        I: IntoIterator,
+        I::Item: Into<Attribute<'b>>,
+    {
+        self.start_tag.extend_attributes(attributes);
+        self
+    }
+
+    /// Write some text inside the current element.
+    pub fn write_text_content(self, text: BytesText) -> Result<&'a mut Writer<W>> {
```
dralley, line 254: Does a pair of tags with no text always get represented as [Start] [Text ""] [End]?

dralley:
@tafia Apart from the 3 questions, and your approval of the docstrings I just added, I think this is good to go.

src/writer.rs
```diff
@@ -169,20 +169,134 @@ impl<W: Write> Writer<W> {
...
+    }
+
+    /// Write an empty (self-closing) tag.
+    pub fn write_empty(self) -> Result<&'a mut Writer<W>> {
+        self.writer.write_event(Event::Empty(self.start_tag))?;
+        Ok(self.writer)
+    }
+
+    /// Create a new scope for writing XML inside the current element.
+    pub fn write_nested_content<F>(mut self, closure: F) -> Result<&'a mut Writer<W>>
```
dralley, line 290: Would `inner` or `scoped` or any other word make more sense here?

src/events/mod.rs
```diff
@@ -315,10 +315,10 @@ impl<'a> BytesStart<'a> {
     }
 
     /// Try to get an attribute
-    pub fn try_get_attribute(&'a self, attr_name: &[u8]) -> Result<Option<Attribute<'a>>> {
+    pub fn try_get_attribute<N: AsRef<[u8]> + Sized>(&'a self, attr_name: N) -> Result<Option<Attribute<'a>>> {
```
dralley, line 318: @tafia This lets the user pass either an `&str` or a `&[u8]`

src/events/mod.rs
```diff
@@ -315,10 +315,13 @@ impl<'a> BytesStart<'a> {
     }
 
     /// Try to get an attribute
-    pub fn try_get_attribute(&'a self, attr_name: &[u8]) -> Result<Option<Attribute<'a>>> {
+    pub fn try_get_attribute<N: AsRef<[u8]> + Sized>(
```
dralley, line 318: Attribute names don't need to be escaped right?  This would let users pass `&str`, `&[u8]`, or `Vec<u8>`.

src/writer.rs
tafia: What do you mean? If possible I think we could *optimize* it and write an empty event but this is really not very important.

src/writer.rs
tafia: `inner` seems better. Not a fan of `scoped` I think.

src/events/mod.rs
tafia: right, names don't need escaping.

src/writer.rs
dralley: Done

src/writer.rs
dralley: I just wanted to be sure that what I'm doing is correct - if it was ever necessary to write two separate start and end events with no text event (even an empty one) in the middle, this wouldn't be convenient. But if there needs to always be one, then this is fine.

tafia:
Great! Thanks a lot for your time.

src/events/mod.rs
dralley: I noticed one difference between a library I'm writing and an existing one that uses libxml2 - the one based on libxml2 seems to strip tab characters `\t` out of attribute values.  I can't find anything in the spec that mentions that nor can I find it in libxml2 though - do you know whether that might be required?

PR #279, opened 2021-03-30
Add a from_slice() function like other serde datatype libraries

dralley:
`serde_json` and `serde_yaml` and a few others have a `from_slice` function, adding one here for parity.

https://docs.serde.rs/serde_yaml/index.html


tafia:
Thanks

PR #281, opened 2021-04-03
Add a function for partially escaping element text

dralley:
Quote characters (", ') are not strictly disallowed inside of element
text. Provide a variant of the escape function which allows these
characters.


dralley:
I think doing this is still "discouraged" so it's probably not worth adding any more API to make this easy to do.  It's nice to have the option though.

https://stackoverflow.com/questions/150423/whats-the-accepted-way-of-storing-quoted-data-in-xml
https://stackoverflow.com/questions/730133/what-are-invalid-characters-in-xml

tafia:
Sorry adding the option is fine indeed. Thanks

Issue #291, opened 2021-06-13
Add new tests for recent changes


andreivasiliu:
Merging https://github.com/tafia/quick-xml/pull/290 as-is feels a bit scary, since some of the more subtle changes that were committed since then might be reverted.

I'd like to write some additional tests to make sure they won't be.

I went through the commit log, and these are all the changes that have happened since I made the original PR (2020-06):

* [ ] [Fix unescaping broken codepoints](https://github.com/tafia/quick-xml/commit/20322285c24e16c9a56716b312ee7411c3e4f2fe)
  * This likely broke the html5 test, and needs some new tests
* [ ] [Add '*_with_custom_entities' versions of unescape methods](https://github.com/tafia/quick-xml/commit/4c853840f770cb87cb5482db3476ee262bb2ed55)
  * Has some private tests, but needs tests on new public methods
  * Note to self: Probably a good place to check for the billion laughs attack vulnerability too
* [ ] [Put escaping html entities behind a 'escape-html' feature](https://github.com/tafia/quick-xml/commit/745341a755fd4871668a985b6fa8391344101322)
  * Has no tests; also kinda abuses features, as features should not change behavior of existing APIs
* [x] [Inline CData events when pretty printing](https://github.com/tafia/quick-xml/commit/d3c495ed29f26ecbff2b33aba1b452f35be1e4eb)
  * Only concerns the writer.
* [ ] [Split text trim into start and end](https://github.com/tafia/quick-xml/commit/f644c9f3053a7649fda3948051775713f007a412)
  * `trim_text` option becomes two options: `trim_text_start` and `trim_text_end`
  * Definitely needs tests
* [x] [`$value` rename should work the same for serialization too](https://github.com/tafia/quick-xml/commit/b46d9f414f9a2ab8ed114d8eb70a54a6e313fffc)
  * Already has tests! Hurray!
* [x] [Parse & in cdata correctly](https://github.com/tafia/quick-xml/commit/b3167bfe4ed1deea377eed001eebc581a43850ba)
  * Already has tests! Hurray!
* [ ] [Accept html5 doctype](https://github.com/tafia/quick-xml/commit/f3f88b6cdddb18c672133e917913346e5017d5ed)
  * Allows case-insensitive matching of the doctype tag
  * Has tests, but they fail (likely broken by https://github.com/tafia/quick-xml/pull/271)
[...]

PR #292, opened 2021-06-13
Improve CI: test with more feature combinations, test on windows-latest too

andreivasiliu:
I've had issues with tests failing on Windows on line-breaks and such, so I'd like to add `windows-latest` to the list of platforms that the GitHub Actions workflow will test.

This crate sadly doesn't use features correctly; that should be fixed, but until then, I'd like to test more combinations of features to make sure a change that was supposed to be gated won't be a compile error.

I also fixed some warnings that Rust gives for some semicolons, and for `panic!(e)`, whose behavior will change in Rust 2021.

This is mostly to make my life easier in https://github.com/tafia/quick-xml/issues/291.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,12 +4,21 @@ on: [push, pull_request]
 
 jobs:
   build:
+    strategy:
+      matrix:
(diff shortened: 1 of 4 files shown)


tafia:
Thanks!

PR #295, opened 2021-07-17
Fix warnings from newer rust compiler which will be hard errors in Rust 2021

Mingun:
Specifically fix the following warnings:

```console
warning: panic message is not a string literal
   --> tests\unit_tests.rs:760:30
    |
760 |             Err(e) => panic!(e),
    |                              ^
    |
    = note: this is no longer accepted in Rust 2021
help: add a "{}" format string to Display the message
    |
760 |             Err(e) => panic!("{}", e),
    |                              ^^^^^
help: or use std::panic::panic_any instead
    |
760 |             Err(e) => std::panic::panic_any(e),
    |                       ^^^^^^^^^^^^^^^^^^^^^
```

All `panic!(e)` invocations replaced by `panic!("{}", e)` as suggested because `panic_any` is availible only from `1.51.0`

### DIFF
--- a/tests/serde-migrated.rs
+++ b/tests/serde-migrated.rs
@@ -953,12 +953,12 @@ fn futile2() {
     #[derive(Eq, PartialEq, Debug, Serialize, Deserialize)]
     struct Object {
         field: Option<Null>,
-    };
+    }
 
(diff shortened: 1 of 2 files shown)


tafia:
Thanks

Issue #245, opened 2020-12-23
Serialize struct as children of struct name and not attributes.


tomgie:
Is there any way to serialize a struct as children instead of attributes? An example of this would be:
```xml
<MyStruct>
    <ValueOne>A</ValueOne>
    <ValueTwo>B</ValueTwo>
</MyStruct>
```

quebin31:
Currently it seems that there's no way to tell the serializer how to treat different fields, I was thinking on a trait (e.g. `Xml`) that should be derived and adds the ability to do something like this:


```rs
#[derive(Serialize, XML)]
pub struct MyStruct {
    #[xml(attr)]
    something: String,

    #[serde(rename = "Child")]
    #[xml(child)]
    child: String,
}
```

Serializing into:

```xml
<MyStruct something="A">
    <Child>B</Child>
</MyStruct>
```

tafia:
It could probably work but I think it is really not trivial.

quebin31:
@tafia yep, it's not trivial, I tried to take hands-on the code but not success, meanwhile I was using the fork from #251

lukts30:
Could a feature flag be introduced to just make every struct member a child instead of an attribute?
This seems to be the default for pre 0.19. Every newer version breaks serialisation for me since everything is now an attribute instead.

Essentially, I want to opt out of *0.19.0: (Serde) Serialize basic types as attributes (breaking change)*

tafia:
I'm ok for the feature flag (probably defaulting to attributes).

woodworker:
Would this be solved with the `#[serde(rename = "$value")]` 

As seen in https://github.com/tafia/quick-xml/blob/a4be48431a2c512683538b81d7ee037af70d5cce/tests/serde_roundtrip.rs#L23-L26 and https://github.com/tafia/quick-xml/blob/a4be48431a2c512683538b81d7ee037af70d5cce/tests/serde_roundtrip.rs#L59-L69

tomgie:
> Currently it seems that there's no way to tell the serializer how to treat different fields, I was thinking on a trait (e.g. `Xml`) that should be derived and adds the ability to do something like this:
> 
> ```rust
> #[derive(Serialize, Xml)]
> pub struct MyStruct {
>     #[xml(attr)]
>     something: String,
> 
>     #[serde(rename = "Child")]
>     #[xml(child)]
>     child: String,
> }
> ```
> 
> Serializing into:
> 
> ```
> <MyStruct something="A">
[...]

PR #298, opened 2021-07-19
Implement unflattening structs

solita-jaakkoha:
This pull gives an (non-ideal) implementation for #245.

Essentially this allows `<root><first>value</first><second>value</second></root>` to be (de)serialized with the special name prefix `$unflatten=`:
```rust,ignore
struct Root {
    #[serde(rename = "$unflatten=first")]
    first: String,
    #[serde(rename = "$unflatten=second")]
    other_field: String,
}
```

Ideally this would be something that could be configured on a struct level, but since from what I understand Serde doesn't allow custom per-implementation attributes... so this uses `rename` as a backchannel, like $value does.

It would be nicer to implement a separate XML trait, but Serde doesn't really like that either, like the comment in #205 knows.
https://github.com/tafia/quick-xml/pull/205#issuecomment-605680462

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,20 @@ struct Foo {
 }
 ```
 
+### Unflattening structs into verbose XML
+
+If your XML files look like `<root><first>value</first><second>value</second></root>`, you can
+(de)serialize them with the special name prefix `$unflatten=`:
+
+```rust,ignore
+struct Root {
+    #[serde(rename = "$unflatten=first")]
+    first: String,
+    #[serde(rename = "$unflatten=second")]
+    other_field: String,
+}
+```
+
 ### Performance
 
 Note that despite not focusing on performance (there are several unecessary copies), it remains about 10x faster than serde-xml-rs.
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -1,7 +1,7 @@
 //! Serde `Deserializer` module
 
 use crate::{
-    de::{escape::EscapedDeserializer, Deserializer, INNER_VALUE},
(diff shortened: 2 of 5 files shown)


doowonee:
This is the feature that I need right now! please merge it 🙏🏼

tafia:
This is great thanks!

mohe2015:
@solita-jaakkoha Nice work but it seems like this doesn't work properly with namespaced tags (tags with column). You need to use the tag without the namespace which is weird.

solita-jaakkoha:
Meh, XML is hard. I haven't ever used namespaced XMLs so you're probably in a better spot to fix that than I am :)

mohe2015:
> Meh, XML is hard. I haven't ever used namespaced XMLs so you're probably in a better spot to fix that than I am :)

Unfortunately I had other major issues with parsing nested structures that I have decided to switch libraries. The namespace thing was not that bad as it worked for me without specifying the namespace. Maybe I will get some time to create a proof of concept for both issues so I can report them at least (or find my mistake)

Mingun:
I'm not quite understand what the problem is in the original code snippet?

```xml
<root>
  <first>value</first>
  <second>value</second>
</root>
```
should map to the
```rust
struct AnyName {
    first: String,
    #[serde(rename = "second")]
    other_field: String,
}
```
without any additions and that proved by this test:
https://github.com/tafia/quick-xml/blob/aa6ec2dd25b83c3b36cf87c19c27349a02b619b2/src/de/mod.rs#L945-L962
So what the original problem? Right now it seems to me that `$unflatten=` prefix is not required. Moreover, #245 is about serialization only, I don't even see how it is related to this PR.

PR #300, opened 2021-08-05
Implemented From trait for EscapeError to Error

SenneH:
A slight change that helps when dealing with EscapeError.


tafia:
Thanks

PR #301, opened 2021-08-07
Add partial escape to API

dralley:
I added the function but forgot to add it to the API

@tafia Are there any plans for a new release soon?  I can't publish my crate as long as it has to depend on non-released git dependencies


tafia:
I'll make a new release this week. Thanks

PR #302, opened 2021-08-10
Fix warnings

Mingun:
- Fix warnings from https://github.com/rust-lang/rust/issues/79813:
  ```console
  warning: trailing semicolon in macro used in expression position
    --> src\events\attributes.rs:362:20
      |
  362 |                 }));
      |                    ^
  ...
  443 |             None => attr!(start_key..end_key),
      |                     ------------------------- in this macro invocation
      |
      = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
      = note: for more information, see issue #79813 <https://github.com/rust-lang/rust/issues/79813>
      = note: macro invocations at the end of a block are treated as expressions
      = note: to ignore the value produced by the macro, add a semicolon after the invocation of `attr`
      = note: this warning originates in the macro `attr` (in Nightly builds, run with -Z macro-backtrace for more info)
  ```
  Because in the end it's stay non-obvilious that macro return some value, I've moved out the `return` keyword from the macro.
- Fix a bug introduced in the 8481c44c6b0134ec62226a80efc482f7ce7f8f6e -- some errors do not propagated to the caller
- Remove unused import
- Some bit of documentation in the deserializer
- Remove excess `cfg` check -- it was already checked couple lines above.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,7 +15,6 @@ jobs:
     - name: Build
       run: cargo build
     - name: Run tests (no features)
-      run: cargo test
       run: cargo test --no-default-features
     - name: Run tests (serialize)
       run: cargo test --features serialize
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -9,8 +9,14 @@ use serde::de::{self, DeserializeSeed, IntoDeserializer};
 
 enum MapValue {
     Empty,
+    /// Value should be deserialized from the attribute value
     Attribute { value: Vec<u8> },
     Nested,
+    /// Value should be deserialized from the text content of the XML node:
+    ///
+    /// ```xml
+    /// <...>text content for field value<...>
+    /// ```
     InnerValue,
 }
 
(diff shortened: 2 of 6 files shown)


dralley:
Could you also fix the CI which is apparently [silently broken](https://github.com/tafia/quick-xml/actions/runs/1117558602)?

dralley:
https://github.com/tafia/quick-xml/blob/master/.github/workflows/rust.yml#L18

This

```
    - name: Run tests (no features)
      run: cargo test
      run: cargo test --no-default-features
```

needs to become

```
    - name: Run tests (no features)
      run: |
          cargo test
          cargo test --no-default-features
```

dralley:
Notice the green check mark on this PR even though the tests fail to run, I'm pretty sure that's what happened with the other PR as well.

Mingun:
> needs to become

Are you sure? It seems that line
```yaml
run: cargo test
```
need to be completely removed

dralley:
Sure, that's possible too.  Maybe the duplicate line was introduced as part of a bad merge conflict resolution.  I'm only speaking about what needs to be done to make it parse properly

tafia:
Thanks! Indeed I blindly followed the green mark on the PR. Seems to be better now.
@Mingun do you mind fixing the last error?

Mingun:
Yeah, need always to test changes themself... and do not forget to add `--all-features` 😄 

Fixed the last error.

tafia:
Awesome thanks!

Issue #280, opened 2021-03-30
Vec<String> inside of struct serializes incorrectly with serde


dralley:
I was attempting to add quick-xml to the [Rust serializer benchmarks](https://github.com/djkoloski/rust_serialization_benchmark) and discovered that one of the complex benchmarks serialized incorrectly, resulting in errors on deserialization.  The benchmark in question is the "minecraft savedata" benchmark.

I minimized it so it's less unwieldy.  Here's a test case.

```rust

const RECIPES: [&'static str; 8] = [
    "pickaxe",
    "torch",
    "bow",
    "crafting table",
    "furnace",
    "shears",
    "arrow",
    "tnt",
];

#[derive(serde::Deserialize, serde::Serialize)]
pub struct RecipeBook {
    pub recipes: Vec<String>,
}

#[test]
fn test_fail_serialize() {
    let recipe_book = RecipeBook {
        recipes: RECIPES.iter().map(|s| s.to_string()).collect(),
    };

    let serialized = to_string(&recipe_book).unwrap();

    println!("{}", serialized);
    let deserialized: RecipeBook = from_str(&serialized).unwrap();
}

```

Prints the following when tests are run:

```
---- test_fail_serialize stdout ----
<RecipeBook recipes="pickaxetorchbowcrafting tablefurnaceshearsarrowtnt"/>
thread 'test_fail_serialize' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("invalid type: string \"pickaxetorchbowcrafting tablefurnaceshearsarrowtnt\", expected a sequence")', tests/serde_roundtrip.rs:55:58
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The problem is that `recipes: Vec<String>` and `to_be_displayed: Vec<String>` are being serialized into one long concatenated string and then assigned to an attribute of a self-closing element, rather than being serialized to a list of inner elements.  Deserialize does the correct thing and expects a list of strings.

dralley:
This may be related to https://github.com/tafia/quick-xml/pull/205

shalinnijel2:
Is there a potential work around for this issue while this is still being worked on?

shalinnijel2:
I guess using a custom serializer which uses serialize_field() is one option...

zmrow:
I'm also wondering what the workaround is here... :)

Mingun:
Workaround is use Newtype idiom:
```rust
#[test]
fn test_fail_serialize() {
[...]

PR #303, opened 2021-08-11
Add a test for #280

dralley:
Currently fails: https://github.com/tafia/quick-xml/issues/280

### DIFF
--- a/tests/serde_roundtrip.rs
+++ b/tests/serde_roundtrip.rs
@@ -78,6 +78,27 @@ fn round_trip_list_of_enums() {
 }
 
 #[test]
+#[ignore]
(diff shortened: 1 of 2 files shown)


Mingun:
As I already said in https://github.com/Mingun/fast-xml/pull/2, it is better to test actual result and not roundtrips, because deserializer is able to deserialize different shaped into the same structure, and for serializer we usually want some exact format (because usually XML has a scheme and that scheme is not under our control).

Because I'll plan to look serializer issues more closely when I'll finish with deserializer, I suggest close this for now.

dralley:
I'll close it for now, once I get some free time on a weekend I will try to address your feedback, presuming you don't get to it first :)

PR #304, opened 2021-08-12
Add $primitive for serializing unit variants as string

### LINKED ISSUE #283: Is this result expected when serializing in the enum type?
Thanks for your effort. It is awesome that using serde to serialize or deserialize the xml.
There are some differences in serializing enum type between the output in quick_xml and that in serde_json. 
```rs

#[derive(Serialize, Deserialize, Debug)]
enum Type {
    Foo,
    Bar,
}

#[derive(Serialize, Deserialize, Debug)]
struct A {
    pub member: Type
}

fn main() {
    let data = A {member: Type::Foo};
    let json_result = serde_json::to_string(&data).unwrap();
    println!("{:?}", json_result);
    let xml_result = quick_xml::se::to_string(&data).unwrap();
    println!("{:?}", xml_result);
}

```
The output of this code is
```
"{\"member\":\"Foo\"}"
"<A><Foo/></A>"
```

Is this result is expected? I think it should be 
```xml
"<A member="Foo"><A/>"
```

fwcd:
### Fixes #283 

This PR updates the serializer for unit variants to serialize them as string primitives instead of self-closing tags if prefixed with `$primitive`. For example, if we define

```rust
enum X {
    #[serde(rename = "$primitive=A")]
    A
}

struct Y {
    x: X
}
```

`X::A` will now serialize to `A` instead of `<A/>` and `Y { x: X::A }` will now serialize to `<Y x="A"/>` instead of `<Y><A/></Y>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,23 @@ struct Root {
 }
 ```
 
+### Serializing unit variants as primitives
+
+The `$primitive` prefix lets you serialize enum variants without associated values (internally referred to as _unit variants_) as primitive strings rather than self-closing tags. Consider the following definitions:
+
+```rust,ignore
+enum Foo {
+    #[serde(rename = "$primitive=Bar")]
+    Bar
+}
+
+struct Root {
+    foo: Foo
+}
+```
+
+Serializing `Root { foo: Foo::Bar }` will then yield `<Root foo="Bar"/>` instead of `<Root><Bar/></Root>`.
+
 ### Performance
 
 Note that despite not focusing on performance (there are several unecessary copies), it remains about 10x faster than serde-xml-rs.
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -125,6 +125,7 @@ use std::io::BufRead;
 
 pub(crate) const INNER_VALUE: &str = "$value";
 pub(crate) const UNFLATTEN_PREFIX: &str = "$unflatten=";
+pub(crate) const PRIMITIVE_PREFIX: &str = "$primitive=";
 
 /// An xml deserializer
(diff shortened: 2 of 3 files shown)


bachp:
@fwcd Could this be somehow controlled via an `#[]` attribute on an enum or enum variant level?

fwcd:
Perhaps, I am not that familiar with serde's macro system though. We might be able to use a special dollar-named prefix, like already used for unflattening structs (see [here](https://github.com/tafia/quick-xml#unflattening-structs-into-verbose-xml)).

I have a feeling that most use cases where unit variants are involved are concerned with serializing 'simple' enums without associated values that are stored as a field, and making unit variants-as-strings (i.e. this PR) the default behavior would match the intuitive serialization there. Any thoughts?

tafia:
I think the default behavior isn't necessarily unintuitive. Sadly this discussion is probably to late because I am afraid to break quite a lot of code.

I agree with a new $ attribute to make it work. This is far from perfect but xml really isn't that great with serde and this is the best workaround I can see.

Eventually we'll have to properly document all these $ attributes.

fwcd:
I have restored the previous behavior and added a `$primitive=` prefix that can be placed on unit variants to optionally serialize them as strings. A new paragraph in the README documents this behavior.

djc:
@tafia would really like to be able to use this. Is there some chance of getting this merged?

tafia:
Thanks, I'll publish a new version this weekend

djc:
@tafia awesome, thanks!

djc:
@tafia happy new year! Gentle ping on getting this out to crates.io?

PR #306, opened 2021-08-14
fix deserializing multiple unflatten fields

### LINKED ISSUE #305: $unflatten breaks with namespaces and nesting
Reproducers at https://github.com/mohe2015/quick-xml-bugs. Hopefully I didn't mess anything up.

Crossref https://github.com/tafia/quick-xml/pull/298

cc @solita-jaakkoha

tafia:
Close #305

### DIFF
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -104,13 +104,26 @@ impl<'de, 'a, R: BorrowingReader<'de> + 'a> de::MapAccess<'de> for MapAccess<'de
                     self.value = MapValue::InnerValue;
                     seed.deserialize(INNER_VALUE.into_deserializer()).map(Some)
                 }
+                // Used to deserialize collections of enums, like:
+                // <root>
+                //   <xxx>
+                //   </xxx>
+                // </root>
+                //
+                // into
+                //
+                // struct Root {
+                //     #[serde(rename = "$unflatten=xxx")]
+                //     xxx: String,
+                // }
+                // TODO: This should be handled by #[serde(flatten)]
+                // See https://github.com/serde-rs/serde/issues/1905
                 Some(Event::Start(e)) if has_unflatten_field => {
                     self.value = MapValue::InnerValue;
                     let key = format!(
                         "{}{}",
                         UNFLATTEN_PREFIX,
-                        String::from_utf8(e.local_name().to_vec())
-                            .expect("$unflatten= did not contain valid Rust identifier")
+                        decoder.decode(e.name())?
                     );
                     seed.deserialize(key.into_deserializer()).map(Some)
                 }
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -36,17 +36,19 @@ pub(crate) struct MapAccess<'de, 'a, R: BorrowingReader<'de> + 'a> {
     /// to restore last position before advance.
     position: usize,
     value: MapValue,
+    size_hint: Option<usize>,
 }
 
 impl<'de, 'a, R: BorrowingReader<'de>> MapAccess<'de, 'a, R> {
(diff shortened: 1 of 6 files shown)


mohe2015:
@tafia The two bugs are fixed! I found another one though: https://github.com/mohe2015/quick-xml-bugs/commit/8dfcb238406143102664fc186480758786423d90. I would expect to get an empty Vec if I'm not mistaken.

Also the namespace thing is probably not a final fix. Although I don't know XML really well I think we would need some mapping from namespace urls. Let me give an example (CalDAV):
```xml
<?xml version="1.0" encoding="utf-8"?>
<d:propfind xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:d="DAV:" xmlns:nc="http://nextcloud.org/ns" xmlns:oc="http://owncloud.org/ns" xmlns:s="http://sabredav.org/ns">
  <d:prop xmlns:x1="http://apple.com/ns/ical/" xmlns:x2="http://nextcloud.com/ns">
    <d:current-user-principal />
  </d:prop>
</d:propfind>
```
should as far as I understand it should be equivalent to:
```xml
<?xml version="1.0" encoding="utf-8"?>
<dav:propfind xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:dav="DAV:" xmlns:nc="http://nextcloud.org/ns" xmlns:oc="http://owncloud.org/ns" xmlns:s="http://sabredav.org/ns">
  <dav:prop xmlns:x1="http://apple.com/ns/ical/" xmlns:x2="http://nextcloud.com/ns">
    <dav:current-user-principal />
  </dav:prop>
</dav:propfind>
```
The way yaserde implements this is you specify a mapping of namespace-shorthands to namespace urls:
```rust
#[yaserde(
    namespace = "d: DAV:",
    namespace = "s: http://sabredav.org/ns",
    namespace = "cal: urn:ietf:params:xml:ns:caldav",
    namespace = "cs: http://calendarserver.org/ns/",
    namespace = "oc: http://owncloud.org/ns",
    namespace = "nc: http://nextcloud.org/ns"
)]
```
and then can use these shorthands. A potentially easier but maybe also annoying way would be to directly use the url maybe like:
```rust
#[serde(rename = "$unflatten=urn:ietf:params:xml:ns:caldav href")]
```
where we should choose delimiters as invalid characters of the namespace url. But then you would need to store all `xmlns:` attributes on a stack or so to know which shorthand this belongs to.

Either way I think this could be fixed later.

Mingun:
Be noted, that this PR uses `IntoDeserializer for &[u8]` implementation that was made public only from [serde 1.0.122](https://github.com/serde-rs/serde/releases/tag/v1.0.122). Need to bump serde requirement

mohe2015:
Should I open issues for the remaining two things?

src/de/map.rs
```diff
@@ -62,16 +79,20 @@ impl<'de, 'a, R: BorrowingReader<'de>> MapAccess<'de, 'a, R> {
 impl<'de, 'a, R: BorrowingReader<'de> + 'a> de::MapAccess<'de> for MapAccess<'de, 'a, R> {
     type Error = DeError;
 
+    fn size_hint(&self) -> Option<usize> {
+        self.size_hint.clone()
+    }
+
```
Mingun, line 85: This is not what the `size_hint()` method is intended. It should return count of elements that deserializer _have_, not count of elements that deserialized type _requested_ from it. In our case default implementation (that returns `None`) is only possible solution because XML reader can't say how many elements it have unless it parses them all

PR #308, opened 2021-08-15
Add tests for escaping/encoding, trimming, and cloning

andreivasiliu:
Add tests for escaping/encoding, trimming, and cloning, which are some of the items in https://github.com/tafia/quick-xml/issues/291.

Also fixed the html5 test, which, surprisingly, is no longer different on Windows; the CI workflow says it succeeds everywhere with the same expected output, which to me sounds like a good thing.


tafia:
Thanks!

PR #309, opened 2021-08-21
Additional basic tests for serde deserializer

Mingun:
Added tests for parsing maps and structs from malformed inputs:
- unclosed tags
- mismatched tags

Add tests that ensures that excess attributes/elements do not break deserialization of complex structures (unis/newtypes/tuples/structs).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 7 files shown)


tafia:
Thanks!!!!
That's awesome.

PR #314, opened 2021-08-25
Allow using tokio's AsyncBufRead [Rebased]

itsgreggreg:
This is a rebase of this PR https://github.com/tafia/quick-xml/pull/233 on top of the current master.

I don't know if there was a better way to do this and my intention is not to step on anyone's work but rather keep it moving along. I'm happy to take this PR down if you want to update yours @endor or honestly whatever works for people. This PR is a proper rebase of the endor/async on top of tafia/master so line by line credit is still in tact except for the few places I had to fix rebase conflicts.

@endor, @tafia happy to help with whatever y'all need help with here, lemme know!

### DIFF
--- a/src/reader/asynchronous.rs
+++ b/src/reader/asynchronous.rs
@@ -1,4 +1,4 @@
-//! A module to handle the async `Reader`
+//! A module to handle the `AsyncReader`
 
 use async_recursion::async_recursion;
 #[cfg(feature = "encoding")]
@@ -18,7 +18,7 @@ use crate::events::{BytesDecl, BytesEnd, BytesStart, BytesText, Event};
 
 use super::{is_whitespace, Decode, Decoder, NamespaceBufferIndex, TagState};
 
-impl<B: AsyncBufRead> Decode for Reader<B> {
+impl<B: AsyncBufRead> Decode for AsyncReader<B> {
     #[cfg(feature = "encoding")]
     fn read_encoding(&self) -> &'static Encoding {
         self.encoding
@@ -47,7 +47,7 @@ impl<B: AsyncBufRead> Decode for Reader<B> {
 /// # Examples
 ///
 /// ```
-/// use quick_xml::Reader;
+/// use quick_xml::AsyncReader;
 /// use quick_xml::events::Event;
 ///
 /// #[tokio::main]
@@ -56,7 +56,7 @@ impl<B: AsyncBufRead> Decode for Reader<B> {
 ///                     <tag2><!--Test comment-->Test</tag2>
 ///                     <tag2>Test 2</tag2>
 ///                 </tag1>"#;
-///     let mut reader = Reader::from_str(xml);
+///     let mut reader = AsyncReader::from_str(xml);
 ///     reader.trim_text(true);
 ///     let mut count = 0;
 ///     let mut txt = Vec::new();
@@ -81,7 +81,7 @@ impl<B: AsyncBufRead> Decode for Reader<B> {
 ///     }
 /// }
 /// ```
-pub struct Reader<B: AsyncBufRead> {
+pub struct AsyncReader<B: AsyncBufRead> {
     /// reader
     reader: B,
     /// current buffer position, useful for debuging errors
@@ -113,10 +113,10 @@ pub struct Reader<B: AsyncBufRead> {
     is_encoding_set: bool,
 }
 
-impl<B: AsyncBufRead + Unpin + Send> Reader<B> {
+impl<B: AsyncBufRead + Unpin + Send> AsyncReader<B> {
     /// Creates a `Reader` that reads from a reader implementing `BufRead`.
-    pub fn from_reader(reader: B) -> Reader<B> {
-        Reader {
+    pub fn from_reader(reader: B) -> AsyncReader<B> {
+        AsyncReader {
             reader,
             opened_buffer: Vec::new(),
             opened_starts: Vec::new(),
@@ -146,7 +146,7 @@ impl<B: AsyncBufRead + Unpin + Send> Reader<B> {
     /// [`Empty`]: events/enum.Event.html#variant.Empty
     /// [`Start`]: events/enum.Event.html#variant.Start
     /// [`End`]: events/enum.Event.html#variant.End
-    pub fn expand_empty_elements(&mut self, val: bool) -> &mut Reader<B> {
+    pub fn expand_empty_elements(&mut self, val: bool) -> &mut AsyncReader<B> {
         self.expand_empty_elements = val;
         self
     }
@@ -159,7 +159,7 @@ impl<B: AsyncBufRead + Unpin + Send> Reader<B> {
     /// (`false` by default)
     ///
     /// [`Text`]: events/enum.Event.html#variant.Text
-    pub fn trim_text(&mut self, val: bool) -> &mut Reader<B> {
+    pub fn trim_text(&mut self, val: bool) -> &mut AsyncReader<B> {
         self.trim_text = val;
         self
     }
@@ -175,7 +175,7 @@ impl<B: AsyncBufRead + Unpin + Send> Reader<B> {
     /// (`true` by default)
     ///
     /// [`End`]: events/enum.Event.html#variant.End
-    pub fn trim_markup_names_in_closing_tags(&mut self, val: bool) -> &mut Reader<B> {
+    pub fn trim_markup_names_in_closing_tags(&mut self, val: bool) -> &mut AsyncReader<B> {
         self.trim_markup_names_in_closing_tags = val;
         self
     }
@@ -193,7 +193,7 @@ impl<B: AsyncBufRead + Unpin + Send> Reader<B> {
     /// (`true` by default)
     ///
     /// [`End`]: events/enum.Event.html#variant.End
-    pub fn check_end_names(&mut self, val: bool) -> &mut Reader<B> {
+    pub fn check_end_names(&mut self, val: bool) -> &mut AsyncReader<B> {
         self.check_end_names = val;
         self
     }
@@ -208,7 +208,7 @@ impl<B: AsyncBufRead + Unpin + Send> Reader<B> {
     /// (`false` by default)
     ///
     /// [`Comment`]: events/enum.Event.html#variant.Comment
-    pub fn check_comments(&mut self, val: bool) -> &mut Reader<B> {
+    pub fn check_comments(&mut self, val: bool) -> &mut AsyncReader<B> {
         self.check_comments = val;
         self
     }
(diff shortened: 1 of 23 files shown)


endor:
Looks good to me. After we merge, it would probably make sense to extract the `BufferedInput` from the sync reader into the mod and have one implementation in sync and one in asynchronous.

src/utils.rs
```diff
@@ -1,17 +1,7 @@
...
-            write!(f, "Owned(")?;
-            write_byte_string(f, &s)?;
-        }
-        Cow::Borrowed(s) => {
-            write!(f, "Borrowed(")?;
-            write_byte_string(f, s)?;
-        }
-    }
+pub fn write_cow_string(f: &mut Formatter<'_>, s: &[u8]) -> Result {
+    write_byte_string(f, s)?;
```
tafia, line 4: is it intentional?

tests/unit_tests.rs
```diff
@@ -318,11 +329,12 @@ fn test_write_attrs() -> Result<()> {
     reader.trim_text(true);
     let mut writer = Writer::new(Cursor::new(Vec::new()));
     let mut buf = Vec::new();
+
     loop {
         let event = match reader.read_event(&mut buf)? {
             Eof => break,
             Start(elem) => {
-                let mut attrs = elem.attributes().collect::<Result<Vec<_>>>()?;
+                let mut attrs = elem.attributes().collect::<Result<Vec<_>>>().unwrap();
```
tafia, line 337: why not keeping the `?`?

tafia (approved):
Awesome!

I have a few small questions but I am willing to merge it so we can iterate on it.

I don't really like all the code duplication and I believe there should be a way to be more DRY.

src/utils.rs
itsgreggreg: nope! good eyes, reverted.

tests/unit_tests.rs
itsgreggreg: reverted!

itsgreggreg:
Ran the formatter, couple small cleanups.

src/reader/asynchronous.rs
```diff
@@ -0,0 +1,1036 @@
...
+
+    #[cfg(feature = "encoding")]
+    fn write_is_encoding_set(&mut self, val: bool) {
+        self.is_encoding_set = val;
+    }
+}
+
+/// A low level encoding-agnostic XML event reader.
+///
+/// Consumes a `BufRead` and streams XML `Event`s.
```
Richterrettich, line 45: Should  this be "Consumes a `AsyncBufRead`..."?

endor: Yes.

endor:
What is this waiting for?

tafia:
The is a missing file, @itsgreggreg any idea?

itsgreggreg:
It's a path joining error, If you look at the error the first separator in the path is a windows style `\` and the rest of them are unix style `/`.

https://github.com/tafia/quick-xml/runs/3644904506?check_suite_focus=true#step:7:5

I don't know what's causing that.

itsgreggreg:
whelp, here's why https://doc.rust-lang.org/std/macro.include_bytes.html

tafia:
I'm a bit surprised, I have used forward slash in windows quite a lot at work and I never had this issue.

itsgreggreg:
Upon closer inspection it seems that through the rebase I missed some changes to xmlrs_reader_tests.rs that removed that path but the file was still removed. I went through the file and hopefully made sure it was up to date.

Interesting that the rust docs say paths in include_str and friends have to be OS specific but unix paths work on windows. I wonder if windows paths work on *nix.

999eagle:
I'm currently building a crate for an XML-based protocol based on this PR's source branch as I need pull-based async XML deserialization for my crate. It'd be really helpful if this could be merged so I can publish my work to crates.io. Is there anything I could help with to get this merged?

notgull:
You could use `tokio`'s `spawn_blocking` functionality until this is merged. Alternatively, since `&[u8]` implements `Read`, you could use the asynchronous file system functions to read into a buffer and then pass that buffer to this crate

999eagle:
@notgull sadly the specifics of the protocol I'm dealing with require me to read from a network socket until I have read one full XML element. As my network socket should be async, I need pull-based XML parsing from an `AsyncRead`. Currently I've implemented this by reading into a buffer until the last read bytes look like the end tag corresponding to the first start tag received, but this PR provides a much cleaner solution.

notgull:
What protocol is this?

999eagle:
It's the Greenbone Management Protocol. A vendor-specific protocol that, as far as I can tell, only has an official Python implementation, which also does nothing more than use a pull-based XML parser to read all received data until one full XML element has been read. Yes, I know it's a bit stupid but I can't really change it. The protocol really just sends XML data over a TLS-encrypted TCP stream without any kind of header or footer with the connection being kept open for further requests.

Mingun (changes requested):
>  Is there anything I could help with to get this merged?

@999eagle, Yes, you can. I'd like to see the following changes:
- investigate the possibility to depend on the crate with only `async` trait, without async runtime. Runtime is a quite big thing and I would prefer not to have it because we don't need it. It is fine to have it if there is no other way, but this should be proved
- (**required**) eliminate code duplication either by introducing a trait and different implementations for the sync and async readers, or by using macros
- I prefer to have a `read_event_async` (borrowed version) / `read_event_into_async` (copying into provided buffer version) instead of a separate reader, because I want to introduce a dedicated reader for namespace support and do not want to have 4 readers (non-ns/ns + sync/async). Async abilities seems does not needed a dedicated reader. I'm open for discussion, although
- I prefer to put all async examples/tests to their own files
- tests for internal functions similar to that should be added: https://github.com/tafia/quick-xml/blob/0febc2bbe806cc530a659b2795cdd35489f8d1ac/src/reader.rs#L2591-L2599
  Preferably if it could be added to that group of tests

`async` support definitely won't be in 0.24.0 release, because I plan to release it when I finish with namespaces support which could be soon. 0.25.0 is an option.

Probably it is better to rewrite `async` support from scratch using this PR as inspiration

999eagle:
@Mingun a few thoughts on your points:

- There's the option of using `futures_io::AsyncBufRead` instead of `tokio`'s types, but the traits from `futures` aren't implemented for `tokio`'s types and vice-versa. Also this PR doesn't include the full tokio runtime by default as it only depends on the feature flags `io-util` and `fs`, which is why I think the `tokio`-dependency is okay as it is right now.
- I think sync and async reading are different enough to require two different readers as they require different traits for the underlying reader. One option for implementing this would be to separate reading from parsing. That is, there's an async reader and a sync reader, but both would pass the read bytes on to a parser. As parsing is only CPU-bound as far as I can tell and doesn't involve any reading or waiting on external resources, parsing would always be pureley sync. This also ties in with the next point:
- The parser could be some implemented using a trait to keep the possibility of having a namespaced and a non-namespaced parser. The trait would be sync and both reader implementations could do whatever they need to do to read bytes and once they have a buffer, just call into the sync parser trait. That would allow for all combinations of sync/async and namespaced/non-namespaced while having minimal duplicate code.

What do you think about those ideas? And if I was to start that work, should it be as a new PR or as a change to this one?

I definitely agree with the last two points that async tests should be separate and that more tests are required.

Mingun:
I've learned a bit about current async support in Rust and it seems that there is no a single crate with just the `AsyncRead` / `AsyncWrite` trait. There was some in the past ([`tokio_io`](https://docs.rs/tokio-io/latest/tokio_io/)), but it was then merged into `tokio` crate as [`io` module](https://docs.rs/tokio/latest/tokio/io/index.html). [Different](https://github.com/rust-lang/wg-async/issues/23) [sources](https://rust-lang.github.io/async-book/08_ecosystem/00_chapter.html) says that there are about three different implementations of such traits in [`tokio`](https://lib.rs/crates/tokio), [`futures-rs`](https://lib.rs/crates/futures) and [`async-std`](https://lib.rs/crates/async-std). [Async Book also note](https://rust-lang.github.io/async-book/08_ecosystem/00_chapter.html#determining-ecosystem-compatibility), that there is [a compatibility layer](https://lib.rs/crates/async-compat) between some of them, but without performance analysis. So it should fine to have `tokio`

-------

> I think sync and async reading are different enough to require two different readers as they require different traits for the underlying reader.

Actually, `BufRead` bound required only on two `impl` blocks here:
https://github.com/tafia/quick-xml/blob/0febc2bbe806cc530a659b2795cdd35489f8d1ac/src/reader.rs#L464-L465
and there
https://github.com/tafia/quick-xml/blob/0febc2bbe806cc530a659b2795cdd35489f8d1ac/src/reader.rs#L683-L684

Actually, I had already planned to remove that bound from another places but doubted would that will provide a better ergonomic, but now I have arguments. Perhaps I will make a separate PR with this (and some other) change.

Sync and async readers have the same set of fields, and it is safe to read one event via sync interface and other via async interface -- this would not break anything. So all that we need are a separate `impl` blocks:
```rust
// sync methods
impl<R: BufRead> Reader<R> { /*...*/ }

// async methods
impl<R: AsyncBufRead> Reader<R> { /*...*/ }
```
We even can have a native implementation for each async library, using different `feature`s: `async-tokio`, `async-features`, `async-std`.

So I think, the separate reader is not necessary. For comparison's sake, namespaced reader have additional fields, so a separate `struct` is required. It is also an error to mix calls to the non-namespace aware methods and a namespace-aware one (currently it is [`read_event`](https://github.com/tafia/quick-xml/blob/0febc2bbe806cc530a659b2795cdd35489f8d1ac/src/reader.rs#L510-L512) and [`read_namespaced_event`](https://github.com/tafia/quick-xml/blob/0febc2bbe806cc530a659b2795cdd35489f8d1ac/src/reader.rs#L563-L567)). For example, if you read a `Start` event using `read_namespaced_event` and read an `End` event using `read_event`, the namespaces, declared in the `Start` event would not be end their scope and further parsing could be with errors.

----

> One option for implementing this would be to separate reading from parsing. That is, there's an async reader and a sync reader, but both would pass the read bytes on to a parser. As parsing is only CPU-bound as far as I can tell and doesn't involve any reading or waiting on external resources, parsing would always be pureley sync.

Yes, it makes sense to split reading from parsing where possible. Small code duplications are allowed or may be hidden behind macros.

----

> What do you think about those ideas? And if I was to start that work, should it be as a new PR or as a change to this one?

I think it is better to start a new PR -- some parts of code was changed significantly since this PR was created; I also prefer to have a clean history without fixup commits

999eagle:
Okay, I've started working on this (see the linked PR #417 which I've opened to allow for discussion). For now, I had to use an `async`-specific version of the `reader::XmlSource` trait because the functions need to return futures instead of results. Some of the `impl`-blocks for `Reader` have to be adapted to this now as well.

dralley:
Closing the old PRs in favor of https://github.com/tafia/quick-xml/pull/417

PR #325, opened 2021-10-11
Event::DocType skip initial whitespaces

BlueGreenMagick:
Document that the whitespace right after `<!DOCTYPE` is included when reading, and should be included when writing.

I personally think it may be better to exclude the first whitespace from DocType, but that might be a breaking change, so I added a line to documentation instead.


Mingun:
I think it is better to fix such kinds of problems while we at the 0.x release cycle which is intended for such fixes. Breaking changes is allowed (and expected) between 0.x releases.

Issue #329, opened 2021-10-25
Lifetime issue in new unbuffered logic


stevenroose:
I have been trying out the new alpha version and I think I stumbled upon a kinda-bug. Somehow something that should be possible is not, I think because of the way lifetimes are indicated in some methods. I'm trying to find a fix, but to reproduce the issue, I have the following snippet.

In the snippet, I'm trying to use a `&'x [u8]` and trying to deserialize into a struct that holds a `Cow<'x, str>` with the content of a tag. Somehow this snippet gives a lifetime error. I think the main culprit is the combination of `do_unescape` that equates the lifetime of the slice 
```rust
pub fn do_unescape<'a>(
    raw: &'a [u8],
    custom_entities: Option<&HashMap<Vec<u8>, Vec<u8>>>,
) -> Result<Cow<'a, [u8]>, EscapeError> {
```
together with calling into `do_escape` as follows which actually overwrites the `'a` lifetime with a new local lifetime: `do_unescape(&*self.value, custom_entities).map_err(Error::EscapeError)`

Here is the snippet:

```rust
use std::str;
use std::borrow::Cow;
use std::error::Error;

use quick_xml::Reader;
use quick_xml::events::Event;


pub struct Example<'a> {
	pub content: Cow<'a, str>,
}

/// Convert a UTF8-encoded [Cow<[u8]>] into a [Cow<str>].
fn cow_from_utf8<'a>(cow: Cow<'a, [u8]>) -> Result<Cow<'a, str>, str::Utf8Error> {
	Ok(match cow {
		Cow::Borrowed(v) => Cow::Borrowed(str::from_utf8(v)?),
		Cow::Owned(v) => Cow::Owned(String::from_utf8(v).map_err(|e| e.utf8_error())?),
	})
}

pub fn parse<'a>(xml: &'a impl AsRef<[u8]>) -> Result<Example<'a>, Box<dyn Error>> {
	let mut reader = Reader::from_reader(xml.as_ref());

	match reader.read_event_unbuffered()? {
		Event::Start(x) => match str::from_utf8(x.name())? {
			"Content" => {
				let content = match reader.read_event_unbuffered()? {
					Event::Text(t) => cow_from_utf8(t.unescaped()?)?,
					Event::End(e) if e.name() == x.name() => Cow::Borrowed(""),
					e => return Err(format!("unexpected event '{:?}'", e).into()),
				};
				match reader.read_event_unbuffered()? {
[...]

Issue #328, opened 2021-10-20
Support zero-copy reading in serde deserializer


stevenroose:
It seems that the current usage of Reader requires you to copy a lot of bytes from the reader into the buffer. Why is there no way to get access to slices into the original bytes that were provided?

serde, and serde_json allow for Deserializing borrowed data by borrowing into the original content. I can't find out if quick_xml supports it, but I don't think so.

Could I f.e. parse an XML like this:
```xml
<map>
  <entry>
    <key>test</key>
    <value>testvalue</value>
  </entry>
</map>
```
Into a `HashMap<&str, &str>`? Where the keys and values are references into the original string passed into `Reader::from_str`.

From the method signature 
```rust
pub fn read_event<'a, 'b>(
    &'a mut self,
    buf: &'b mut Vec<u8>
) -> Result<Event<'b>>
```
it seems like all data the events contain is first copied into the buffer and then references there...

dralley:
`read_event()` now behaves as `read_event_unbuffered()` did (although not released yet)

Mingun:
Well, serde part are still not fully zero-copy, and the original request seems to requests this for serde

PR #330, opened 2021-10-25
Do unescaping on Cows instead of slices to fix lifetimes

stevenroose:
Fixes https://github.com/tafia/quick-xml/issues/329 and relates to https://github.com/tafia/quick-xml/issues/328.

The second commit has lifetime indications that could have been inferred, but it's safer to explicitly add them so that accidentally violating them gets noticed more easily.

Also, is there a way in which I can include my example snippet from https://github.com/tafia/quick-xml/issues/329 as a test somehow? Where would I put that?

### DIFF
--- a/src/escapei.rs
+++ b/src/escapei.rs
@@ -126,7 +126,7 @@ fn _escape<F: Fn(u8) -> bool>(raw: &[u8], escape_chars: F) -> Cow<[u8]> {
 
 /// Unescape a `&[u8]` and replaces all xml escaped characters ('&...;') into their corresponding
 /// value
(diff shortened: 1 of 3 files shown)


dralley:
Would `Into<Cow<..>>` work?  I ask because requiring `Cow<..>` for the escape routines would be rather inconvenient

`Into<Cow<..>>` ought to be implemented for `String`, `&str`, and of course `Cow<..>`

dralley:
Also, apologies for the inactivity, the original maintainer has been busy for the past few months

dralley:
This is likely to be addressed soon

Issue #331, opened 2021-10-25
Don't use 'a for the underlying slice's lifetime


stevenroose:
I'd like to make a proposal, just so I don't go implement it without having consensus on it being fine. Since `'a` is kinda the default generic lifetime name, it's kinda annoying that it's also the one used for the "god slice" as I call it, the big `&[u8]` underlying a `Reader`. This makes that when a "ad-hoc temporary" lifetime is chosen, we will either (1) have to come up with another name because `'a` is also used in the same context, or (2) use `'a` with another meaning than all other `'a`s in the code.

So I'd like to propose using some kind of two-character lifetime name to indicate that it's a more permanent and cross-signature lifetime than `'a`. Possible proposed names would be `'sl` (for slice), or `'in` (for input), or `'bf` (for buffer), ...

I'd be wiling to do the work to use it everywhere, it helps me with the work I'm doing with fixing the borrows everywhere.

PR #333, opened 2021-10-26
Name some of the longer-lived lifetimes 

stevenroose:
Based on https://github.com/tafia/quick-xml/pull/330. Closes https://github.com/tafia/quick-xml/issues/331.

Lifetime names like 'a are usually more suitable for "anonymous"
lifetimes that have a limited scope. Lifetimes defined on types and
implementations can sometimes be more significant and it's usual to name
those with multi-character lifetime names.

I chose the following names:
- `'bf`: whenever there is some underlying (usually ephemeral) buffer
  This includes all the events and the ephemeral buffer in buffered
  readers.
- `'inp`: the underlying input bytes for a borrowing reader,
  Note that for events, the `'bf` refers to this buffer too in the case of
  these readers.
- In the `BufferedInput` trait, I used `'bf` for slices it outputs from the
  buffer and `'int` for the lifetime of the internal type, which is equal
  to '`inp` for borrowing readers and the type lifetime of the `BufRead` for
  buffered readers (usually `'static`).

### DIFF
--- a/src/de/escape.rs
+++ b/src/de/escape.rs
@@ -14,16 +14,16 @@ use std::borrow::Cow;
 /// when converting to float, we don't expect any escapable character
 /// anyway
 #[derive(Clone, Debug)]
-pub struct EscapedDeserializer<'a> {
+pub struct EscapedDeserializer<'bf> {
     decoder: Decoder,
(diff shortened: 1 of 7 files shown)


dralley:
Some progress has been made on this

Issue #332, opened 2021-10-25
Lifetime on Attributes is wrong


stevenroose:
So `Attributes` has a generic parameter `'a`, but it doesn't correspond to the same lifetime of the slice inside the reader. It corresponds to the lifetime of the borrow of the reader itself when the `Attributes` object was created.

One can easily see that making this change will break the code:
```diff
diff --git a/src/events/mod.rs b/src/events/mod.rs
index 5e598bd..861ce7b 100644
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -223,12 +223,12 @@ impl<'a> BytesStart<'a> {
     }
 
     /// Returns an iterator over the attributes of this tag.
-    pub fn attributes(&self) -> Attributes {
+    pub fn attributes(&self) -> Attributes<'a> {
         Attributes::new(&self.buf, self.name_len)
     }
 
     /// Returns an iterator over the HTML-like attributes of this tag (no mandatory quotes or `=`).
-    pub fn html_attributes(&self) -> Attributes {
+    pub fn html_attributes(&self) -> Attributes<'a> {
         Attributes::html(self, self.name_len)
     }
 
```

I'm working on fixing that using as little as possible reallocations, which would mean having a double lifetime on `Attributes`.

PR #337, opened 2021-10-26
Correctly use Cow for attribute values to return useful lifetimes

stevenroose:
Closes https://github.com/tafia/quick-xml/issues/332.

Based on other MRs, only last 2 commits are relevant here. Can be rebased if needed.

This implementation unfortunately means a bit more allocations in the case of data that needed unescaping or decoding or so, but I think in general it's ok. It allows to get attributes that borrow into the underlying input slice.

I'm thinking about how we could use the Bytes crate to optimize these allocations. I'm not sure if having an extra `BufferedInput` impl for Bytes would be enough. You can't `Cow` a `Bytes` unfortunately. So I think it'd need some kind of other construction.

### DIFF
--- a/src/de/escape.rs
+++ b/src/de/escape.rs
@@ -14,16 +14,16 @@ use std::borrow::Cow;
 /// when converting to float, we don't expect any escapable character
 /// anyway
 #[derive(Clone, Debug)]
-pub struct EscapedDeserializer<'a> {
+pub struct EscapedDeserializer<'bf> {
     decoder: Decoder,
     /// Possible escaped value of text/CDATA or attribute value
-    escaped_value: Cow<'a, [u8]>,
+    escaped_value: Cow<'bf, [u8]>,
     /// If `true`, value requires unescaping before using
     escaped: bool,
 }
 
-impl<'a> EscapedDeserializer<'a> {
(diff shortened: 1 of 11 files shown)


Mingun:
I'm not sure, that current `Attributes` design is purely wrong. If you draw an analogy with the `Vec` type from the standard library, then `Attributes` corresponds to the `Iter` type -- non-owning iterator, and you need the `IntoIter` type -- owning iterator:

|`Vec`|`quck-xml`|
|----|-----|
|[`[T]::iter()`](https://doc.rust-lang.org/std/primitive.slice.html#method.iter)|[`BytesStart::attributes()`](https://docs.rs/quick-xml/latest/quick_xml/events/struct.BytesStart.html#method.attributes)|
|[`Iter`](https://doc.rust-lang.org/std/slice/struct.Iter.html)|[`Attributes`](https://docs.rs/quick-xml/latest/quick_xml/events/attributes/struct.Attributes.html)|
|[`Vec::into_iter()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_iter)|_nothing_|
|[`IntoIter`](https://doc.rust-lang.org/std/vec/struct.IntoIter.html)|_nothing_|

You trying to combine both types into one struct, but it is probably better implement them as two distinct types.

PR #348, opened 2022-01-05
Switch from libtest to criterion for benchmarks, add new benchmarks for attribute access, do some optimizations

dralley:
Benefits to Criterion:

* more statistically sound
* works on stable Rust (no nightly features)
* auto-generates a lot of useful plots and statistical information
* more momentum / ecosystem

`try_get_attribute()` is ~50% faster after these changes

Screenshot from 2022-01-04 23-46-10


src/events/attributes.rs
```diff
@@ -409,7 +409,7 @@ impl<'a> Iterator for Attributes<'a> {
                 .consumed
                 .iter()
                 .filter(|r| r.len() == end_key - start_key)
-                .find(|r| self.bytes[(*r).clone()] == self.bytes[start_key..end_key])
+                .find(|r| self.bytes[r.start..r.end] == self.bytes[start_key..end_key])
```
dralley, line 412: Not actually relevant to performance but it's a little cleaner.

src/events/attributes.rs
```diff
@@ -364,7 +364,7 @@ impl<'a> Iterator for Attributes<'a> {
             return None;
         }
 
-        let mut bytes = self.bytes.iter().enumerate().skip(self.position);
+        let mut bytes = (self.position..).zip(self.bytes[self.position..].iter());
```
dralley, line 367: Skip straight to the section of the buffer we're interested in instead of iterating over the preceding bytes.

src/events/mod.rs
```diff
@@ -353,7 +353,7 @@ impl<'a> BytesStart<'a> {
         &'a self,
         attr_name: N,
     ) -> Result<Option<Attribute<'a>>> {
-        for a in self.attributes() {
+        for a in self.attributes().with_checks(false) {
```
dralley, line 356: My thinking here is, the code already stops at the first instance of the correct key and returns early, so there's not much point doing checks.  And if you're asking for one specific element you probably don't care so much that others might be in violation of the rules.

This was about 50% faster on top of the other improvement.

dralley:
Comparison w/ xml_rs
Screenshot from 2022-01-05 00-11-15

dralley:
@tafia Will you have time to review open PRs any time soon?  Not this particular PR necessarily, just in general.

dralley:
@tafia Would it be possible to at push a new alpha release at least?  I need a few of the fixes since the last one and I can't release to crates.io with local patches.

src/events/attributes.rs
Mingun: I'm wonder: doesn't the iterator just jump to the right position for slices?

dralley: It's been a few months but I remember isolating this as having a performance impact independent from the other change.

src/events/attributes.rs
```diff
@@ -355,7 +444,7 @@ impl<'a> Iterator for Attributes<'a> {
             ($key:expr, $val:expr) => {
                 Some(Ok(Attribute {
                     key: &self.bytes[$key],
-                    value: Cow::Borrowed(&self.bytes[$val]),
+                    value: normalize_attribute_value(Cow::Borrowed(&self.bytes[$val])),
```
Mingun: I think it is better to provide an API method on `Attribute` to get normalized value only when its needed

src/events/attributes.rs
```diff
@@ -331,6 +331,95 @@ impl<'a> From<(&'a str, &'a str)> for Attribute<'a> {
...
+//   * For a white space character (#x20, #xD, #xA, #x9), append a space character (#x20) to the normalized value.
+//   * For another character, append the character to the normalized value.
+//
+// If the attribute type is not CDATA, then the XML processor MUST further process the normalized attribute value by discarding any leading and trailing space (#x20) characters,
+// and by replacing sequences of space (#x20) characters by a single space (#x20) character.
+//
+// Note that if the unnormalized attribute value contains a character reference to a white space character other than space (#x20), the normalized value contains the referenced
+// character itself (#xD, #xA or #x9). This contrasts with the case where the unnormalized value contains a white space character (not a reference), which is replaced with a
+// space character (#x20) in the normalized value and also contrasts with the case where the unnormalized value contains an entity reference whose replacement text contains a
+// white space character; being recursively processed, the white space character is replaced with a space character (#x20) in the normalized value.
```
Mingun, line 348: It would be nice to add a reference to the specification, as well as convert this into doc comment on the method on `Attribute`

src/events/attributes.rs
dralley: The spec-compliant value needs to be what you get by default, but I would be OK with a `raw_value()`.

dralley:
FYI I'm moving this conversation to the dedicated PR.  I accidentally pushed this branch since I was using the benchmarks locally.

src/events/attributes.rs
Mingun: But if you just want to iterate over some attributes to skip them (for example, you're search `xmlns` attributes), you'll do unnecessary work.

dralley:
https://github.com/Mingun/fast-xml/pull/5

Can be merged back in if / when @tafia returns.

PR #356, opened 2022-02-06
Fix and extend documentation on XML Declaration methods

Mingun:
Documentation erroneously claimed that quotation marks would be part of the value

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -387,7 +387,59 @@ impl<'a> BytesDecl<'a> {
         BytesDecl { element: start }
     }
 
-    /// Gets xml version, including quotes (' or ")
+    /// Gets xml version, excluding quotes (' or ")
+    ///
+    /// According to the [grammar], the version *must* be the first thing in the declaration.
+    /// This method tries to extract the first thing in the declaration and return it.
+    /// In case of multiple attributes value of the first one is returned.
+    ///
+    /// If version is missed in the declaration, or the first thing is not a version,
+    /// [`Error::XmlDeclWithoutVersion`] will be returned.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::borrow::Cow;
+    /// use quick_xml::Error;
+    /// use quick_xml::events::{BytesDecl, BytesStart};
+    ///
+    /// // <?xml version='1.1'?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" version='1.1'", 0));
+    /// assert_eq!(
+    ///     decl.version().unwrap(),
+    ///     Cow::Borrowed(b"1.1".as_ref())
+    /// );
+    ///
+    /// // <?xml version='1.0' version='1.1'?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" version='1.0' version='1.1'", 0));
+    /// assert_eq!(
+    ///     decl.version().unwrap(),
+    ///     Cow::Borrowed(b"1.0".as_ref())
+    /// );
+    ///
+    /// // <?xml encoding='utf-8'?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" encoding='utf-8'", 0));
+    /// match decl.version() {
+    ///     Err(Error::XmlDeclWithoutVersion(Some(key))) => assert_eq!(key, "encoding".to_string()),
+    ///     _ => assert!(false),
+    /// }
+    ///
+    /// // <?xml encoding='utf-8' version='1.1'?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" encoding='utf-8' version='1.1'", 0));
+    /// match decl.version() {
+    ///     Err(Error::XmlDeclWithoutVersion(Some(key))) => assert_eq!(key, "encoding".to_string()),
+    ///     _ => assert!(false),
+    /// }
+    ///
+    /// // <?xml?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b"", 0));
+    /// match decl.version() {
+    ///     Err(Error::XmlDeclWithoutVersion(None)) => {},
+    ///     _ => assert!(false),
+    /// }
+    /// ```
+    ///
+    /// [grammar]: https://www.w3.org/TR/xml11/#NT-XMLDecl
     pub fn version(&self) -> Result<Cow<[u8]>> {
         // The version *must* be the first thing in the declaration.
         match self.element.attributes().next() {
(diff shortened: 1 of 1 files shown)


src/events/mod.rs
```diff
@@ -387,7 +387,59 @@ impl<'a> BytesDecl<'a> {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::borrow::Cow;
+    /// use quick_xml::Error;
+    /// use quick_xml::events::{BytesDecl, BytesStart};
+    ///
+    /// // <?xml version='1.1'?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" version='1.1'", 0));
```
dralley, line 407: For the sake of documenting the most convenient / foolproof method, maybe use [`BytesDecl::new`](https://docs.rs/quick-xml/0.22.0/quick_xml/events/struct.BytesDecl.html#method.new) instead

src/events/mod.rs
```diff
@@ -387,7 +387,59 @@ impl<'a> BytesDecl<'a> {
...
+    /// // <?xml version='1.0' version='1.1'?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" version='1.0' version='1.1'", 0));
+    /// assert_eq!(
+    ///     decl.version().unwrap(),
+    ///     Cow::Borrowed(b"1.0".as_ref())
+    /// );
+    ///
+    /// // <?xml encoding='utf-8'?>
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" encoding='utf-8'", 0));
+    /// match decl.version() {
```
dralley, line 422: I'm neutral on these examples of the error being raised, you described it well enough in the description.  Just providing a couple examples of XML declarations that are invalid seems sufficient.

For test purposes though these are nice to have.  If this isn't already covered by the tests and you did decide to remove them, they could be unit tests.

If you would rather keep them, I think this can be simplified to some variation of `assert!(matches!(decl.version(), Err(Error::XmlDeclWithoutVersion(_))))`

src/events/mod.rs
```diff
@@ -404,7 +456,41 @@ impl<'a> BytesDecl<'a> {
...
+    /// name may not correspond to the grammar.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::borrow::Cow;
+    /// use quick_xml::Error;
+    /// use quick_xml::events::{BytesDecl, BytesStart};
+    ///
+    /// // <?xml version='1.1'?>
```
dralley, line 474: Same suggestions

src/events/mod.rs
```diff
@@ -419,7 +505,41 @@ impl<'a> BytesDecl<'a> {
...
-    /// Gets xml standalone, including quotes (' or ")
+    /// Gets xml standalone, excluding quotes (' or ")
+    ///
+    /// Although according to the [grammar] standalone flag must appear after `"version"`
+    /// and `"encoding"`, this method does not check that. The first occurrence of the
+    /// attribute will be returned even if there are several. Also, method does not
+    /// restrict symbols that can forming the value, so the returned flag name may not
+    /// correspond to the grammar.
+    ///
+    /// # Examples
```
dralley, line 516: Same suggestions.

src/events/mod.rs
```diff
@@ -387,7 +387,59 @@ impl<'a> BytesDecl<'a> {
...
+    /// match decl.version() {
+    ///     Err(Error::XmlDeclWithoutVersion(None)) => {},
+    ///     _ => assert!(false),
+    /// }
+    /// ```
+    ///
+    /// [grammar]: https://www.w3.org/TR/xml11/#NT-XMLDecl
     pub fn version(&self) -> Result<Cow<[u8]>> {
         // The version *must* be the first thing in the declaration.
         match self.element.attributes().next() {
```
dralley, line 445: This suggestion has nothing to do with your PR, but the implementations of all these methods can be switched to use [`BytesStart::try_get_attribute()`](https://github.com/tafia/quick-xml/blob/521d72a28449eaa0fcb6a2cbeaead72921d92b81/src/events/mod.rs#L352-L364=) instead.

As a bonus it will get a speedup from my PR whenever it gets merged. https://github.com/tafia/quick-xml/pull/348

src/events/mod.rs
```diff
@@ -404,7 +456,41 @@ impl<'a> BytesDecl<'a> {
...
+    /// let decl = BytesDecl::from_start(BytesStart::borrowed(b" encoding='something_WRONG' encoding='utf-8'", 0));
+    /// match decl.encoding() {
+    ///     Some(Ok(Cow::Borrowed(encoding))) => assert_eq!(encoding, b"something_WRONG"),
+    ///     _ => assert!(false),
+    /// }
+    /// ```
+    ///
+    /// [grammar]: https://www.w3.org/TR/xml11/#NT-XMLDecl
     pub fn encoding(&self) -> Option<Result<Cow<[u8]>>> {
         for a in self.element.attributes() {
```
dralley, line 495: For these two methods you can probably simplify the implementation by using [`BytesStart::try_get_attribute()`](https://github.com/tafia/quick-xml/blob/521d72a28449eaa0fcb6a2cbeaead72921d92b81/src/events/mod.rs#L352-L364=) instead.

And as a bonus it will get a speedup from my PR whenever it gets merged. https://github.com/tafia/quick-xml/pull/348

You can't do that for `::version()` above because of the requirement to check that the version attribute comes first.  But you can still do some optimizatoin by setting `.with_checks(false)` manually.

src/events/mod.rs
Mingun: Yes, that of course idiomatic method, but this three examples shows, how input transformed into `BytesDecl`. Because other tho examples shows, how semantically incorrect input is parsed, so as not to interrupt continuity the same methods is called in all three situations. It is possible, however, mention the right method

src/events/mod.rs
Mingun: I think is is better to leave it as is. I would personally prefer more complete examples, than less. Current examples shows what data to expect in each variant.

src/events/mod.rs
Mingun: Good catch! Then I think you should just include that change into your PR, because it is already about optimisations

dralley: Fair point, I'll do that

src/events/mod.rs
Mingun: Besides, I think that [BytesDecl::new](https://docs.rs/quick-xml/0.22.0/quick_xml/events/struct.BytesDecl.html#method.new) are the thing that should be changed. That too unsafe for public API, because it can generate invalid XML. I think, it will be better to change it to

```rust
pub enum Version {
  V1_0,
  V1_1,
  Other(String),
}

pub fn new(
  version: Version,
  // Option<Encoding> would be better, but this will require "encoding" feature
  encoding: Option<String>,
  standalone: Option<bool>,
);
```
instead

src/events/mod.rs
Mingun: I think, it will be better. The `Result<Option<...>>` seems more idiomatic and opens a door for an internal changes that could require to return an error even if no attributes defined, if that changes will be necessary sometime

dralley: Separate PR: https://github.com/tafia/quick-xml/pull/378/

src/events/mod.rs
dralley: I meant each individual test can be simplified to that one-liner.

Mingun:
Merged in https://github.com/Mingun/fast-xml/commit/9d241bb90369f46ad53ad50e6980926b5cdf9b4d

PR #357, opened 2022-02-12
Fix an error in reading `<!doctype >` definition from `&[u8]` (in lower case)

Mingun:
Bug was since merging 8d3cf19df74f143a15822a5723cb5bd361fb4310 (zero-copy parsing)
Lowercased `doctype` was introduced in 3c3a82cdae6db29003e065dc368e7e62a28286f4 (html5 support)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,7 @@
   from the attribute and element names and attribute values
 - fix: allow to deserialize `unit`s from text and CDATA content.
   `DeError::InvalidUnit` variant is removed, because after fix it is no longer used
+- fix: allow lowercase `<!doctype >` definition (used in HTML 5) when parse document from `&[u8]`
 
 ## 0.23.0-alpha3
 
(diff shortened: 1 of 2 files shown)


dralley:
For `Attributes` there is an "html mode" bool which you have to toggle if you want the parser to accept the looser format.  That seems like it might be a little bit overkill for this particular scenario but I want to call it out.  If there are more situations like this, perhaps the same should be done.

https://github.com/tafia/quick-xml/blob/516f1ea7178e179ed7c9be9d6efbb4221242aa60/src/events/attributes.rs#L25=

Libxml2 does something similar

https://gitlab.gnome.org/search?group_id=8&project_id=1665&repository_ref=&scope=blobs&search=ctxt-%3Ehtml&snippets=false

Mingun:
Yeah, this requires more tests, but in that particular PR I fix out-of-sync implementation for a borrowing and a non-borrowing readers

Mingun:
Merged in https://github.com/Mingun/fast-xml/commit/bad11787756de4c700bda3433827c53f9526e438

Issue #346, opened 2021-12-17
Serialization of enums adds extra tags


Storyyeller:
Currently, we have some types using quick_xml and serde to parse XML, which works well. However, we want to also serialize them back to XML, and the XML that quick_xml produces differs from the original (expected) format, and I can't figure out any way to fix it.

Our type definitions look like this:

```rust
#[derive(Clone, Debug, Deserialize, Serialize)]
pub struct Contactinfo {
    pub bugurl: String,
}

#[derive(Clone, Debug, Deserialize, Serialize)]
pub struct Include {
    pub name: String,
    #[serde(default)]
    pub groups: String,
}

#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(rename_all = "kebab-case")]
pub enum ManifestChild {
    Include(Include),
    Contactinfo(Contactinfo),

    #[serde(other)]
    Unknown,
}

#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(rename_all = "kebab-case")]
pub struct Manifest {
    #[serde(default, rename = "$value")]
    pub children: Vec<ManifestChild>,
}
```

And we can parse XML with it just fine:

```rust
let text = r#"
<?xml version="1.0" encoding="UTF-8"?>
<manifest>
  <include name="bar"/>
  <contactinfo bugurl="foo"/>
  <include name="hello" groups="world"/>
</manifest>"#;

    let result = quick_xml::de::from_str(text);
    let m: Manifest = result.unwrap();

    let s = quick_xml::se::to_string(&m).unwrap();
    println!("{}", s);
```

However, the output XML is in the wrong format, wrapping all the enum contents in an extra tag. How can we get rid of this and make it serialize in the same way as the input xml? Also, why is Manifest capitalized in the output when it isn't in the input? It seems like serialization ignores our `#[serde(rename_all = "kebab-case")]` annotations!

Here is what the output looks like with quick_xml 0.23alpha3:

```xml
<Manifest><include><Include name="bar"/></include><contactinfo><Contactinfo bugurl="foo"/></contactinfo><include><Include name="hello" groups="world"/></include></Manifest>
```

spikespaz:
Really need this to be fixed. Currently this is the best-working serializer for XML but it still doesn't cut the cake.

Storyyeller:
So far the best workaround I came up with involved writing a combination of four wrapper types with hundreds of lines of boilerplate to change the way it serializes.

spikespaz:
[...]

PR #359, opened 2022-02-16
Added '$struct' tag to identify structs in newtype enum variants.

gimait:
Trying out the serialization options included in quick-xml I came across an issue with the serialization of newtype variants in enums, where structs would have duplicated nodes when serialized. This relates to open issues like #183 and #346.

I believe there might be a better way to do this, but a very basic concept I came up with is to simply mark which enum variants contain a struct, and then skip adding an extra node for such variant.

This makes it for me for now, and maybe some others could benefit from it until we find the right way to do it.
I would like to look into an automatic way of doing this if someone has any ideas of how I could detect automatically if the given newvariant is a struct.

Let me know what you think, any feedback is welcome :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,8 +213,8 @@ fn crates_io() -> Result<Html, DeError> {
 
 ### Credits
 
-This has largely been inspired by [serde-xml-rs](https://github.com/RReverser/serde-xml-rs). 
-quick-xml follows its convention for deserialization, including the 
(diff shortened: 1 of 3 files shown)


gimait:
I'm doing some more work on this for a project I'm working on, and in the meanwhile, I'm seeing other features that I'm missing for serde. I'll close this PR for now and create a new more complete one later on, including a proposal to include those features.

PR #367, opened 2022-03-08
Add tests for `BufferedInput` and fix couple of bugs

### LINKED ISSUE #334: What is unused method BufferedInput::input_borrowed for?
That method is unused. It has this definition:
```rust
    fn input_borrowed(event: Event<'bf>) -> Event<'int>;
```
And it's implemented trivially for both implementations. (Borrowing input just returns self and buffered input returns into_owned.)

It's introduced in this commit:
```
commit 8d3cf19df74f143a15822a5723cb5bd361fb4310
Author: Andrei Vasiliu <whyte.vuhuni@gmail.com>
Date:   Fri Apr 17 14:18:51 2020 +0300

    Add borrowing support to deserializer
    
    Co-authored-by: Sergio Benitez <sb@sergio.bz>
```

The method intoduces an extra lifetime definition that seems like it could be removed if that method wouldn't exist.

### LINKED ISSUE #344: Panic when parsing malformed XML: "Proper bang start guaranteed by read_bang_elem"
Reproduction:

```rust
fn main() {
    let doc = "<!D>";
    let mut reader = quick_xml::Reader::from_str(doc);
    let mut buf = Vec::new();
    reader.read_event(&mut buf);
    reader.read_event(&mut buf);
}
```

Tested on 521d72a28449eaa

Mingun:
This PR supersedes #357 because it refactors corresponding place.

This PR:
- fixes #344 by replacing panic handler by returning an error because it is possible to get there as not of preconditions is checked in the `BufferedInput` implementations
- closes #334 by removing unused method
- unify some code for parsing from buffered reader and borrowed reader (`&[u8]`)
- add tests that proves that both readers behaves the same way
- as a result of tests fixes positions reported by borrowed reader (I assume that positions from buffered reader are correct because it was introduced earlier)

Number of commits a bit more that I could prefered in the PRs, but most of them are intended to show refactoring process in order to anybody could check that behavior is not changed -- I suppose it is obvious from the commits. Therefore, I recommend review PR by commit basis. Also, each commit in a fully buildable state except 9582fe504abe32ae9de1642b3110315478398efe which is fixed by the following commit and failed by design.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -20,6 +20,9 @@
 - fix: allow to deserialize `unit`s from text and CDATA content.
   `DeError::InvalidUnit` variant is removed, because after fix it is no longer used
 - fix: allow lowercase `<!doctype >` definition (used in HTML 5) when parse document from `&[u8]`
+  ([#357](https://github.com/tafia/quick-xml/pull/357))
+- test: add tests for consistence behavior of buffered and borrowed readers
+  ([#367](https://github.com/tafia/quick-xml/pull/367))
 
 ## 0.23.0-alpha3
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -23,6 +23,8 @@
   ([#357](https://github.com/tafia/quick-xml/pull/357))
 - test: add tests for consistence behavior of buffered and borrowed readers
   ([#367](https://github.com/tafia/quick-xml/pull/367))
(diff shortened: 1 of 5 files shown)


Mingun:
Merged in https://github.com/Mingun/fast-xml/commit/9770634158834210257a3d517a2a48ec18cf42f4

PR #370, opened 2022-03-14
Add CDATA to match after peek in next_key_seed in MapAccess

pfrom-vizio:
We found that when the text for a node was wrapped in a CDATA tag, it was not consistently being used in MapAccess. It seems that adding that to the match in `next_key_seed` after peeking ensures the text in CDATA tag are deserialized as expected.


Mingun:
Actually, CDATA handling missed not only here -- I already fixed most cases in my not-yet-published branch. I'll push it the next few days

Mingun:
Because fast-xml merged back into quick-xml, this fix will be available in the next release which would be soon.

Merged in 9795a68c9a7db81a54157bb98ba8b6cbb87f0659

PR #374, opened 2022-03-21
Fix CDATA handling

### LINKED ISSUE #311: CDATA deserialization
It seems that CDATA parsing performed incorrectly. CDATA should contain logically unescaped content (i.e. you don't need to unescape it). So the string `<![CDATA[escaped&#x20;string]]>` should be parsed into string `escaped&#x20;string` in contrast to the ordinal text content where the result should be `escaped string`.

According to the https://www.w3.org/TR/xml/#sec-cdata-sect CDATA sections cannot contain any escape sequences. I've checked several online validators (for example https://jsonformatter.org/xml-parser) and they agree with that.

However, quick-xml has different opinion. This test seems invalid:
https://github.com/tafia/quick-xml/blob/516f1ea7178e179ed7c9be9d6efbb4221242aa60/tests/unit_tests.rs#L179-L184

As a consequence, all CDATA processing in the `Deserializer` is broken because it is simply handled the same as the `Text` event, and in several places handling of the CDATA just absent.

I preparing a PR with these fixes, but because I'm not working with XML very often, I'm not sure that I'm not wrong. So I open this issue to get feedback.

Mingun:
Fixes #311, closes #370.

Now `CData` event has its own dedicated type `BytesCData` instead of `BytesText`. New type can be converted to `BytesText` by calling `.escape()` method, so migrating should be easy, but it is better to review usages, because usually you want to have _unescaped_ data and CDATA section already provide them.

The API a bit dangerous because it not checked for forbidden sequence `]]>` -- you can create a `BytesCData` containing it. Because this is general problem with other event content types, I leaved it now as is. This API should be fixed in another PR.

This PR changes how raw bytes are deserialized, for example, to a `Vec<u8>`. Before such field always get a unmodified content from the reader -- it means that you can consider raw bytes types as a way to access to raw reader content in your structs. This is dangerous practice, because other formats do not behave like this. Actually, I would prefer to prohibit (de)serializing (into) `Vec<u8>` and similar at all. This is conceptually right, but maybe unpractical. So now bytes deserialized similarly to strings -- text content unescaped, CDATA content remained unchanged.

I'm thinking about providing encoding handler to a (de)serializer, that should be used to encode / decode binary data to strings, that will be stored in XML, for example, hex, or base64, and by default (`handler == None`) emit an error on attempt (de)serializing binary data. But this is for another PR.

Unescaping takes into account only default XML entities, namely `&lt;`, `&gt;`, `&amp;`, `&apos;`, and `&quot;`. Adding support for document-defined entities is the purpose for another PR.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,7 @@
   from the attribute and element names and attribute values
 - fix: allow to deserialize `unit`s from text and CDATA content.
   `DeError::InvalidUnit` variant is removed, because after fix it is no longer used
+- test: add tests for trivial documents (empty / only comment / `<root>...</root>` -- one tag with content)
 
 ## 0.23.0-alpha3
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,6 +21,13 @@
   `DeError::InvalidUnit` variant is removed, because after fix it is no longer used
 - test: add tests for trivial documents (empty / only comment / `<root>...</root>` -- one tag with content)
 - fix: CDATA was not handled in many cases where it should
+- fix: do not unescape CDATA content because it never escaped by design
+  ([#311](https://github.com/tafia/quick-xml/issues/311)).
+
+  NOTE: now text content when deserialized into bytes (`Vec<u8>` / `&[u8]`), also unescaped.
+  It is impossible to get a raw XML data in bytes buffer. Actually, deserializing of bytes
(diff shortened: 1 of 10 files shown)


Mingun:
Merged in https://github.com/Mingun/fast-xml/commit/9795a68c9a7db81a54157bb98ba8b6cbb87f0659

PR #369, opened 2022-03-12
Add documentation for mapping from XML to Rust used by deserializer

Mingun:
This is the my vision of further evolution of the serde integration in this crate. Some parts of this is discussible or maybe even impossible to implement -- this is the first iteration of what I would like to see. For now I'm making this draft PR to:
- share my vision
- invitation to discussion
- creating a roadmap for necessary fixes (marked with `FIXME` in doctests)

You can get a rustdoc documentation by running
```console
cargo doc --features serialize --open
```
in the crate root an navigate to `quick_xml::de` module.

Below the (approximately) rendered version of this proposal:

# Mapping XML to Rust types

Type names are never considered when deserializing, so you could name your
types as you wish. Other general rules:
- `struct` field name could be represented in XML only as attribute name or
  element name.
- `enum` variant name could be represented in XML only as attribute name or
  element name.
- the unit struct, unit type `()` and unit enum variant can be deserialized
  from any valid XML content:
  - attribute and element names
  - attribute and element values
  - text or CDATA content
- when deserializing attribute names have precedence over element names.
  So if your XML have both attribute and element named equally, the Rust
  field/variant will be deserialized from the attribute.

> NOTE: examples, marked with `FIXME:` do not work yet -- any PRs that fixes
> that are welcome! The message after marker is a test failure message.
> Also, all that tests are marked with an `ignore` option, although their
> compiles. This is by intention, because rustdoc marks such blocks with
> an exclamation mark unlike `no_run` blocks.

<table>
<thead>
<tr><th>To parse all these XML's...</th><th>...use that Rust type</th></tr>
</thead>
<tbody>
<tr>
<td>
Root tag name do not matter

```xml
<any-tag one="..." two="..."/>
```
```xml
<any-tag>
  <one>...</one>
  <two>...</two>
</any-tag>
```
```xml
<any-tag one="...">
  <two>...</two>
</any-tag>
```
> NOTE: such XML's are NOT supported because deserializer will always
> report a duplicated field error:
> ```xml
> <any-tag field="...">
>   <field>...</field>
> </any-tag>
> ```
</td>
<td>

All these struct can be used to deserialize from specified XML depending on
amount of information that you want to get:

```rust
// Get both elements/attributes
struct AnyName {
  one: T,
  two: U,
}
```
```rust
// Get only one element/attribute, ignore other
struct AnyName {
  one: T,
}
```
```rust
// Ignore all attributes/elements
// You can also use the `()` type (unit type)
struct AnyName;
```

[...]

PR #376, opened 2022-03-23
Add support for XML Schema `xs:list` type

Mingun:
This PR allows deserializing `Vec` and other sequences from space-separated lists, as [defined](https://www.w3.org/TR/xmlschema11-2/#atomic-vs-list) by XML Schema specification. So now this is possible:
Deserialize
```rust
struct AnyName {
  // Will contain vec!["first", "second", "third"]
  attribute: Vec<String>,
}
```
from
```xml
<any-tag attribute="first second third">
```

This PR is based on #374, because before parsing text/CDATA content I will need to fix CDATA handling. It also cancels change in the bytes deserializing: now it is again possible to get a raw content of the parsed buffer using deserializing into `Vec<u8>` / `&[u8]`. This is not much useful in my opinion, although, so I put a TODO to look at this later.

### DIFF
--- a/src/de/byte_buf.rs
+++ b/src/de/byte_buf.rs
@@ -1,11 +1,12 @@
 //! Helper types for tests
 
+use crate::utils::write_byte_string;
 use serde::de::{self, Deserialize, Deserializer, Error};
 use std::fmt;
 
 /// Wrapper around `Vec<u8>` that deserialized using `deserialize_byte_buf`
 /// instead of vector's generic `deserialize_seq`
-#[derive(Debug, PartialEq)]
+#[derive(PartialEq)]
 pub struct ByteBuf(pub Vec<u8>);
 
 impl<'de> Deserialize<'de> for ByteBuf {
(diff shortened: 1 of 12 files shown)


src/de/map.rs
```diff
@@ -166,12 +178,36 @@ impl<'de, 'a, R: BorrowingReader<'de>> de::MapAccess<'de> for MapAccess<'de, 'a,
...
                 let decoder = self.de.reader.decoder();
                 match self.next_attr()? {
-                    Some(a) => seed.deserialize(EscapedDeserializer::new(a.value, decoder, true)),
+                    Some(a) => {
+                        //FIXME: we have to clone value because of wrong lifetimes on `a`
+                        // It should be bound to the input lifetime, but it instead bound
+                        // to a deserializer lifetime
+                        let value: Vec<_> = a.value.into_owned();
+                        seed.deserialize(SimpleTypeDeserializer::new(value.into(), true, decoder))
+                    }
```
Mingun, line 187: This is a temporary solution. I'll remove this clone in next PRs, when I figure out how to do that. In any case, struct / map deserialization should be reviewed to match expectations that I documenting in #369

src/de/byte_buf.rs
```diff
@@ -35,9 +36,15 @@ impl<'de> Deserialize<'de> for ByteBuf {
     }
 }
 
+impl fmt::Debug for ByteBuf {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write_byte_string(f, &self.0)
```
dralley, line 41: `write_byte_string` could use some documentation

src/de/var.rs
```diff
@@ -36,6 +36,8 @@ where
         let decoder = self.de.reader.decoder();
         let de = match self.de.peek()? {
             DeEvent::Text(t) => EscapedDeserializer::new(Cow::Borrowed(t), decoder, true),
+            // Escape sequences does not processed inside CDATA section
```
dralley, line 39:
```suggestion
            // Escape sequences are not processed inside CDATA section
```

Mingun:
GitHub thinks that this PR coming from https://github.com/Mingun/fast-xml/tree/xs-list repo and I cannot change that, so I'll close it and make a new PR

PR #378, opened 2022-04-02
Simplify named attribute access code on BytesDecl

dralley:


src/events/mod.rs
```diff
@@ -390,48 +390,34 @@ impl<'a> BytesDecl<'a> {
...
             }
+            // error parsing attributes
+            Some(Err(e)) => Err(e),
+            // no attributes
             None => Err(Error::XmlDeclWithoutVersion(None)),
         }
     }
 
     /// Gets xml encoding, including quotes (' or ")
     pub fn encoding(&self) -> Option<Result<Cow<[u8]>>> {
```
dralley, line 408: @tafia, I was discussing with @Mingun whether this signature should be changed to `<Result<Option<Cow<[u8]>>` instead of the way it is currently.  That feels more idiomatic and also more consistent with `version()`.  What do you think?

Mingun (approved):

dralley:
Need these changes on my other PR...

PR #379, opened 2022-04-03
Properly normalize attribute values

### LINKED ISSUE #371: Attribute values aren't normalized properly
Section 3.3.3 of the XML spec (the attribute value normalization section) defines that 

> 
> For each character, entity reference, or character reference in the unnormalized attribute value, beginning with the first and continuing to the last, do the following:
> 
>   *  For a character reference, append the referenced character to the normalized value.
> 
>   *  For an entity reference, recursively apply step 3 of this algorithm to the replacement text of the entity.
> 
>   *  For a white space character (#x20, #xD, #xA, #x9), append a space character (#x20) to the normalized value.
> 
>   *  For another character, append the character to the normalized value.
> 
> If the attribute type is not CDATA, then the XML processor MUST further process the normalized attribute value by discarding any leading and trailing space (#x20) characters, and by replacing sequences of space (#x20) characters by a single space (#x20) character.
> 
> Note that if the unnormalized attribute value contains a character [reference](https://www.w3.org/TR/xml/#dt-entref) to a white space character other than space (#x20), the normalized value contains the referenced character itself (#xD, #xA or #x9). This contrasts with the case where the unnormalized value contains a white space character (not a reference), which is replaced with a space character (#x20) in the normalized value and also contrasts with the case where the unnormalized value contains an entity reference whose replacement text contains a white space character; being recursively processed, the white space character is replaced with a space character (#x20) in the normalized value.
> 
> All attributes for which no declaration has been read SHOULD be treated by a non-validating processor as if declared CDATA.
> 
> It is an error if an attribute value contains a reference to an entity for which no declaration has been read.

TL;DR: Bullet point 3 states that 0xD, 0xA, and 0x9 (`\r`, `\n` and `\t`) within an attribute value ought to be treated as whitespace and replaced with a whitespace character (0x20), unless the character was explicitly referenced.

https://www.w3.org/TR/xml/#AVNormalize

Here's a half-written test which demonstrates the failure:

```Rust
#[test]
fn test_attribute_value_normalization() {
    let mut r = Reader::from_str("<a attr=\"tab character\tor newline\nor return\rshould be replaced\">");
    r.trim_text(true);
    let mut buf = Vec::new();
    match r.read_event(&mut buf) {
        Ok(Start(e)) => {
            let mut attrs = e.attributes();
            match attrs.next() {
                Some(Ok(attr)) => assert_eq!(attr, ("attr".as_bytes(), "tab character or newline or return should be replaced".as_bytes()).into()),
                x => panic!("expected attribute 'attr', got {:?}", x),
            }
        }
        x => panic!("expected <a attr='>'>, got {:?}", x),
    }
}
```

```
---- test_attribute_value_normalization stdout ----
thread 'test_attribute_value_normalization' panicked at 'assertion failed: `(left == right)`
  left: `Attribute { key: "attr", value: Borrowed("tab character0x9or newline0xAor return0xDshould be replaced") }`,
 right: `Attribute { key: "attr", value: Borrowed("tab character or newline or return should be replaced") }`', tests/unit_tests.rs:850:35
```

dralley:
closes #371


dralley:
Suggestions :

* Move this functionality directly to `Attribute`
* Provide a fast path to get the raw value

TODO:

* Character reference & entity reference substitution with associated error handling

dralley:
I plan to continue working on this over the next week or two

dralley:
@Mingun Questions:

Currently we have the functions `unescaped_value`, `unescaped_value_with_custom_entities` and their `decode` equivalents, that do the escaping part but don't implement the rest of the spec.  I'm not sure I see any reason for those to continue to exist as far as XML is concerned, but for HTML it makes some sense, as HTML only seems to do unescaping without any other normalization of the value.

1) Does that sound accurate / align with your knowledge
2) Do you think it would make more sense to stick with functional names, like `normalized_value` / `unescaped_value` and rely on the documentation to tell users what they ought to be using, or switch  more descriptive names such as `html_value` / `xml_value`
3) The behavior of `Attributes` depends on whether or not `.html` is set, should we look at doing something similar here, or would that not be worth the additional complication

Mingun:
1. I haven't studied a situation about HTML attributes, therefore, I rely on your understanding of the situation. Then, if you include references to relevant resources in the documentation, I'll be able to learn something about that
2. I think that functional names are better
3. Probably we should inverse things and introduce different types for XML / HTML attributes, then implement only relevant methods on each. This will also solve some unpleasant things in the current API -- we can change `html`ity of the attributes in the middle of iteration that probably could to lead to sophisticated bugs. I'd would like to avoid such dangerous usage.

dralley:
I'm basically going off of the lack of any kind of discussion of attribute value normalization in the HTML living spec, and this discussion on stackoverflow

https://html.spec.whatwg.org/multipage/dom.html#attributes
https://stackoverflow.com/questions/63906320/html5-attribute-value-normalization

It makes sense when you think about it - with HTML supporting name-only attributes, not strictly deliminated with quotes, as a natural consequence most of the things which "normalization" would remove would impact the parsing of the attributes themselves.  So normalization is pointless, unescaping is all you can really do, with HTML.

benches/microbenches.rs
```diff
@@ -364,6 +284,130 @@ fn attributes(c: &mut Criterion) {
...
+            criterion::black_box(unescape(text)).unwrap();
+        })
+    });
+
+    group.finish();
+}
+
+/// Benchmarks the `BytesText::unescaped()` method (includes time of `read_event`
+/// benchmark)
+fn bytes_text_unescaped(c: &mut Criterion) {
```
dralley, line 336: @Mingun Now that we have benchmarks for `unescape()` directly, this benchmark is a little redundant, do you think it still provides enough value to keep?

(it's not a new bench, I just relocated it within the file)

This benchmark (not a new one, it's just moved here from up above) is a little redundant now that there are benchmarks for `unescape()` directly, should we kee

src/events/attributes.rs
```diff
@@ -32,6 +35,93 @@ pub struct Attribute<'a> {
...
+                    let entity = &attr[idx + 1..end]; // starts after the &
+
+                    if let Some(s) = escapei::named_entity(entity) {
+                        normalized.extend_from_slice(s.as_bytes());
+                    } else if entity.starts_with(b"#") {
+                        let entity = &entity[1..]; // starts after the #
+                        let codepoint = escapei::parse_number(entity, idx..end)?;
+                        escapei::push_utf8(&mut normalized, codepoint);
+                    } else if let Some(value) = custom_entities.and_then(|hm| hm.get(entity)) {
+                        // TODO: recursively apply entity substitution
```
dralley, line 109: @Mingun Does the normal `unescape()` function need to take this into account also?

benches/microbenches.rs
```diff
@@ -125,86 +125,6 @@ fn read_namespaced_event(c: &mut Criterion) {
     group.finish();
 }
 
-/// Benchmarks the `BytesText::unescaped()` method (includes time of `read_event`
-/// benchmark)
-fn bytes_text_unescaped(c: &mut Criterion) {
```
dralley, line 130: @Mingun Now that we have benchmarks for unescape() directly, this benchmark is a little redundant, do you think it still provides enough value to keep?

(it's not a new bench, I just relocated it within the file)

Mingun: > this benchmark is a little redundant, 

Agree, can be removed

src/events/attributes.rs
```diff
@@ -32,6 +35,93 @@ pub struct Attribute<'a> {
...
+    ///
+    /// Do not use this method with HTML attributes.
+    ///
+    /// Escape sequences such as `&gt;` are replaced with their unescaped equivalents such as `>`
+    /// and the characters \t, \r, \n are replaced with whitespace characters.
+    ///
+    /// This will allocate unless the raw attribute value does not require normalization.
+    ///
+    /// See also [`normalized_value_with_custom_entities()`](#method.normalized_value_with_custom_entities)
+    pub fn normalized_value(&'a self) -> Result<Cow<'a, [u8]>, EscapeError> {
```
Mingun, line 50: I think, it will be valuable to add examples here. Actually, probably you can convert your new tests to doc examples and that will be enough

src/events/attributes.rs
```diff
@@ -4,13 +4,16 @@
...
 use crate::escape::{do_unescape, escape};
+use crate::escapei::{self, EscapeError};
 use crate::name::QName;
 use crate::reader::{is_whitespace, Reader};
 use crate::utils::{write_byte_string, write_cow_string, Bytes};
 use std::fmt::{self, Debug, Display, Formatter};
 use std::iter::FusedIterator;
 use std::{borrow::Cow, collections::HashMap, io::BufRead, ops::Range};
 
+type CustomEntityMap = HashMap<Vec<u8>, Vec<u8>>;
```
Mingun, line 15: I propose to step further and change the API, because I'm unpleasant with the current one:
- keys and values should be UTF-8 strings (as stated in one functions' comment), so it is logical to use `String` type instead of `Vec<u8>`
- this type forcing you to allocate even when you don't need that
- at the end it is better to replace all `*_with_custom_entities` methods with the `*_with(expand_entity: impl Fn(&str)->&str)` method

src/events/attributes.rs
Mingun: Yes, I think so

Mingun (commented):
> Can they? It looks like all these fields are private.

No, they can't. I confused with the ability to disable `with_checks` in the middle of an iteration.

> But, I feel like this is still the best option. There is already `attributes()` and `html_attributes()`, it's only a matter of changing the types.

Ok, then let's do that.

> The unfortunate thing will be, that the code between the two is _almost_ the same, just enough so that it will be really annoying to duplicate.

If you talk about `make_normalized_value`, you can convert it to a free function and use it in both API methods.

src/events/attributes.rs
Mingun: Actually, we should add a decoder parameter to that (and similar) functions and decode first before normalization.

Maybe also try another approach: introduce a new type of always UTF-8 encoded attributes and add a `Attribute::decode(&self) -> Result<Utf8Attribute>` function.

Instead of completely new type we could try to use `const bool` generic parameter (stable since 1.51, current MSRV is 1.41.1, from `memchr`)

src/events/attributes.rs
dralley: So basically something along these lines

```Rust
pub fn unescape(raw: &[u8]) -> Result<Cow<[u8]>, EscapeError> {
    do_unescape(raw, |_| None)
}

pub fn unescape_with<'a>(raw: &'a [u8], lookup_custom_entity: impl Fn(&str)->Option<&str>) {
     do_unescape(raw, lookup_custom_entity)
}

pub fn do_unescape<'a>(raw: &'a [u8], lookup_custom_entity: impl Fn(&str) -> Option<&str>) {
    // ....
    // ...
                let pat = &raw[start + 1..end];
                if pat.starts_with(b"#") {
                    let entity = &pat[1..]; // starts after the #
                    let codepoint = parse_number(entity, start..end)?;
                    push_utf8(unescaped, codepoint);
                } else if let Some(s) = named_entity(pat) {
                    unescaped.extend_from_slice(s.as_bytes());
                } else if let Some(value) = lookup_custom_entity(pat) {
                    unescaped.extend_from_slice(value.as_bytes());
                } else {
                    return Err(EscapeError::UnrecognizedSymbol(
                        start + 1..end,
                        String::from_utf8(pat.to_vec()),
                    ));
                }
     // ...
     // ...
}

fn main() {

   let custom_entities: HashMap<&'static str, &'static str> = HashMap::new();
   custom_entities.insert("anim", "lamb");

   let lookup_entity: |&str| -> Option<&str> = |ent| {
        custom_entities.get(ent)
   };

    println!("{}" unescape_with("mary had a little &anim;".as_bytes(), lookup_entity));
}

```

dralley: I'm going to split this off into a separate PR

dralley:
Apologies on the lack of forward progress, I haven't been able / willing to devote a lot of extra time to this (or the encoding work) in a while.

@Mingun The partially-but-not-completely recursive aspect of the algorithm in step 3b has ended up being a massive timesink, and I haven't yet gotten it to work.  For the sake of making some forward progress, would you be OK with an implementation that handles all of the parts of the algorithm except for the recursive resolution of entity references, and a TODO item for the recursive aspect?

I have no issues with making the other adjustments (such as adding examples), though I'd rather handle the structural API changes separately from this PR.

src/events/attributes.rs
```diff
@@ -791,6 +870,57 @@ mod xml {
     use super::*;
     use pretty_assertions::assert_eq;
 
+    #[test]
+    fn attribute_value_normalization() {
```
Mingun, line 874: It is better to have a `mod` with few test functions that tests different aspects of normalization. Then, if tests failed, we will known what is exactly wrong

src/events/attributes.rs
```diff
@@ -30,7 +31,85 @@ pub struct Attribute<'a> {
...
+            match ch {
+                b' ' | b'\n' | b'\r' | b'\t' => normalized.push(' '),
+                b'&' => {
+                    let end = idx
+                        + 1
+                        + attr_iter
+                            .position(|(_, c)| *c == b';')
+                            .ok_or_else(|| EscapeError::UnterminatedEntity(idx..attr.len()))?;
+                    let entity = &attr[idx + 1..end]; // starts after the &
+                    let entity_str = std::str::from_utf8(entity).expect("failed UTF-8 check");
```
Mingun, line 87: This will not work, `attr` in the document encoding

src/events/attributes.rs
```diff
@@ -30,7 +31,85 @@ pub struct Attribute<'a> {
...
 impl<'a> Attribute<'a> {
-    /// Decodes using UTF-8 then unescapes the value.
+    /// Returns the attribute value normalized as per the XML specification.
+    ///
+    /// https://www.w3.org/TR/xml/#AVNormalize
+    ///
+    /// Do not use this method with HTML attributes.
+    ///
+    /// Escape sequences such as `&gt;` are replaced with their unescaped equivalents such as `>`
+    /// and the characters \t, \r, \n are replaced with whitespace characters.
```
Mingun, line 41:
```suggestion
    /// and the characters `\t`, `\r`, `\n` are replaced with whitespace characters.
```
And the same in another function. Also, the wording in the documentation are outdated, there are no more `custom_entities`, `resolve_entity` is used instead, etc.

src/events/attributes.rs
```diff
@@ -30,7 +31,85 @@ pub struct Attribute<'a> {
...
+    ///
+    /// This will allocate unless the raw attribute value does not require normalization.
+    ///
+    /// See also [`normalized_value_with_custom_entities()`](#method.normalized_value_with_custom_entities)
```
Mingun, line 45:
```suggestion
    /// See also [`normalized_value_with()`](Self::normalized_value_with)
```

Mingun (changes requested):
As it implemented, it will not correctly work for non-UTF-8 documents. The support of recursive algorithm seems not so hard thing, I can look at it on weekend. I think, the key is a function like:
```rust
fn resolve(result: &mut String, value: &str, resolve_entity: impl Fn(...), ...) {
  for ch in value.iter() {
    match ch {
      '&' => {
        let entity = ...;
        let value = resolve_entity(entity);
        resolve(result, value, resolve_entity)
      }
      _ => result.push(ch),
    }
  }
}
```
Of course, it is needed to restrict recursion depth and somehow control cycles to prevent possibilities of DoS.

For your reference -- recently discovered an embryonic project for testing XML parser implementations to the standard's conformance: https://github.com/JOSEPHGILBY/xml-conformance-rs. Not looked at it yet, but seems encouraging.

dralley:
Well, it's still in draft for a reason :)  I'll clean it up before asking for a review

src/events/attributes.rs
dralley: The only way around that until the encoding PR is ready (which is blocked on async support atm, async unfortunately makes it much more challenging) is to change the signature to `decode_and_normalize_value` or disable it if the decoding feature is enabled.  Do you have a preference?

dralley:
>For your reference -- recently discovered an embryonic project for testing XML parser implementations to the standard's conformance: https://github.com/JOSEPHGILBY/xml-conformance-rs. Not looked at it yet, but seems encouraging.

Looks nice, it may even make sense to integrate the test code into our test suite.

benches/microbenches.rs
```diff
@@ -242,6 +242,50 @@ fn attributes(c: &mut Criterion) {
...
+    group.finish();
+}
+
+/// Benchmarks normalizing attribute values
+fn attribute_value_normalization(c: &mut Criterion) {
+    let mut group = c.benchmark_group("attribute_value_normalization");
+
+    group.bench_function("noop_short", |b| {
+        b.iter(|| {
+            criterion::black_box(unescape("foobar")).unwrap();
```
Mingun, line 255: `unescape`? It seems that those benchmarks does that that they are not supposed to do

src/escapei.rs
```diff
@@ -212,8 +212,66 @@ where
...
+                    // TODO: recursively apply entity substitution
+                    normalized.push_str(&value);
+                } else {
+                    return Err(EscapeError::UnrecognizedSymbol(
+                        idx + 1..end,
+                        String::from_utf8(entity.to_vec()).expect("failed UTF-8 check"),
+                    ));
+                }
+            }
+            _ => normalized.push(*ch as char),
```
Mingun, line 266: This is incorrect -- all multibyte characters (for example, Russian letters) would be transformed into several ASCII characters. I've implemented correct algorithm in my https://github.com/Mingun/quick-xml/tree/attr-val-normalization branch, based on your work. It is not finished yet, because it does not process built-in entities. I've implemented recursion and do not allocate if there are nothing to replace.

src/escapei.rs
```diff
@@ -212,8 +212,66 @@ where
...
+        }
+    }
+
+    Ok(Cow::Owned(normalized))
+}
+
 #[cfg(not(feature = "escape-html"))]
-fn named_entity(name: &str) -> Option<&str> {
+pub(crate) fn named_entity(name: &str) -> Option<&str> {
```
Mingun, line 274: Changes in visibility not required in all places. I've fixed that in my https://github.com/Mingun/quick-xml/tree/attr-val-normalization branch

src/events/attributes.rs
```diff
@@ -791,6 +887,94 @@ mod xml {
...
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        /// Empty values returned are unchanged
+        #[test]
+        fn test_empty() {
+            let raw_value = "".as_bytes();
+            let output = "".to_owned();
+            let attr = Attribute::from(("foo".as_bytes(), raw_value));
+            assert_eq!(attr.normalized_value(), Ok(Cow::Owned(output))); // TODO: this should not allocate
```
Mingun, line 902: A note: this does not check that value is owned, as was noted in https://github.com/tafia/quick-xml/pull/423#issuecomment-1407425485

Mingun (changes requested):
I've fixed the algorithm and other noted things in my branch, but I do not yet consider it as finished.

dralley:
I did a cursory review of the changes and it looks fine.  Would you prefer the commits squashed or kept separate?

I'll address everything else tomorrow.

Mingun:
I prefer to kept separate. It is somehow psychologically uncomfortable for review when one commit changes more than ~200 lines in each (or just several) files, even if half of them -- new tests. ¯\_(ツ)_/¯

I think, that at least separating normalization method to it's own commit would be a good idea. This is pretty isolated thing which, however, is big enough. That commit needed to be updated:
- Add tests for non-ASCII input
- Introduce new error kind and return it when `depth` become `0`. That means that we reach recursion limit
- (could be postponed) I would like to have limit configurable
- (could be postponed) Add a way to explicitly detect recursion (i.e. track the resolved entities and report which entity was defined recursively)
- (could be postponed) Add an ability to pass metainfo around entities. That way we can provide a way to report in error where the erroneous entity is defined, if resolver function provide that information

benches/microbenches.rs
```diff
@@ -242,6 +242,50 @@ fn attributes(c: &mut Criterion) {
...
+        })
+    });
+
+    group.bench_function("noop_long", |b| {
+        b.iter(|| {
+            criterion::black_box(unescape("just a bit of text without any entities")).unwrap();
```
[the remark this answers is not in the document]
dralley: I would hope that 1kb attribute values isn't a common thing.  But you're right that it should be a bit longer.

src/events/attributes.rs
```diff
@@ -796,6 +892,94 @@ mod xml {
...
+        #[test]
+        fn test_unrecognized_entity() {
+            let raw_value = "abc&unkn;def".as_bytes();
+            let attr = Attribute::from(("foo".as_bytes(), raw_value));
+            assert_eq!(
+                attr.normalized_value(),
+                Err(EscapeError::UnrecognizedSymbol(4..8, "unkn".to_owned()))
+                // TODO: UnrecognizedSymbol and UnterminatedEntity have divergent behavior
+                // regarding the start index of the range. Inappropriate?  Existing
+                // unescape code has this same issue.
```
dralley, line 949: @Mingun I can see a potential argument for having the start point of the range be different from each, but I'd just like to check with you, whether you think this is OK or an oversight to be fixed?

Mingun: I also noticed that and at first glance it seems that both errors should cover equivalent ranged. I'd tend to fix that, but I didn't dig deep.

dralley:
> Introduce new error kind and return it when depth become 0. That means that we reach recursion limit

Should this limit be configurable, and how would one configure it?  At some point it becomes unwieldy to pass around such state to each method call against `Attribute::normalize_value()` (thinking here also of the XML / HTML divergence in attribute handling).

Should we consider keeping the state in `Reader` and doing something along the lines of

* `Reader::normalize_attribute_value(attr)`, or
* `Attribute::normalize_value_with(resolve_entity, reader)` or  `Attribute::normalize_value_with(reader)` (moving the `resolve_entity` into `Reader` entirely)?

dralley:
@Mingun ^

Changelog.md
```diff
@@ -210,6 +215,10 @@ serde >= 1.0.181
 - [#565]: Fix compilation error when build with serde <1.0.139
 
 
+### New Tests
+
+- [#379]: Added tests for attribute value normalization
+
```
Mingun, line 221: Need be moved under the correct heading

Mingun (commented):
> How should one configure the limit? At some point it becomes unwieldy to keep all of this state external and provide it in each method call (we also have the XML / HTML divergence in attribute handling to consider).

Most naturally would be have a new option in `reader::Config`. That mean, we need somehow propagate it to the actual method. Some methods already takes `Reader`, so it will simple for them. Maybe we just need to start from only those methods and add shortcuts only when them explicitly will requested.

Actually, I already though about storing `Decoder` in the attribute itself (but because currently `Attribute` is a struct with public fields it will be a breaking change and I not very like the idea of making that new `decoder` field public, because it is implementation detail. Maybe in the end we will store already decoded data)

> Also I've noticed that some implementations detect an entity loop [immediately](https://gitlab.gnome.org/GNOME/libxml2/-/commit/a41b09c739db9cfc25e4c7172c3e1ebe5d9f6d82#26d7ed03847b376afab0a0c04b076525263956e0_2242_2247) instead of processing until the recursion limit is reached.

Yes, of course we should return error as soon as we found loop _or_ if recursion limit was exceeded.

> Should that be two separate errors in your opinion, or one error?

Two different. libxml2 also have two different errors, as you could notice from your link: one is "Detected an entity reference loop", other is "Maximum entity nesting depth exceeded"

dralley:
Obviously I haven't gotten around to finishing this

If you feel inclined to do so, I wouldn't mind if you picked it up.  If not, I'll get around to it eventually, I just haven't been doing a lot of coding in my free time.

src/events/attributes.rs
```diff
@@ -32,7 +34,175 @@ pub struct Attribute<'a> {
 }
 
 impl<'a> Attribute<'a> {
-    /// Decodes using UTF-8 then unescapes the value.
+    /// Returns the attribute value normalized as per [the XML specification].
+    ///
+    /// Do not use this method with HTML attributes.
```
Mingun, line 39: @dralley, you originally wrote this line of documentation. It seems to me, that we don't need two methods: one that performs normalization and one that doesn't. Why normalization methods should not be used for HTML attributes? Could you give a link for proof?

Mingun: Never mind. You already [answered](https://github.com/tafia/quick-xml/pull/379#issuecomment-1166179282) this question at the beginning of issue

Mingun: I think, this warning is useless and we can just change behavior of existing methods instead of adding new or more precisely deprecate them in favor of new names.

Actually, I doubt that we need HTML style of attributes, because we anyway cannot correcly parse HTML. We could work with XHTML because this is XML with semantics. But event in XHTML HTML-style attributes are not allowed. So in reality we hardly ever get an `Attribute` from HTML source and therefore will not need to get the value without normalization.

I think, in the future we should also remove `html` feature. All that it does just includes bunch of HTML entities which increases size of binaries and compile time. I do not want to support actuality of this list, and those consumers who need HTML-entities can use the corresponding `entity_resolver`.

Maybe we can keep ability to parse HTML-like attributes for XML-like-but-not-really-XML formats as long as it does not significantly affect parsing speed.

dralley:
Alright - I will close this one then and review whenever you get it posted.

Mingun:
Nothing, just that that branch is currently based on #941 (which a want to include in 0.39.2, just need some labour work to change return type from `Cow<str>` to `BytesText`), upcoming commit for 0.39.2 release, #938 and #914 in that order. I plan to finish #941 today or tomorrow, then merge, release 0.39.2, merge the last two PRs and publish that branch in new PR

PR #380, opened 2022-04-03
Fix errors in sequence deserialization

Mingun:
This PR adds a bunch of tests to ensure, that deserialization of sequences is correct in all situations.

This PR is based on #374 and #376, because they make a huge refactoring in files, that this PR touches.

For now all this works:

# 1. Interleaved sequence fields with non-sequence fields

XML
```xml
<any-tag>
  <item/>
  <item/>
  <before/>
  <item/>
  <after/>
  <item/>
</any-tag>
```
can be deserialized into
```rust
struct AnyName {
  before: (),
  item: Vec<()>,
  after: (),
}
```

# 2. List of choices

```xml
<any-tag>
  <one/>
  <two/>
  <three/>
</any-tag>
```
can be deserialized into
```rust
enum AnyNameChoice {
  one,
  two,
  #[serde(other)]
  other,
}
struct AnyName {
  #[serde(rename = "$value")]
  any_name: Vec<AnyNameChoice>,
}
```

# 3. Combination of above


Mingun:
Replaced by https://github.com/Mingun/fast-xml/pull/12

PR #381, opened 2022-04-10
Make `Deserializer::from_str` and `::from_slice` actually accessible

Mingun:
Previously, I've [added](https://github.com/tafia/quick-xml/commit/d683c377d9d9b75c2daabbb384030a30287ab624) this two methods to the `Deserializer` in #309, but because they was implemented for private type `SliceReader`, it was not actually possible to use that methods outside of crate.

Here a made types `SliceReader` and `IoReader` public, add a documentation for them and a new method `Deserializer::from_reader`. Method `Deserializer::from_bytes` was renamed to `Deserializer::from_slice` to follow convention that `serde-json`, `serde-yaml` and other follows.

Also free function `from_slice` now behaves the same as `from_bytes` (i.e. borrow from input instead of copying, like `from_reader`).

`from_bytes` was deprecated in preference to `from_slice`, because now those functions a fully equivalently.
`Deserializer::from_borrowing_reader` was deprecated in preference to `Deserializer::new`, because those methods a fully equivalently.

You may want to create a deserializer for a [transcoding](https://serde.rs/transcode.html) purposes.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,7 +14,7 @@
 - fix: allow to deserialize `unit`s from any data in attribute values and text nodes
 - refactor: unify errors when EOF encountered during serde deserialization
 - test: ensure that after deserializing all XML was consumed
-- feat: add `Deserializer::from_str` and `Deserializer::from_bytes`
+- feat: add `Deserializer::from_str`, `Deserializer::from_slice` and `Deserializer::from_reader`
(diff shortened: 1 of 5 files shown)


Mingun:
Merged in https://github.com/Mingun/fast-xml/compare/6855f43fe13c3d5e6985e1f38a359297f50380c1...deb6b450cf254dc098c40bbc4d97cdf8accd3f4d

PR #382, opened 2022-04-17
Add function Reader::underlying_reader_ref

ISibboI:
This is used to get an immutable reference to the reader underlying a `Reader`, without consuming the `Reader`.


Mingun:
This was added as `.get_ref()` / `.get_mut()` in 1d9af3780e73db7ad4d949482ba1ab289ad060b1

PR #387, opened 2022-05-21
Fix errors in sequence deserialization

### LINKED ISSUE #342: Deserialization fails when choice and element are in complex element
Serialization works, but deserialization not:

```rust
#[derive(Debug, serde::Serialize, serde::Deserialize)]
struct Root {
    element: Element,
    #[serde(rename = "$value")]
    choice: Choice,
}

#[derive(Debug, serde::Serialize, serde::Deserialize)]
struct Element(String);

#[derive(Debug, serde::Serialize, serde::Deserialize)]
enum Choice {
    ChoiceOne(String),
    ChoiceTwo(String),
}

fn main() {
    let root = Root {
        element: Element("element".to_string()),
        choice: Choice::ChoiceTwo("choice_two".to_string()),
    };

    let s = quick_xml::se::to_string(&root).unwrap();
    println!("{:?}", s);
    let r: Root = quick_xml::de::from_str(&s).unwrap();
    println!("{:?}", r);
}
```

Serialization:
```xml
<Root><Element>element</Element><ChoiceTwo>choice_two</ChoiceTwo></Root>
```

Deserialization error:
```
'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("unknown variant `Element`, expected `ChoiceOne` or `ChoiceTwo`")'
```

Mingun:
Replaces https://github.com/Mingun/fast-xml/pull/12

Fixes #342

This PR adds a bunch of tests to ensure, that deserialization of sequences is correct in all situations.

- Structs with `$value` field can now deserialize their other fields from elements, not only from attributes, as it was before:
    ```xml
    <any-tag>
      <before/>
      <item/>
      <item/>
      <item/>
      <after/>
    </any-tag>
    ```
    can be deserialized into
    ```rust
    struct AnyName {
      before: (),
      item: Vec<()>,
      after: (),
    }
    ```

- Added a new cargo feature `overlapped-lists` which allows also interleave sequence
  tags with non-sequence tags in the XML. When use that feature the following XML
  can be deserialized to the struct shown above:
    ```xml
    <any-tag>
      <item/>
      <item/>
      <before/>
      <item/>
      <after/>
      <item/>
    </any-tag>
    ```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -40,6 +40,10 @@ jobs:
       env:
         LLVM_PROFILE_FILE: coverage/serialize-escape-html-%p-%m.profraw
       run: cargo test --features serialize,escape-html
+    - name: Run tests (all features)
+      env:
+        LLVM_PROFILE_FILE: coverage/all-features-%p-%m.profraw
+      run: cargo test --all-features
     - name: Prepare coverage information for upload
       if: runner.os == 'Linux'
       # --token is required by grcov, but not required by coveralls.io, so pass
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,8 +78,14 @@ encoding = ["encoding_rs"]
 ## to quadratic parsing time, because the deserializer must check the list of
 ## events as many times as the number of sequence fields present in the schema.
 ##
+## To reduce negative consequences, always [limit] the maximum number of events
+## that [`Deserializer`] will buffer.
+##
 ## This feature works only with `serialize` feature and has no effect if `serialize`
 ## is not enabled.
+##
+## [limit]: crate::de::Deserializer::event_buffer_size
+## [`Deserializer`]: crate::de::Deserializer
 overlapped-lists = []
 
 ## Enables support for [`serde`] serialization and deserialization
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,46 @@ default = []
 ## [standard compliant]: https://www.w3.org/TR/xml11/#charencoding
 encoding = ["encoding_rs"]
 
+## This feature enables support for deserializing lists where tags are overlapped
+## with tags that do not correspond to the list.
+##
+## When this feature is enabled, the XML:
+## ```xml
+## <any-name>
+##   <item/>
+##   <another-item/>
+##   <item/>
+##   <item/>
+## </any-name>
+## ```
+## could be deserialized to a struct:
+## ```ignore
+## #[derive(Deserialize)]
+## #[serde(rename_all = "kebab-case")]
+## struct AnyName {
+##   item: Vec<()>,
+##   another_item: (),
+## }
+## ```
+##
+## When this feature is not enabled (default), only the first element will be
+## associated with the field, and the deserialized type will report an error
+## (duplicated field) when the deserializer encounters a second `<item/>`.
+##
+## Note, that enabling this feature can lead to high and even unlimited memory
+## consumption, because deserializer should check all events up to the end of a
+## container tag (`</any-name>` in that example) to figure out that there are no
+## more items for a field. If `</any-name>` or even EOF is not encountered, the
+## parsing will never end which can lead to a denial-of-service (DoS) scenario.
+##
+## Having several lists and overlapped elements for them in XML could also lead
+## to quadratic parsing time, because the deserializer must check the list of
+## events as many times as the number of sequence fields present in the schema.
+##
+## This feature works only with `serialize` feature and has no effect if `serialize`
+## is not enabled.
+overlapped-lists = []
+
 ## Enables support for [`serde`] serialization and deserialization
 serialize = ["serde"]
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,9 @@
 - [#9]: Deserialization erroneously was successful in some cases where error is expected.
   This broke deserialization of untagged enums which rely on error if variant cannot be parsed
 - [#387]: Allow to have an ordinary elements together with a `$value` field
+- [#387]: Internal deserializer state can be broken when deserializing a map with
+  a sequence field (such as `Vec<T>`), where elements of this sequence contains
+  another sequence. This error affects only users with the `serialize` feature enabled
 
 ### Misc Changes
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -36,9 +36,11 @@
 ### New Tests
 
 - [#9]: Added tests for incorrect nested tags in input
+- [#387]: Added a bunch of tests for sequences deserialization
 
 [#8]: https://github.com/Mingun/fast-xml/pull/8
 [#9]: https://github.com/Mingun/fast-xml/pull/9
+[#387]: https://github.com/tafia/quick-xml/pull/387
 [#391]: https://github.com/tafia/quick-xml/pull/391
 
 ## 0.23.0 -- 2022-05-08
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,11 @@
 
 ## Unreleased
 
+### New Features
+
+- [#387]: Allow overlapping between elements of sequence and other elements
+  (using new feature `overlapped-lists`)
+
 ### Bug Fixes
 
 - [#9]: Deserialization erroneously was successful in some cases where error is expected.
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@
 
 - [#9]: Deserialization erroneously was successful in some cases where error is expected.
   This broke deserialization of untagged enums which rely on error if variant cannot be parsed
+- [#387]: Allow to have an ordinary elements together with a `$value` field
 
 ### Misc Changes
 
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -105,6 +105,9 @@ enum ValueSource {
     /// [list of known fields]: MapAccess::fields
     Content,
     /// Next value should be deserialized from an element with a dedicated name.
+    /// If deserialized type is a sequence, then that sequence will collect all
+    /// elements with the same name until it will be filled. If not all elements
+    /// would be consumed, the rest will be ignored.
     ///
     /// That state is set when call to [`peek()`] returns a [`Start`] event, which
     /// [`name()`] represents a field name. That name will be deserialized as a key.
@@ -569,7 +572,11 @@ where
     map: &'m mut MapAccess<'de, 'a, R>,
     /// Filter that determines whether a tag is a part of this sequence.
     ///
-    /// Iteration will stop when found a tag that does not pass this filter.
+    /// When feature `overlapped-lists` is not activated, iteration will stop
+    /// when found a tag that does not pass this filter.
+    ///
+    /// When feature `overlapped-lists` is activated, all tags, that not pass
+    /// this check, will be skipped.
     filter: TagFilter<'de>,
 }
 
@@ -584,20 +591,29 @@ where
         T: DeserializeSeed<'de>,
     {
         let decoder = self.map.de.reader.decoder();
-        match self.map.de.peek()? {
-            // Stop iteration when list elements ends
-            DeEvent::Start(e) if !self.filter.is_suitable(&e, decoder)? => Ok(None),
+        loop {
+            break match self.map.de.peek()? {
+                // If we see a tag that we not interested, skip it
+                #[cfg(feature = "overlapped-lists")]
+                DeEvent::Start(e) if !self.filter.is_suitable(&e, decoder)? => {
+                    self.map.de.skip()?;
+                    continue;
+                }
+                // Stop iteration when list elements ends
+                #[cfg(not(feature = "overlapped-lists"))]
+                DeEvent::Start(e) if !self.filter.is_suitable(&e, decoder)? => Ok(None),
 
-            // Stop iteration after reaching a closing tag
-            DeEvent::End(e) if e.name() == self.map.start.name() => Ok(None),
-            // This is a unmatched closing tag, so the XML is invalid
-            DeEvent::End(e) => Err(DeError::UnexpectedEnd(e.name().to_owned())),
-            // We cannot get `Eof` legally, because we always inside of the
-            // opened tag `self.map.start`
-            DeEvent::Eof => Err(DeError::UnexpectedEof),
+                // Stop iteration after reaching a closing tag
+                DeEvent::End(e) if e.name() == self.map.start.name() => Ok(None),
+                // This is a unmatched closing tag, so the XML is invalid
+                DeEvent::End(e) => Err(DeError::UnexpectedEnd(e.name().to_owned())),
+                // We cannot get `Eof` legally, because we always inside of the
+                // opened tag `self.map.start`
+                DeEvent::Eof => Err(DeError::UnexpectedEof),
 
-            // Start(tag), Text, CData
-            _ => seed.deserialize(&mut *self.map.de).map(Some),
+                // Start(tag), Text, CData
+                _ => seed.deserialize(&mut *self.map.de).map(Some),
+            };
         }
     }
 }
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -2,14 +2,14 @@
 
 use crate::{
     de::escape::EscapedDeserializer,
-    de::seq::not_in,
+    de::seq::{not_in, TagFilter},
     de::{deserialize_bool, DeEvent, Deserializer, XmlRead, INNER_VALUE, UNFLATTEN_PREFIX},
     errors::serialize::DeError,
     events::attributes::IterState,
     events::{BytesCData, BytesStart},
     reader::Decoder,
 };
-use serde::de::{self, DeserializeSeed, IntoDeserializer, Visitor};
+use serde::de::{self, DeserializeSeed, IntoDeserializer, SeqAccess, Visitor};
 use serde::serde_if_integer128;
 use std::borrow::Cow;
 use std::ops::Range;
@@ -312,7 +312,10 @@ where
             // is implicit and equals to the `INNER_VALUE` constant, and the value
             // is a `Text` or a `CData` event (the value deserializer will see one
             // of that events)
-            ValueSource::Text => seed.deserialize(MapValueDeserializer { map: self }),
+            ValueSource::Text => seed.deserialize(MapValueDeserializer {
+                map: self,
+                allow_start: false,
+            }),
             // This arm processes the following XML shape:
             // <any-tag>
             //   <any>...</any>
@@ -320,7 +323,10 @@ where
             // The whole map represented by an `<any-tag>` element, the map key
             // is implicit and equals to the `INNER_VALUE` constant, and the value
             // is a `Start` event (the value deserializer will see that event)
-            ValueSource::Content => seed.deserialize(MapValueDeserializer { map: self }),
+            ValueSource::Content => seed.deserialize(MapValueDeserializer {
+                map: self,
+                allow_start: false,
+            }),
             // This arm processes the following XML shape:
             // <any-tag>
             //   <tag>...</tag>
@@ -328,7 +334,10 @@ where
             // The whole map represented by an `<any-tag>` element, the map key
             // is a `tag`, and the value is a `Start` event (the value deserializer
             // will see that event)
-            ValueSource::Nested => seed.deserialize(&mut *self.de),
+            ValueSource::Nested => seed.deserialize(MapValueDeserializer {
+                map: self,
+                allow_start: true,
+            }),
             ValueSource::Unknown => Err(DeError::KeyNotRead),
         }
     }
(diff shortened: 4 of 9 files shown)


tests/serde-de.rs
```diff
@@ -512,6 +512,2288 @@ mod tuple_struct {
...
+    use super::*;
+
+    /// Check that top-level sequences can be deserialized from the multi-root XML documents
+    mod top_level {
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        #[test]
+        fn simple() {
+            from_str::<[(); 3]>("<root/><root>42</root><root>answer</root>").unwrap();
```
dralley, line 525: Why not make this consistent with the Vec test?

tests/serde-de.rs
```diff
@@ -512,6 +512,2530 @@ mod tuple_struct {
...
```
dralley, line 525: This will panic if it cannot be deserialized to an array of size 3 yes?  (I am not as familiar with serde, my code doesn't use it)

Just curious why it's not asserted like the Vec scenario, I'm assuming this is why.

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
...
+## ```ignore
+## #[derive(Deserialize)]
+## #[serde(rename_all = "kebab-case")]
+## struct AnyName {
+##   item: Vec<()>,
+##   another_item: (),
+## }
+## ```
+##
+## When feature is not enabled (default), only first element will be assotiated
```
dralley, line 67:
```suggestion
## When this feature is not enabled (default), only the first element will be associated
```

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
 ## [standard compliant]: https://www.w3.org/TR/xml11/#charencoding
 encoding = ["encoding_rs"]
 
+## This feature enables support of deserializing of lists which tags are overlapped
```
dralley, line 45:
```suggestion
## This feature enables support for deserializing lists where tags are overlapped
```

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
...
+## with tags that does not correspond to the list.
```
dralley, line 46:
```suggestion
## with tags that do not correspond to the list.
```

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
...
+##
+## When this feature is enabled, that XML:
```
dralley, line 48:
```suggestion
## When this feature is enabled, the XML:
```

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
...
+## #[derive(Deserialize)]
+## #[serde(rename_all = "kebab-case")]
+## struct AnyName {
+##   item: Vec<()>,
+##   another_item: (),
+## }
+## ```
+##
+## When feature is not enabled (default), only first element will be assotiated
+## with a field, and deserializer will report an error when it encounter a second
```
dralley, line 68:
```suggestion
## with the field, and the deserializer will report an error when it encounter a second
```

Cargo.toml
dralley, line 48:
```suggestion
## When this feature is enabled, that XML:
```

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
...
+## `<item/>`.
+##
+## Note, that enabling this feature can lead to high and even unlimited memory
+## consumption, because deserializer should check all events up to the end of a
+## container tag (`</any-name>` in that example) to figure out that there are no
+## more items for a field. If `</any-name>` or EOF even not encountered, the
```
dralley, line 74:
```suggestion
## more items for a field. If `</any-name>` or even EOF is not encountered, the
```

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
...
+## parsing will never end which can lead to DoS.
```
dralley, line 75:
```suggestion
## parsing will never end which can lead to a denial-of-service (DoS) scenario.
```

dralley: I suppose that anyone with control over the XML could accomplish this regardless by just endlessly stuffing more items into the list, but this would be more difficult to notice.

Cargo.toml
```diff
@@ -42,6 +42,46 @@ default = []
...
+##
+## Having several lists and overlapped elements for them in XML could also lead
+## to quadratic parsing time, because deserialzier have to check list of events
+## as many times as count of sequence fields present.
```
dralley, line 79:
```
## Having several lists and overlapped elements for them in XML could also lead
## to quadratic parsing time, because the deserializer must check the list of events
## as many times as there are sequence fields present in the schema.
```

dralley:
This is a long PR so it might take me a few days to get through it.  Handling the documentation first since it gives me a better picture of what the code ought to look like.

src/de/mod.rs
```diff
@@ -277,6 +279,11 @@ where
     /// moved from this to [`Self::read`].
     #[cfg(feature = "overlapped-lists")]
     write: VecDeque<DeEvent<'de>>,
+    /// Maximum number of events that could be skipped. This field is used to prevent
```
dralley, line 282:
```suggestion
/// Maximum number of events that can be skipped when processing sequences that occur out-of-order.
```

tests/serde-de.rs
dralley: It's fine

Cargo.toml
```diff
@@ -78,8 +78,14 @@ encoding = ["encoding_rs"]
 ## to quadratic parsing time, because deserialzier have to check list of events
 ## as many times as count of sequence fields present.
 ##
+## To reduce negative consequences always [limit] maximum number of events that
```
dralley, line 81:
```suggestion
## To reduce negative consequences, always [limit] the maximum number of events that
```

src/errors.rs
```diff
@@ -159,6 +161,10 @@ pub mod serialize {
         ExpectedStart,
         /// Unsupported operation
         Unsupported(&'static str),
+        /// Too many events was skipped during deserializing a sequence, limit of
```
dralley, line 164:
```suggestion
        /// Too many events were skipped while deserializing a sequence, event limit
```

src/errors.rs
```diff
@@ -159,6 +161,10 @@ pub mod serialize {
...
+        /// events exceeded. The limit provides as an argument
```
dralley, line 165:
```suggestion
        /// exceeded. The limit was provided as an argument
```

src/de/mod.rs
```diff
@@ -445,24 +520,36 @@ where
...
+                        _ => self.skip_event(event)?,
                     }
                 }
             }
             _ => Ok(()),
         }
     }
 
+    #[cfg(feature = "overlapped-lists")]
+    #[inline]
```
dralley, line 542: This is short enough that, since it's only used in the function above, it could be defined as a closure there.

No strong feelings about it.

src/de/mod.rs
```diff
@@ -387,6 +396,72 @@ where
...
 
+    /// Set the maximum number of events that could be skipped during deserialization
+    /// of sequences.
+    ///
+    /// If `<element>` contains more than specified nested elements, `#text` or
+    /// CDATA nodes, then [`DeError::TooManyEvents`] will be returned during
+    /// deserialization of sequence field (any type that uses [`deserialize_seq`]
+    /// for the deserialization, for example, `Vec<T>`).
+    ///
+    /// This method can be used to prevent a [DoS] attack and infinity memory
```
dralley, line 407:
```suggestion
    /// This method can be used to prevent a [DoS] attack and infinite memory
```

or "unbounded"

src/de/mod.rs
```diff
@@ -387,6 +396,72 @@ where
...
+    /// consumption when parse a very large or even an infinity XML to a sequence
```
dralley, line 408:
```suggestion
    /// consumption when parsing a very large or even an infinite XML to a sequence
```

Could you explain "infinite xml"?  (to me, not in the text)

src/de/mod.rs
```diff
@@ -387,6 +396,72 @@ where
...
+    /// field.
+    ///
+    /// It is strongly recommended to set limit to some value when you parse data
+    /// from untrusted source. You should choose a value that you typical XMLs
```
dralley, line 412: "untrusted sources" or "an untrusted source"

dralley: your typical XML

src/de/mod.rs
```diff
@@ -387,6 +396,72 @@ where
...
+    /// The XML that we try to parse look like this:
+    /// ```xml
+    /// <any-name>
+    ///   <item/>
+    ///   <!-- Bufferisation starts at this point -->
+    ///   <another-item>
+    ///     <some-element>with text</some-element>
+    ///     <yet-another-element/>
+    ///   </another-item>
+    ///   <!-- Buffer will be emptied at this point; was buffered 7 events -->
```
dralley, line 433:
```suggestion
    ///   <!-- Buffer will be emptied at this point; 7 events will be drained -->
```

src/de/mod.rs
```diff
@@ -387,6 +396,72 @@ where
...
+    /// struct List {
+    ///   item: Vec<()>,
+    /// }
+    /// ```
+    ///
+    /// The XML that we try to parse look like this:
+    /// ```xml
+    /// <any-name>
+    ///   <item/>
+    ///   <!-- Bufferisation starts at this point -->
```
dralley, line 428:
```suggestion
    ///   <!-- Bufferization starts at this point -->
```

src/de/mod.rs
Mingun: I mean, that someone could try to parse an XML stream what will never end, for example, some live event log:
```xml
<EventList>
  <Event/>
  <Event/>
  <Event/>
  ...
```
But without deserializing such stream in a streaming way it anyway impossible to get full list regardless of this option. I'll remove that sentence

src/de/mod.rs
Mingun: It seems that suggested change will change an attention point slightly. I try to emphasize, that buffer must have size enough to store at least 7 events (which reader may unexpect, because from his point of view it may seem that only one element need to be skipped and calculations for possible buffer size can be totally wrong if follow that logic. The real size should be at least 7 times bigger), and after that change is seems that sentence mostly gives information that you get exactly 7 events. Maybe I'm wrong and that change does not change the sence... please tell me your opinion.

Mingun:
@dralley , I've address all your suggestions except one (I have a question about it). Could you finish your review this week? I have a tons of plans and unfortunately they all are a strictly coupled with each other.

dralley:
@Mingun Yeah, I'll take another look tonight

src/de/mod.rs
dralley: You could reword it to "7 events were buffered" in that case

src/de/seq.rs
```diff
@@ -1,18 +1,74 @@
...
+use crate::reader::Decoder;
 use serde::de::{self, DeserializeSeed};
+#[cfg(not(feature = "encoding"))]
+use std::borrow::Cow;
 
+/// Check if tag `start` is included in the `fields` list. `decoder` is used to
+/// get a string representation of a tag.
+///
+/// Returns `true`, if `start` is not in the `fields` list and `false` otherwise.
+pub fn is_unknown(
```
dralley, line 12: Could this be named more descriptively e.g. `is_unknown_{field|tag|element}`?

src/de/map.rs
```diff
@@ -135,7 +140,24 @@ enum ValueSource {
...
+/// ```
+///
+/// Name of this tag is stored in a [`Self::start`] property.
+///
+/// # Lifetimes
+///
+/// - `'de` lifetime represents a buffer, from which deserialized values can
+///   borrow their data. Depending on the underlying reader, there can be an
+///   internal buffer of deserializer (i.e. deserializer itself) or an input
+///   (in that case it is possible to approach zero-cost deserialization).
```
dralley, line 160: Zero-cost, or zero-copy?  The latter is what is specifically being achieved here while the former is kind of a fuzzy statement.  I take it to mean that "deserialization is fast enough that the CPU can keep up with IO" in the same vein as e.g. `simdjson`

src/de/seq.rs
```diff
@@ -1,18 +1,74 @@
...
+
+    #[cfg(feature = "encoding")]
+    let tag = decoder.decode(start.name());
+
+    Ok(fields.iter().all(|&field| field != tag.as_ref()))
+}
+
+/// A filter that determines, what tags should form a sequence.
+///
+/// There is a two variant of sequences:
```
dralley, line 28: /// There are two variants of sequences:

src/de/map.rs
```diff
@@ -43,12 +44,14 @@ enum ValueSource {
     /// </any-tag>
     /// ```
     Text,
-    /// Next value should be deserialized from an element with an any name.
-    /// Corresponding tag name will always be associated with a field with name
-    /// [`INNER_VALUE`].
+    /// Next value should be deserialized from an element with an any name, except
+    /// elements which name matched one of the struct fields. Corresponding tag
```
dralley, line 48:
```suggestion
    /// elements with a name matching one of the struct fields. Corresponding tag
```

src/de/seq.rs
dralley: "unknown" also sounds a little strange in this context since we theoretically do known the names of the enum variants?

re:

> 
>     enum Choice { one, two, three }
>     struct Root {
>         #[serde(rename = "$value")]
>         item: [Choice; 3],
>         node: (),
>     }
> 
> and this struct should be able to deserialized from
> 
>     <root>
>         <one/>
>         <two/>
>         <three/>
>         <node/>
>     </root>

Maybe `tag_maps_to_field`?  (That name sounds a bit wrong as well, so if you prefer the former names I'm fine with that)

src/de/seq.rs
```diff
@@ -1,18 +1,74 @@
...
+/// - sequence where each element represented by tags with the same name
+/// - sequence where each element can have a different tag
+///
+/// The first variant could represent a collection of structs, the second --
+/// a collection of enums.
```
dralley, line 33:
```suggestion
/// a collection of enum variants
```

src/de/seq.rs
dralley: To deconflict with "enum variants" perhaps "types of sequences" would be better here.

src/de/seq.rs
```diff
@@ -1,18 +1,74 @@
...
+///
+/// In the second case we don't know what name sequence element will accept,
+/// so we take an any element. But because in XML sequences a flattened into a
+/// maps, then we could take an elements that have their own dedicated fields
+/// in a struct. To prevent this we use an `Exclude` filter, that filters out
+/// any known names of a struct fields.
```
dralley, line 39:
```suggestion
/// tags with any known field names from the struct.
```

src/de/seq.rs
```diff
@@ -22,8 +78,10 @@ pub struct SeqAccess<'de, 'a, R>
 where
     R: XmlRead<'de>,
 {
+    /// Deserializer used to deserialize sequence items
     de: &'a mut Deserializer<'de, R>,
-    names: Names,
+    /// Filter that determines is that tag is a part of this sequence?
```
dralley, line 83:
```suggestion
    /// Filter that determines whether a tag is a part of this sequence
```

Changelog.md
```diff
@@ -15,6 +15,9 @@
 - [#9]: Deserialization erroneously was successful in some cases where error is expected.
   This broke deserialization of untagged enums which rely on error if variant cannot be parsed
 - [#387]: Allow to have an ordinary elements together with a `$value` field
+- [#387]: Internal deserializer state can be broken when deserialize a map with
```
dralley, line 23:
```suggestion
- [#387]: Internal deserializer state can be broken when deserializing a map with
```

Changelog.md
```diff
@@ -15,6 +15,9 @@
...
+  a sequence field (such as `Vec<T>`), where elements of this sequence contains
+  another sequence. This error affects only users with a `serialize` feature enabled
```
dralley, line 25:
```suggestion
  another sequence. This error affects only users with the `serialize` feature enabled
```

src/de/map.rs
```diff
@@ -173,6 +173,14 @@ where
     source: ValueSource,
     /// List of field names of the struct. It is empty for maps
     fields: &'static [&'static str],
+    /// If `true`, then deserialized struct have a field with a special name
```
dralley, line 179:
```suggestion
    /// If `true`, then the deserialized struct has a field with a special name:
```

src/de/seq.rs
Mingun: I chose `is_unknown` simple in contrary to `is_known` (i.e. if tag is listed in `fields` it is known for a map), but because each usage of `is_known()` would be `!is_known()` I inverted it. The method does not have any specific cognitive load, so just `not_in` should be enough.

src/de/map.rs
Mingun: I mean that when optimized any serde deserializer code should gone as if you manually locate positions in the string and take a slice. But `zero-copy` is a better anyway.

src/de/seq.rs
Mingun: Agreed

src/de/seq.rs
Mingun: Then

> /// In the second case we don't know what tag name should be expected as a
> /// sequence element, so we accept any element. Since the sequence are flattened
> /// into maps, we _skip_ elements which have dedicated fields in a struct by using an 
> /// `Exclude` filter that filters out elements with names matching field names from the struct.

This describes that fact, that struct
```rust
struct AnyName {
  field: (),
  list: Vec<()>,
}
```
in XML would be represented as
```xml
<any-tag>
  <field/>
  <list/>
  <list/>
<any-tag>
```
The JSON equivalent would be
```json
{
  "field": null,
  "list": null,
  "list": null
}
```
instead of correct
```json
{
  "field": null,
  "list": [null, null]
}
```
Because collection element could be an `enum` and enums in XML could be represented as
```xml
<variant-name .../>
```
and because the type of a sequence element is opacue to us at the `AnyName` deserialization level, we should consider all tags as a potential list items with one exception: if we are know with guarantee that a tag with that name is mapped to another field. We list all names of that tags in the `Exclude` variant and this list is a `fields` list of a struct.
In order to achieve correct deserialization we should collect all tags into a sequence

src/de/map.rs
```diff
@@ -364,6 +373,82 @@ where
     /// Access to the map that created this deserializer. Gives access to the
     /// context, such as list of fields, that current map known about.
     map: &'m mut MapAccess<'de, 'a, R>,
+    /// Determines, should [`Deserializer::next_text_impl()`] expand the second
+    /// level of tags or not.
+    ///
+    /// If this field is `true`, we processes the following XML shape:
```
dralley, line 390:
```suggestion
    /// If this field is `true`, we process the following XML shape:
```

src/de/seq.rs
dralley: Makes sense.  I always thought the most natural XML representation was akin to the JSON one, e.g.

```xml
<any-tag>
  <field/>
  <list>
    <item/>
    <item/>
    <item/>
  </list>
<any-tag>
```

But I can see some rationale for not doing it that way, too.

Mingun: Yes, it would be nice for this to be the case, but I often see such XML. If you think about it, this is natural, given how easy it is to describe such a structure in XSD:
```xml
<xs:complexType>
  <xs:sequence>
    <xs:element name="field"/>
    <xs:element name="list" maxOccurs="unbounded"/>
  </xs:sequence>
</xs:complexType>
```

.github/workflows/rust.yml
```diff
@@ -40,6 +40,10 @@ jobs:
       env:
         LLVM_PROFILE_FILE: coverage/serialize-escape-html-%p-%m.profraw
       run: cargo test --features serialize,escape-html
+    - name: Run tests (all features)
+      env:
+        LLVM_PROFILE_FILE: coverage/all-features-%p-%m.profraw
+      run: cargo test --all-features
```
dralley, line 46: Looks like this should cover testing both with and without `overlapped-lists` which is a concern because of so much alternate code :+1:

src/de/mod.rs
```diff
@@ -370,12 +416,69 @@ where
...
+    /// read and write buffers.
+    ///
+    /// After calling this method, [`Self::peek()`] and [`Self::next()`] starts
+    /// return events that was skipped previously by calling [`Self::skip()`],
+    /// and only when all that events will be consumed, the deserializer starts
+    /// to drain events from underlying reader.
+    ///
+    /// This method MUST be called if any number of [`Self::skip()`] was called
+    /// after [`Self::new()`] or `start_replay()` or you'll lost events.
+    #[cfg(feature = "overlapped-lists")]
```
dralley, line 563: Just curious about the use case, is this basically for situations where you're parsing messy handwritten XML / HTML?  Generally I would think machine-generated XML wouldn't have this problem.

Cargo.toml
```diff
@@ -42,6 +42,52 @@ default = []
 ## [standard compliant]: https://www.w3.org/TR/xml11/#charencoding
 encoding = ["encoding_rs"]
 
+## This feature enables support for deserializing lists where tags are overlapped
+## with tags that do not correspond to the list.
```
dralley, line 46: Just curious about the use case, is this basically for situations where you're parsing messy handwritten XML / HTML? Generally I would think machine-generated XML wouldn't have this problem.

It might be worth providing that as an example of a situation where you might want to enable it if that is the case.

.github/workflows/rust.yml
Mingun: Yes, that is why I add this.

dralley (approved):

Cargo.toml
Mingun: Actually, I cannot remember that would I worked with such XMLs, but all parsers, that I know, very tolerant to overlapped lists (even in `xs:sequence`, although XSD requires exact order of tags. I never seen using `xs:all` that allows arbitrary order). So I thought there should be an option to allow this behavior

k-bx: @dralley I'm trying to parse KML file which can have elements of types Folder, Document and Point interleaved in a single list. I'd like to ideally get them in the same order, so I tried to create an enum covering all three cases, but it didn't work. So this at least gives me an option to parse it as three lists without the preservation of order in-between the items.

k-bx: @Mingun it works, but it isn't really what's needed here. What is the desired way to handle things:

input: `<TagOne></TagOne><TagTwo></TagTwo><TagOne></TagOne>`

parsed as:

```
struct TagOne {};
struct TagTwo {};
enum Tag {
  TagOne(TagOne),
  TagTwo(TagTwo)
}

-- result should be
vec![Tag::TagOne(TagOne {}), Tag::TagTwo(TagTwo {}), Tag::TagOne(TagOne {})]
```

I didn't see a way to do something like this via the current lib ^

Do I need to create a separate issue to make something like this possible?

Mingun: Yes, this is an error, I filled #500 for it

PR #389, opened 2022-05-25
Add method to allow treating start tags as empty

andersjanmyr:
Allows parsing the HTML in the README.
```html
<meta charset="utf-8">
<link rel="manifest" href="/manifest.webmanifest">
```

```rust
let mut reader = Reader::from_str(xml);
reader.treat_tag_as_empty("link meta");
```


andersjanmyr:
@Mingun Yeah, that's what I'm doing now. I just figured it would be nice to have the option to do this, but I understand if you don't want it. This is an XML parser after all. :)

PR #393, opened 2022-06-05
Introduce typified wrappers for names and fix couple of bugs

Mingun:
Fixed bugs:
- not declared prefix in the tags resulted in a namespace resolution, as if the tag were not bound to any namespace:
  ```xml
  <unknown:element/>
  ```
  was considered as
  ```xml
  <element/>
  ```
- elements/attributes with [reserved](https://www.w3.org/TR/xml-names11/#xmlReserved) names (started with `"xml"i`) erroneously was treated as error, although XML spec explicitly requires that fatal error MUST NOT be raised. Because quick-xml does not check names validity, now such names does not produce any error.

To fix this kind of errors and avoid messing up different names, a typified wrappers for names was introduced.

This PR is the first step to properly support namespaces in quick-xml.


src/reader.rs
```diff
@@ -626,18 +638,18 @@ impl<R: BufRead> Reader<R> {
                 self.ns_resolver.push(&e, namespace_buffer);
```
dralley, line 655: The change makes sense, but just checking - the user could also avoid this whole scenario by enabling the expansion of Empty tags into Start and End tags, right?  This is a workaround for this one specific scenario when that feature isn't enabled.

src/reader.rs
```diff
@@ -102,8 +102,14 @@ pub struct Reader<R: BufRead> {
     /// Opened name start indexes into [`Self::opened_buffer`]. See documentation
     /// for that field for details
     opened_starts: Vec<usize>,
-    /// a buffer to manage namespaces
+
+    /// A buffer to manage namespaces
     ns_resolver: NamespaceResolver,
+    /// For `Empty` events keep the 'scope' of the namespace on the stack artificially. That way, the
+    /// consumer has a chance to use `resolve` in the context of the empty element. We perform the
+    /// pop as the first operation in the next `next()` call.
```
dralley, line 110: The change makes sense generally, but just checking - the user could also avoid this whole scenario by enabling the expansion of Empty tags into Start and End tags, right? This is a workaround for this one specific scenario when that feature isn't enabled.

dralley: I'm reading section 6.1 of https://www.w3.org/TR/xml-names11 which says:

>  The scope of a namespace declaration declaring a prefix extends from the beginning of the start-tag in which it appears to the end of the corresponding end-tag, excluding the scope of any inner declarations with the same NSAttName part. In the case of an empty tag, the scope is the tag itself. 

The way this reads to me is that providing this special case is a minor violation of the spec - by making the namespace of an empty tag available in this way, we are lying about the namespace scope.  

(Maybe you address that in later commits, I'm working my way through one by one)

tests/namespaces.rs
```diff
@@ -349,3 +349,29 @@ fn default_ns_shadowing_expanded() {
         e => panic!("Expected End event (</outer>), got {:?}", e),
     }
 }
+
+/// Although XML specification [do not recommend] use names which local name started
```
dralley, line 353: Not sure what you're saying here, do you mean something like 

> Although the XML specification does not recommend the use of names which start with letters "xml"

src/name.rs
```diff
@@ -380,3 +380,237 @@ impl NamespaceResolver {
...
+                (Some(ns), LocalName(b"simple"))
+            );
+            assert_eq!(
+                resolver.resolve(name, &buffer, false),
+                (None, LocalName(b"simple"))
+            );
+            assert_eq!(resolver.find(name, &buffer), Some(ns));
+        }
+
+        /// That test add a second level of namespaces, which replaces the previous binding
```
dralley, line 595:
```suggestion
        /// Test adding a second level of namespaces, which replaces the previous binding
```

src/name.rs
```diff
@@ -380,3 +380,237 @@ impl NamespaceResolver {
...
+        #[test]
+        fn override_() {
```
dralley, line 597:
```suggestion
        fn override_namespace() {
```

src/reader.rs
Mingun: These lines was just moved from `NamespaceResolver`. This code does the right thing, because otherwise you will pop stored namespace before the user would able to use it:
```xml
<tag xmlns="..."/>
<!-- ^^^^^^^^^^^    - namespace introduced and pushed to the stack -->
<!--            ^^  - namespace popped from the stack -->
<!--              ^ - Event::Empty returned to the user, namespace is not availible -->
```
Compare to the events happening for the open-close pair:
```xml
<tag xmlns="...">
<!-- ^^^^^^^^^^^   - namespace introduced and pushed to the stack -->
<!--             ^ - Event::Start returned to the user, namespace is availible -->
</tag>
<!--^^             - namespace popped from the stack -->
<!--  ^            - Event::End returned to the user -->
```
Although `Event::End` is returned after its namespace was popped, `read_namespaced_event` returns that namespace as well, and you does not have attributes in the `End` tag where you could use the popped namespace.

When namespace popping is deferred, the situation is:
```xml
<tag xmlns="..."/>
<!-- ^^^^^^^^^^^    - namespace introduced and pushed to the stack -->
<!--              ^ - Event::Empty returned to the user, namespace is availible -->
<!--                - namespace popped from the stack on next event -->
```

That extending does not violates anything, because the user could access to that namespace only by calling methods on the returned `Event::Empty`, on the next call to `reader.next()` that namespace will be popped before returning the next event.

> the user could also avoid this whole scenario by enabling the expansion of Empty tags into Start and End tags, right?

Right.

src/name.rs
```diff
@@ -380,3 +380,237 @@ impl NamespaceResolver {
...
+                (Some(old_ns), LocalName(b"simple"))
+            );
+            assert_eq!(
+                resolver.resolve(name, &buffer, false),
+                (None, LocalName(b"simple"))
+            );
+            assert_eq!(resolver.find(name, &buffer), Some(old_ns));
+        }
+
+        /// That test add a second level of namespaces, which reset the previous binding
```
dralley, line 632:
```suggestion
        /// Test adding a second level of namespaces, which reset the previous binding
```

src/name.rs
```diff
@@ -380,3 +380,237 @@ impl NamespaceResolver {
...
+            assert_eq!(
+                resolver.resolve(name, &buffer, false),
+                (Some(ns), LocalName(b"with-declared-prefix"))
+            );
+            assert_eq!(resolver.find(name, &buffer), Some(ns));
+        }
+
+        /// That test add a second level of namespaces, which replaces the previous binding
+        #[test]
+        fn override_() {
```
dralley, line 707:
```suggestion
        fn override_namespace() {
```

src/name.rs
```diff
@@ -380,3 +380,237 @@ impl NamespaceResolver {
...
+                (Some(ns), LocalName(b"with-declared-prefix"))
+            );
+            assert_eq!(
+                resolver.resolve(name, &buffer, false),
+                (Some(ns), LocalName(b"with-declared-prefix"))
+            );
+            assert_eq!(resolver.find(name, &buffer), Some(ns));
+        }
+
+        /// That test add a second level of namespaces, which replaces the previous binding
```
dralley, line 705:
```suggestion
        /// Test adding a second level of namespaces, which resets the previous binding
```

src/name.rs
```diff
@@ -380,3 +380,237 @@ impl NamespaceResolver {
...
+                (Some(old_ns), LocalName(b"with-declared-prefix"))
+            );
+            assert_eq!(
+                resolver.resolve(name, &buffer, false),
+                (Some(old_ns), LocalName(b"with-declared-prefix"))
+            );
+            assert_eq!(resolver.find(name, &buffer), Some(old_ns));
+        }
+
+        /// That test add a second level of namespaces, which reset the previous binding
```
dralley, line 742:
```suggestion
        /// Test adding a second level of namespaces, which resets the previous binding
```

src/events/mod.rs
```diff
@@ -156,22 +155,21 @@ impl<'a> BytesStart<'a> {
 
     /// Creates new paired close tag
     pub fn to_end(&self) -> BytesEnd {
-        BytesEnd::borrowed(self.name())
+        BytesEnd::borrowed(self.name().into_inner())
     }
 
     /// Gets the undecoded raw tag name as a `&[u8]`.
```
dralley, line 161: Should these comments be changed?

src/events/mod.rs
```diff
@@ -156,22 +155,21 @@ impl<'a> BytesStart<'a> {
...
     #[inline]
-    pub fn name(&self) -> &[u8] {
-        &self.buf[..self.name_len]
+    pub fn name(&self) -> QName {
+        QName(&self.buf[..self.name_len])
     }
 
     /// Gets the undecoded raw local tag name (excluding namespace) as a `&[u8]`.
```
dralley, line 167: Same

src/events/mod.rs
```diff
@@ -643,20 +641,16 @@ impl<'a> BytesEnd<'a> {
 
     /// Gets `BytesEnd` event name
     #[inline]
-    pub fn name(&self) -> &[u8] {
-        &*self.name
+    pub fn name(&self) -> QName {
+        QName(&*self.name)
     }
 
     /// local name (excluding namespace) as &[u8] (without eventual attributes)
```
dralley, line 648: Same

src/name.rs
```diff
@@ -71,6 +71,41 @@ impl<'a> QName<'a> {
...
+    ///
+    /// ```
+    /// # use quick_xml::name::{QName, PrefixDeclaration};
+    /// let qname = QName(b"xmlns");
+    /// assert_eq!(qname.as_namespace_binding(), Some(PrefixDeclaration::Default));
+    ///
+    /// let qname = QName(b"xmlns:prefix");
+    /// assert_eq!(qname.as_namespace_binding(), Some(PrefixDeclaration::Named(b"prefix")));
+    ///
+    /// // Be notice, that method does not check validity of prefix - it can be empty!
```
dralley, line 87:
```suggestion
    /// // Be aware that this method does not check the validity of the prefix - it can be empty!
```

dralley:
Code looks fine, review finished.  One remaining general question: 

Could the namespace buffer be made an internal part of `NamespaceResolver`? `NamespaceResolver` carries a lot of internal details about the buffer around, such as indexes (via `NamespaceEntry`), and it's not immediately clear if there is still a reason for the buffer to be user-provided.  It's effectively part of the Reader's overall state.  And using more than one different buffer could potentially be dangerous in terms of corrupting that internal logic.

dralley (approved):

PR #395, opened 2022-06-05
Add support for XML Schema `xs:list` type

Mingun:
This PR allows deserializing `Vec` and other sequences from space-separated lists, as [defined](https://www.w3.org/TR/xmlschema11-2/#atomic-vs-list) by XML Schema specification. So now this is possible:
Deserialize
```rust
struct AnyName {
  // Will contain vec!["first", "second", "third"]
  attribute: Vec<String>,
}
```
from
```xml
<any-tag attribute="first second third">
```

### DIFF
--- a/src/de/simple_type.rs
+++ b/src/de/simple_type.rs
@@ -1,0 +1,557 @@
+//! Contains Serde `Deserializer` for XML [simple types] [as defined] in the XML Schema.
+//!
+//! [simple types]: https://www.w3schools.com/xml/el_simpletype.asp
+//! [as defined]: https://www.w3.org/TR/xmlschema11-1/#Simple_Type_Definition
+
+use crate::de::str2bool;
+use crate::errors::serialize::DeError;
+use crate::escape::unescape;
+use serde::de::{DeserializeSeed, Deserializer, EnumAccess, VariantAccess, Visitor};
+use serde::{self, serde_if_integer128};
+use std::borrow::Cow;
+
+macro_rules! deserialize_num {
+    ($method:ident, $visit:ident) => {
+        fn $method<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+        where
+            V: Visitor<'de>,
+        {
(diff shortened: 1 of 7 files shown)


src/de/simple_type.rs
```diff
@@ -0,0 +1,1311 @@
...
+            Self::Owned(ref v) => v,
+        }
+    }
+}
+
+/// A deserializer for an xml probably escaped and encoded value of XSD [simple types].
+/// This deserializer will borrow from the input as much as possible.
+///
+/// When asking for any date, deserializer always return a whole string that
+/// it contain.
```
dralley, line 507: Not sure what this is saying

src/de/simple_type.rs
```diff
@@ -0,0 +1,1311 @@
...
+///
+/// Escaping the value is actually are not always necessary, for instance when
```
dralley, line 509:
```suggestion
/// Escaping the value is actually not always necessary, for instance when
```

dralley:
I tried to do a code review on this PR but since I don't use quick-xml via serde, am not familiar with Serde's parsing, etc. I realized I don't have time to do all of that research, plus dig through the XML spec, plus review the PR, and keep everything in my head long enough to get it done.

So I have just given this a brief look through.  You've done a great job with the testing

src/de/simple_type.rs
Mingun: This is a misprint -- should be "When asking for any data". This sentence should mean, that calling `Deserializer::deserialize_any()` always will call `Visitor::visit_str/visit_borrowed_str/visit_string`, because the string is the only type that XML stores internally (i.e. without type hints you cannot understand what type have, for example, `1` in `<tag>1</tag>`). I agree that this sentence is too non-obvious even when misprint is fixed.

dralley:
Feel free to merge this without a full approval

src/de/simple_type.rs
dralley: Nice explanation, thanks.

PR #396, opened 2022-06-19
Fix caldav example

kornelski:
`BytesStart::unescape_and_decode` on `<D:href>` element is returning `"D:href"`, which isn't the URL. The URL is in the text nodes of the element.


Mingun:
Thanks for contribution. I think, the example could be even more useful if you add an `assert_eq!` in the main with the expected result. Would you like to do that?

Also, you should run `cargo fmt`.

kornelski:
Sorry, I don't use rustfmt. I think it makes too many formatting mistakes, and the tool is not good enough to be used uncritically. It's especially annoying when it's used to block CI.

PR #397, opened 2022-06-19
Deprecate unnecessary unescaping of tags

### LINKED ISSUE #341: what's BytesStart::unescaped for?
I'm trying to switch to `quick-xml` and am struggling to understand the API for just getting strings (for tag names, attribute names, attribute values, and text/cdata nodes).

I saw that the caller must unescaped text/cdata/attribute values, and [this comment](https://github.com/tafia/quick-xml/issues/158#issuecomment-529355417) that the caller must decode the character encoding.

Next question: what is [`BytesStart::unescaped`](https://docs.rs/quick-xml/0.22.0/quick_xml/events/struct.BytesStart.html#method.unescaped) for? It says it handles escapes/entities like `&lt;` but that doesn't make sense to me. Is this the right grammar for the XML tag name from the [spec](https://www.w3.org/TR/xml/#sec-common-syn)?

```text
NameStartChar ::= ":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
NameChar ::= NameStartChar | "-" | "." | [0-9] | #xB7 | [#x0300-#x036F] | [#x203F-#x2040]
Name ::= NameStartChar (NameChar)*
```

It looks like a `&` (aka `#x26`) isn't allowed. So why unescape? or what am I missing?

kornelski:
Fixes  #341

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -179,7 +179,10 @@ impl<'a> BytesStart<'a> {
     /// "`<`".
     ///
     /// See also [`unescaped_with_custom_entities()`](#method.unescaped_with_custom_entities)
-    #[inline]
+    #[cold]
+    #[deprecated(note = "Tag names are never escaped. Attributes must be unescaped individually")]
+    #[allow(deprecated)]
+    #[doc(hidden)]
     pub fn unescaped(&self) -> Result<Cow<[u8]>> {
         self.make_unescaped(None)
     }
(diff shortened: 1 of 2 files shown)


Mingun:
I think, we need to remove that function entirely if is does the wrong thing. I'll hold this PR for a while, because I also prepare a PR that touches many of escaping/decoding functions and I need time to review this and find an initial PR that introduces it to understand what problem it solves.

Anyway, you feel free to change this PR to remove method and update a Changelog.md. Could you also use #118 instead of #341 in messages and in the changelog?

kornelski:
I've hidden these functions from docs, so that they're almost as good as deleted, but don't need semver-major bump. You could delete them next time you release a major version.

PR #399, opened 2022-06-19
Rework BOM handling and encoding API

### LINKED ISSUE #180: encoding feature breaks API
Hi,

the way features work in cargo is that any crate in your dependency tree can enable additional features for its dependencies, and each dependency ("crate") is only compiled once with all features merged; thus they must only add additional "APIs" without breaking what was before.

I.e. enabling the `encoding` feature should provide new structs and functions (or just improve existing ones without changing their signature).

Also the `reader::Decoder` struct is not public and not documented, but `Reader::decoder()` is... I guess either both or none should be public.

### LINKED ISSUE #191: quick-xml captures UTF BOM as Event::Text

### LINKED ISSUE #262: reduce code redundancy for the `encoding` feature
A lot of methods are defined twice in the code, once with `encoding` enabled, and once without it.
Some of these methods have different result types (direct value vs. Result), and this propagate to methods calling them (which must call the former with or without a `?`, depending).
(NB: as pointed out in #180, only private method should have a change of signature, but this is orthogonal, I think)

I suggest the following change:
* instead of returning `T`, private methods in the non-`encoding` configuration should return `std::result::Result<T, std::convert::Infallible>`, which after compilation is equivalent to `T`, but is syntactically is still a result.
* `Error` should implement `From<Infallible>` (again, this is merely to please the compiler during parsing: the `‘from` method will contain an `unreachable!()` statement, since `Infallible` has no value)
* that way, all methods calling the private method can use `?`, regardless of the presence/absence of the `encoding` feature. That way, they don't need two different implementation. The compiler will, however, ignore the `?` when the underlying error type is `Infallible` (i.e. when `encoding` is disabled).

**edited** fixed the path of `Infallible`

Mingun:
- Fixes #180
- Fixes #191. @TakaakiFuruse, could you look at this, whether it solves your problems properly?
- Fixes #262

This PR solves several highly-coupled problems, that blocks changing for safe namespace handling (#59). The main reason in that various `decode` methods accepts a reference to the `Reader`, but if we move namespace buffer inside the `Reader`, we could face with borrowing problems (actually, namespace buffer was decoupled from `Reader` due to this reason in #69).

First, I've fixed #191 by introducing a new `StartText` event which generated when reader encounters "text" before the XML markup. In well-formed XMLs that text could represent a BOM.

Secondly, I've removed all functions that handle BOM in inappropriate time. For example, there a no sense to handle BOM when you decode attribute values, because you will never get BOM there. For that reason I've moved all BOM handling code to the `BytesStartText` struct.

Because after that change you can handle BOM only as the first event from the reader, that allowed me to move the automatic BOM-based encoding detection to the reader instead of triggering it by calling `BytesText::decode_without_bom` (where the name is also incorrect, since the function removes BOM from text and decodes the rest content, so correct one is `decode_with_bom_removal`).

As encoding API already changed very much, I've also solved the #180 by unify function signature with and without `encoding` feature enabled.


Changelog.md
```diff
@@ -55,6 +60,30 @@
 - [#393]: Now `BytesStart::name()` and `BytesEnd::name()` returns `QName`, and
   `BytesStart::local_name()` and `BytesEnd::local_name()` returns `LocalName`
 
+- [#191]: Remove unused `reader.decoder().decode_owned()`. If you ever used it,
+  use `String::from_utf8` instead (which that function did)
+- [#191]: Remove `*_without_bom` methods from the `Attributes` struct because they are useless.
+  Use the same-named methods without that suffix instead. Attribute values cannot contain BOM
+- [#191]: Remove `Reader::decode()` and `Reader::decode_without_bom()`, they was replaced by
```
dralley, line 67:
```suggestion
- [#191]: Remove `Reader::decode()` and `Reader::decode_without_bom()`, they are replaced by
```

Changelog.md
```diff
@@ -55,6 +60,30 @@
...
+- [#191]: Remove poorly designed `BytesText::unescape_and_decode_without_bom()` and
+  `BytesText::unescape_and_decode_without_bom_with_custom_entities()`. Although these methods worked
+  as expected, this was only due to good luck. They was replaced by the
+  `BytesStartText::decode_with_bom_removal()`:
+  - conceptually, you should decode BOM only for the first `Text` event from the
+    reader (since now `StartText` event is emitted instead for this)
+  - text before the first tag is not an XML content at all, so it is meaningless
+    to try to unescape something in it
+
+- [#180]: Define the same API for encoding whether feature `encoding` is defined or not.
```
dralley, line 80: Could this be replaced with "Since the encoding feature no longer performs automatic replacement for invalid characters."?

Changelog.md
```diff
@@ -55,6 +60,30 @@
...
+  Since then no more replacements does for invalid characters when `encoding` feature
+  is enabled. An error will be returned instead, as for the case when `encoding` is disabled.
```
dralley, line 82: Here you say an error will be returned when `encoding` is disabled, but then below you say the opposite?

src/reader.rs
```diff
@@ -16,6 +16,11 @@ use memchr;
 
 #[derive(Clone)]
 enum TagState {
+    /// Initial state in which `StartText`, `Decl`, `Comment` or `Start` event
+    /// is expected
+    Init,
+    /// State after seeing the `<` symbol. After this state reader moves to
+    /// `Closed` state.
```
dralley, line 22: It would be nice if all the variants were documented

dralley (approved):
You do a great job with testing, documentation, splitting up commits and such - I can't find too much to complain about :)

All the changes here look sensible

Changelog.md
Mingun: I'm trying to say, that I've fixed the situation with different `decode*()` signatures when `encoding` feature enabled and when it is disable (returning `Cow<&str>` vs `Result<&str>`). After the fix both realizations returns `Result<Cow<&str>>`

Changelog.md
Mingun: You mean that:
> If `encoding` feature is disabled, `Cow::Borrowed` is returned instead of `&str`.
?
Yes, probably I should write the types completely. It now returns `Result<Cow<&str>>` (instead of `Result<&str>`), always with `Cow::Borrowed` variant

src/reader.rs
Mingun: Yeah, I'm thinking about it but wanted to push earlier to give you time to review the code part. Will add

Mingun:
dralley, I tried to reword some sentences in changelog as best as I can, and also a couple of comments to the `TagState`. Could you take a look at whether is all ok?

Changelog.md
```diff
@@ -56,6 +61,34 @@
...
+  `BytesText::unescape_and_decode_without_bom_with_custom_entities()`. Although these methods worked
+  as expected, this was only due to good luck. They was replaced by the
+  `BytesStartText::decode_with_bom_removal()`:
+  - conceptually, you should decode BOM only for the first `Text` event from the
+    reader (since now `StartText` event is emitted instead for this)
+  - text before the first tag is not an XML content at all, so it is meaningless
+    to try to unescape something in it
+
+- [#180]: Eliminate differences in the decoding API when feature `encoding` enabled and when it is
+  disabled. Since then signatures of functions are the same regardless of the feature enabled and
```
dralley, line 82:
```suggestions
  disabled. Signatures of functions are now the same regardless of whether the feature is enabled and
```

Changelog.md
```diff
@@ -56,6 +61,34 @@
...
+  no more replacements does for invalid characters when feature is enabled. An error will be returned
```
dralley, line 83:
```suggestion
instead of performing replacements for invalid characters when the feature is enabled, as it is when `encoding` is disabled.
```

Changelog.md
dralley, line 82: Eliminated the differences in the decoding API when feature `encoding` enabled and when it is
disabled. Signatures of functions are now the same regardless of whether or not the feature is
enabled, and an error will be returned instead of performing replacements for invalid characters
in both cases.

Changelog.md
```diff
@@ -56,6 +61,34 @@
...
+  instead, as for the case when `encoding` is disabled.
+  Namely, if `encoding` feature is disabled, `Result<Cow<&str>>` is returned instead of `Result<&str>`.
```
dralley, line 85: Previously, if the `encoding` feature was enabled, decoding functions would return `Result<Cow<&str>>` while without this feature they would return `Result<&str>`.  With this change, only `Result<Cow<&str>>` is returned regardless of the status of the feature.

dralley:
I left some suggestions

Mingun:
Thanks! In the final version I've updated benchmarks (they has been broken since #393) and applied you suggestions. Merging now

PR #400, opened 2022-06-22
Remove `BytesStart::unescaped*` set of methods because they could return wrong results

### LINKED ISSUE #118: BytesStart unescaped method returns the attributes in the same result

Mingun:
- Closes #118
- Closes #119 
- Closes #397

I decide to completely remove this set of methods because they could produce misleading results.

I decide to remove example `issue68.rs` because of reasons:
- it used removed method
- it contained a bug, described in #396
- it wasn't a good example at all because of unused code and lacking of comments, instead of fixing that I think it is better to remove it completely, working with namespaces will changed soon anyway


dralley (approved):

dralley:
@Mingun FYI, the benches are broken again

PR #403, opened 2022-06-24
Always use UTF-8 for a `Reader` and a `Deserializer` created from a string

### LINKED ISSUE #324: from_str should ignore encoding in XML declaration
XML can be viewed as a binary format, for instance when parsed using from_reader from a sequence of bytes.
However, when parsing from a &str, the content is already guaranteed to be valid UTF-8.
In that case the encoding hint in the XML declaration should be ignored.

Here is a test that fails (make sure the source code is saved as UTF-8).

```rust
#[test]
fn from_str_should_ignore_encoding() {
    let xml = r#"
        <?xml version="1.0" encoding="windows-1252" ?>
        <A a="€" />
    "#;

    #[derive(Debug, PartialEq, Deserialize)]
    struct A {
        a: String,
    }

    let a: A = quick_xml::de::from_str(xml).unwrap();
    assert_eq!(
        a,
        A {
            a: "€".to_string()
        }
    );
}
```

Test result:
```
---- foo::from_str_should_ignore_encoding stdout ----
thread 'foo::from_str_should_ignore_encoding' panicked at 'assertion failed: `(left == right)`
  left: `A { a: "â‚¬" }`,
 right: `A { a: "€" }`', src/foo.rs:415:5
```

Mingun:
Fixes #324

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -56,6 +56,55 @@ enum TagState {
     Exit,
 }
 
+/// A reference to an encoding together with information about how it was retrieved.
+///
+/// The state transition diagram:
+///
+/// ```mermaid
+/// flowchart LR
+///   Implicit    -- from_str       --> Explicit
+///   Implicit    -- BOM            --> BomDetected
+///   Implicit    -- "encoding=..." --> XmlDetected
+///   BomDetected -- "encoding=..." --> XmlDetected
+/// ```
+#[cfg(feature = "encoding")]
+#[derive(Clone, Copy)]
+enum EncodingRef {
+    /// Encoding was implicitly assumed to have a specified value. It can be refined
+    /// using BOM or by the XML declaration event (`<?xml encoding=... ?>`)
+    Implicit(&'static Encoding),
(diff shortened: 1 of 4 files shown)


src/reader.rs
```diff
@@ -56,6 +56,55 @@ enum TagState {
     Exit,
 }
 
+/// A reference to an encoding together with an information of how it was retrieved.
```
dralley, line 59:
```suggestion
/// A reference to an encoding together with information about how it was retrieved.
```

src/reader.rs
```diff
@@ -905,6 +949,15 @@ impl Reader<BufReader<File>> {
 impl<'a> Reader<&'a [u8]> {
     /// Creates an XML reader from a string slice.
     pub fn from_str(s: &'a str) -> Self {
+        // Rust strings a guaranteed in UTF-8, so lock the encoding
```
dralley, line 952:
```suggestion
        // Rust strings are guaranteed to be UTF-8, so lock the encoding
```

dralley (approved):

PR #408, opened 2022-06-29
Faster escape routines

dralley:


dralley:
Figuring out how much of an improvement this is (or even if it is one, on average) might be challenging because I get different results on my AMD based desktop and my Intel based laptop (although laptop benchmarks are awful regardless).

Escaping long inputs is much faster on both systems regardless of whether it does any replacements, escaping short inputs on both systems is slower if there are replacements, but my desktop is also slower on short inputs without any replacements whereas my laptop is still faster (just not as much as with the long inputs).

Since most attributes are short and don't contain escapable chars, a regression there concerns me a bit.

PR #412, opened 2022-07-08
Rename reading methods, fix some encoding errors in error-processing paths and tests

Mingun:
This PR includes some of the stuff, improving code here and there, including:
- fix incorrect encoding used when handle some errors
- remove excess `BufRead` constraints which opens door for the `async` support fo the same `Reader` struct
- some minor improvements in tests
- some documentation updates
- rename core methods to better reflect their purpose:
  |Old Name                 |New Name
  |-------------------------|---------------------------------------------------
  |`read_event`             |`read_event_into`
  |`read_to_end`            |`read_to_end_into`
  |`read_text`              |`read_text_into`
  |`read_event_unbuffered`  |`read_event`
  |`read_to_end_unbuffered` |`read_to_end`

  (_`read_namespaced_event` is not renamed because I plan to remove it in the namespace PR_)

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -107,7 +107,7 @@ impl EncodingRef {
 
 /// A low level encoding-agnostic XML event reader.
 ///
-/// Consumes a `BufRead` and streams XML `Event`s.
+/// Consumes bytes and streams XML [`Event`]s.
 ///
 /// # Examples
 ///
@@ -144,7 +144,7 @@ impl EncodingRef {
 /// }
 /// ```
 #[derive(Clone)]
-pub struct Reader<R: BufRead> {
+pub struct Reader<R> {
     /// reader
     pub(crate) reader: R,
     /// current buffer position, useful for debugging errors
@@ -198,8 +198,8 @@ pub struct Reader<R: BufRead> {
 }
 
 /// Builder methods
-impl<R: BufRead> Reader<R> {
-    /// Creates a `Reader` that reads from a reader implementing `BufRead`.
+impl<R> Reader<R> {
+    /// Creates a `Reader` that reads from a given reader.
     pub fn from_reader(reader: R) -> Self {
         Self {
             reader,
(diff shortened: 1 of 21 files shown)


src/reader.rs
```diff
@@ -107,7 +107,7 @@ impl EncodingRef {
 
 /// A low level encoding-agnostic XML event reader.
 ///
-/// Consumes a `BufRead` and streams XML `Event`s.
+/// Consumes a bytes and streams XML [`Event`]s.
```
dralley, line 110:
```suggestion
/// Consumes bytes and streams XML [`Event`]s.
```

tests/xmlrs_reader_tests.rs
```diff
@@ -403,21 +404,21 @@ fn namespace_name(n: ResolveResult, name: QName, decoder: Decoder) -> String {
...
     for a in e.attributes() {
         match a {
             Ok(a) => {
                 if a.key.as_namespace_binding().is_none() {
                     atts.push(format!(
                         "{}=\"{}\"",
-                        from_utf8(a.key.as_ref()).unwrap(),
-                        from_utf8(&*a.unescaped_value().unwrap()).unwrap()
+                        decoder.decode(a.key.as_ref()).unwrap(),
+                        decoder.decode(&*a.unescaped_value().unwrap()).unwrap()
```
dralley, line 421: This seems like it would be another instance of the incorrect "unescape, then decode" behavior mentioned, yes?

dralley:
In the future, something like https://crates.io/crates/buffered-reader might be nice - the internal buffer can dynamically resize whic would allow avoiding the copy between the `BufReader` buffer and an external one.

Unfortunately, at the moment, it is LGPL licensed which makes it a bit challenging to depend on.

tests/xmlrs_reader_tests.rs
Mingun: Yeah, you right. Need to fix that too

dralley (approved):

PR #413, opened 2022-07-09
Make attribute creation more uniform

### LINKED ISSUE #410: Using attribute-from-tuple conversion can be error-prone
Creating an attribute from a `(&[u8], &[u8])` performs no transformation (escaping) on the provided value, but creating an attribute from `(&str, &str)` does.

This feels a bit too implicit.  It should probably be treated like `BytesText` instead, where different methods are provided depending on whether the value has been pre-escaped.  https://docs.rs/quick-xml/latest/quick_xml/events/struct.BytesText.html

```Rust
use quick_xml::events::attributes::Attribute;

fn main() {
    // correct
    println!("{:#?}", Attribute::from(("foo", "Bells & Whistles")));
    println!("{:#?}", Attribute::from(("foo".as_bytes(), "Bells &amp; Whistles".as_bytes())));

    // incorrect
    println!("{:#?}", Attribute::from(("foo", "Bells &amp; Whistles")));
    println!("{:#?}", Attribute::from(("foo".as_bytes(), "Bells & Whistles".as_bytes())));
}
```

```
Attribute { key: "foo", value: Owned("Bells &amp; Whistles") }
Attribute { key: "foo", value: Borrowed("Bells &amp; Whistles") }
Attribute { key: "foo", value: Owned("Bells &amp;amp; Whistles") }
Attribute { key: "foo", value: Borrowed("Bells & Whistles") }
```

dralley:
closes #410

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,8 @@
 - [#395]: Add support for XML Schema `xs:list`
 - [#324]: `Reader::from_str` / `Deserializer::from_str` / `from_str` now ignore
   the XML declared encoding and always use UTF-8
+- [#413]: Added `Attribute::from_bytes()`, `Attribute::from_escaped_bytes()`,
+  `Attribute::from_str()` and `Attribute::from_escaped_str()` constructors.
 - [#416]: Add `borrow()` methods in all event structs which allows to get
   a borrowed version of any event
 
@@ -42,6 +44,9 @@
 - [#363]: Do not generate empty `Event::Text` events
 - [#412]: Fix using incorrect encoding if `read_to_end` family of methods or `read_text`
   method not found a corresponding end tag and reader has non-UTF-8 encoding
+- [#413]: `Attribute::from((&[u8], &[u8]))` was made consistent with `Attribute::from((&str, &str))`
+  in that both will now escape the value. WARNING: if you were passing escaped values via this
+  approach before, your values will be double-escaped and incorrect.
 
 ### Misc Changes
 
@@ -137,6 +142,7 @@
 [#403]: https://github.com/tafia/quick-xml/pull/403
 [#407]: https://github.com/tafia/quick-xml/pull/407
 [#412]: https://github.com/tafia/quick-xml/pull/412
+[#413]: https://github.com/tafia/quick-xml/pull/413
 [#416]: https://github.com/tafia/quick-xml/pull/416
 [#418]: https://github.com/tafia/quick-xml/pull/418
 
--- a/src/events/attributes.rs
+++ b/src/events/attributes.rs
@@ -30,6 +30,82 @@ pub struct Attribute<'a> {
 }
 
 impl<'a> Attribute<'a> {
+    /// Creates new attribute from text representation.
+    /// Key is stored as-is, but the value will be escaped.
+    ///
+    /// # Examples
+    ///
+    /// ```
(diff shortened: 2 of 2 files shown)


dralley:
See question: https://github.com/tafia/quick-xml/issues/410#issuecomment-1179479638=

src/events/attributes.rs
```diff
@@ -32,6 +32,25 @@ pub struct Attribute<'a> {
...
+    /// Creates new attribute from raw bytes.
+    /// Does not apply any transformation to either key or value.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use pretty_assertions::assert_eq;
+    /// use quick_xml::events::attributes::Attribute;
+    ///
+    /// let features = Attribute::new("features".as_bytes(), "Bells &amp; whistles".as_bytes());
```
Mingun, line 44: I think, this looks better:
```suggestion
    /// let features = Attribute::new(b"features", b"Bells &amp; whistles");
```

dralley: The type of this is `[u8; <len>]` rather than `&[u8]`, so you have to do a bit extra, and at that point it's not cleaner anymore.

Changelog.md
```diff
@@ -40,6 +42,9 @@
 - [#363]: Do not generate empty `Event::Text` events
 - [#412]: Fix using incorrect encoding if `read_to_end` family of methods or `read_text`
   method not found a corresponding end tag and reader has non-UTF-8 encoding
+- [#413]: `Attribute::from((&[u8], &[u8]))` was made consistent with `Attribute::from((&str, &str))`
+  in that both will now escape the value. WARNING: if you were passing escaped values via this
```
dralley, line 46: I'm not sure how to draw more attention to this but it is a somewhat consequential change and users won't get a compile error.

dralley:
@Mingun I think it makes sense, now, to make the members of `Attribute` private and use accessors e.g. `raw_value()`, `normalized_value()`, `unescaped_value()`.  Do you agree?

Also, do you forsee any reason for the key type to be `Into<QName>` as opposed to `&[u8]`?

src/events/attributes.rs
Mingun: This is only when you apply that suggestion to the tuple, because `(b"...", b"...")` will have a type `([u8; <len1>], [u8; <len2>])`. Using byte literals with `from_*` methods is compiled, I checked

Mingun:
With private fields it will be impossible to use `Attribute` in patterns which could break some usages. From the other hand, such usage is potentially unsafe, because `Attribute` could contain data in unknown encoding. I think it is better to introduce such change together with fixing accessors in #379.

> Also, do you forsee any reason for the key parameter type to be `Into<QName>` as opposed to `&[u8]`?

Oh, yees... I totally forget, that I introduced `QName` for type safety which means, that user should always _explicitly_ use it instead of plain `&[u8]`, otherwise it just have no point. So yes, it should be `QName` and could be `Into<QName>`

dralley:
>user should always *explicitly* use it instead of plain &[u8]

Well if this is the goal, `Into<QName>` doesn't make sense and neither does the current  `From<(.., ..)>` syntax.  Both of those allow implicit conversion from bytes.

dralley:
This PR won't be relevant for much longer.

Issue #322, opened 2021-09-27
Cannot parse UTF-16 xml


BlueGreenMagick:
UTF-16 xml parsing does not seem to work at all.

# Current Behaviour

To test this, try changing this function to print tag names. https://github.com/tafia/quick-xml/blob/118c07b456e5a4bb730ca6d762a261ab525784da/tests/unit_tests.rs#L904-L921

```rust
#[test]
#[cfg(feature = "encoding")]
fn test_unescape_and_decode_without_bom_removes_utf16le_bom() {
    let mut reader = Reader::from_file("./tests/documents/utf16le.xml").unwrap();
    reader.trim_text(true);

    let mut txt = Vec::new();
    let mut buf = Vec::new();

    loop {
        match reader.read_event(&mut buf) {
            Ok(Event::Text(e)) => txt.push(e.unescape_and_decode_without_bom(&mut reader).unwrap()),
            Ok(Event::Eof) => break,
            Ok(Event::Start(e)) => txt.push(reader.decode(&e.local_name()).to_string()), // add this line
            _ => (),
        }
    }
   // print the content of txt
    println!("{:?}", txt); 
    panic!();
    assert_eq!(txt[0], "");
}
```

You would expect to see something like `["", "project"]`.
But instead, this comes out: `["", "㼀砀洀氀�", "\u{a00}�", "瀀爀漀樀攀挀琀�", "\u{a00}�", "⼀瀀爀漀樀攀挀琀�", "\u{a00}�"]`

# Likely Reason

It looks like the parser compares _a single byte_ to check if that character is `b'<'`(`0x3C`) or other special characters, then increments the position by 1 byte. But in UTF-16, the `<` character is actually two bytes `0x3C 0x00`. 

So when the parser tries to parse `<?...`, which in UTF-8 is `0xFF 0xFE 0x3C 0x00 0x3F 0x00 ...`, It consumes `0x3C`(`<`) and thinks the element's raw name is `0x00 0x3F 0x00 ...`. (`㼀` is `0x00 0x3F`)

Changing the above code's line to `txt.push(reader.decode(&e.local_name()[1..]).to_string())` makes the element names come out correct. (Even though `Event::End` and `Event::Decl` is lumped with `Event::Start`.):  `["", "?xml", "\u{a00}�", "project", "\u{a00}�", "/project", "\u{a00}�"]`.

dralley:
Considering this a duplicate of https://github.com/tafia/quick-xml/issues/158

Issue #158, opened 2019-05-17
ergonomics & encodings


cmyr:
Hi there,

I've just been playing around with quick-xml to handle reading and writing of [Glyph Interchange Format](http://unifiedfontobject.org/versions/ufo3/glyphs/glif/#contour) files, and I had some observations I wanted to share.

## `[u8]`/`Cow<[u8]>` vs `Cow<str>`

My understanding is that `quick-xml` attempts to do the least work possible, and to this end exposes byte slices by default. I can understand the reasoning (don't allocate unless the user has asked for it explicitly) but there are two things I keep thinking about, here:

- _utf-8 is common_: the majority of XML I've been dealing with is encoded in utf-8. In this case, we should be able to provide direct access to string slices, allocating only if we need to escape something.

- _In a non-utf8 encoding, it is hard to use bytes correctly_: exposing bytes directly encourages users to do things like `if attr.key == b"mykey"` (which is used in the readme). This is fine with an ascii-compatible encoding like utf-8 or Shift JIS, but will fail unexpectly on, say, utf-16; and utf-16 is an explicitly supported encoding according to the [xml spec](https://www.w3.org/TR/REC-xml/#charencoding).

## API thoughts

I haven't thought about this too too much, but I do worry that the current behaviour has worse ergonomics than necessary in the general (utf-8) case, while also making it easy to write bugs in the exceptional (utf-16) case.

One possible alternative I could imagine would be having separate `Event` and `EventRaw` types; `Event` would work with `Cow<str>`, which would be slices of the underlying buffer where possible (when the encoding is utf-8) and which would be decoded and allocated during parsing, otherwise. `EventRaw` would work like the current `event`.

Another option would be to have add some new type, `XmlSlice`, which would include a lazily-allocating `to_str` method, while also exposing access to the underlying bytes and even the encoding; this type would be used everywhere that `[u8]` or `Cow<[u8]>` is, currently. This could also include information like the location of this text in the original document.

## thanks

I want to clarify that this is not a feature request or a concrete proposal for a new API; I just wanted to voice these thoughts and hear if anyone had anything to add. It may be that what I'm describing should just be an alternative crate, or that there are other design constraints I'm not aware of.

Thank you for the time you've put into this code so far!

tafia:
Thanks for reaching out, these kind of feedbacks are very useful!

[...]

PR #415, opened 2022-07-10
Closure-based unescaping with custom entities

dralley:
Re: https://github.com/tafia/quick-xml/pull/379#discussion_r916490057=

### DIFF
--- a/examples/custom_entities.rs
+++ b/examples/custom_entities.rs
@@ -7,10 +7,11 @@
 //! * the regex in this example is simple but brittle;
 //! * it does not support the use of entities in entity declaration.
 
+use std::collections::HashMap;
+
 use quick_xml::events::Event;
 use quick_xml::Reader;
 use regex::bytes::Regex;
-use std::collections::HashMap;
 
 const DATA: &str = r#"
 
@@ -27,35 +28,41 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     reader.trim_text(true);
 
     let mut buf = Vec::new();
-    let mut custom_entities = HashMap::new();
+    let mut custom_entities: HashMap<Vec<u8>, String> = HashMap::new();
     let entity_re = Regex::new(r#"<!ENTITY\s+([^ \t\r\n]+)\s+"([^"]*)"\s*>"#)?;
 
     loop {
         match reader.read_event_into(&mut buf) {
             Ok(Event::DocType(ref e)) => {
                 for cap in entity_re.captures_iter(&e) {
-                    custom_entities.insert(cap[1].to_vec(), cap[2].to_vec());
+                    custom_entities.insert(
+                        cap[1].to_vec(),
+                        reader.decoder().decode(&cap[2])?.into_owned(),
+                    );
                 }
             }
             Ok(Event::Start(ref e)) => match e.name().as_ref() {
-                b"test" => println!(
-                    "attributes values: {:?}",
-                    e.attributes()
-                        .map(|a| a
-                            .unwrap()
-                            .unescape_and_decode_value_with_custom_entities(
-                                &reader,
-                                &custom_entities
-                            )
-                            .unwrap())
-                        .collect::<Vec<_>>()
-                ),
+                b"test" => {
+                    let attributes = e
+                        .attributes()
+                        .map(|a| {
+                            a.unwrap()
+                                .decode_and_unescape_value_with(&reader, |ent| {
+                                    custom_entities.get(ent).map(|s| s.as_str())
+                                })
+                                .unwrap()
+                        })
+                        .collect::<Vec<_>>();
+                    println!("attributes values: {:?}", attributes);
+                }
                 _ => (),
             },
             Ok(Event::Text(ref e)) => {
                 println!(
                     "text value: {}",
-                    e.unescape_and_decode_with_custom_entities(&reader, &custom_entities)
+                    e.decode_and_unescape_with(&reader, |ent| custom_entities
+                        .get(ent)
+                        .map(|s| s.as_str()))
                         .unwrap()
                 );
             }
--- a/src/escapei.rs
+++ b/src/escapei.rs
@@ -2,7 +2,6 @@
 
 use memchr;
 use std::borrow::Cow;
-use std::collections::HashMap;
 use std::ops::Range;
 
 #[cfg(test)]
@@ -66,15 +65,7 @@ impl std::error::Error for EscapeError {}
 /// Escapes a `&[u8]` and replaces all xml special characters (<, >, &, ', ") with their
 /// corresponding xml escaped value.
 pub fn escape(raw: &[u8]) -> Cow<[u8]> {
-    #[inline]
-    fn to_escape(b: u8) -> bool {
-        match b {
-            b'<' | b'>' | b'\'' | b'&' | b'"' => true,
-            _ => false,
-        }
-    }
-
-    _escape(raw, to_escape)
+    _escape(raw, |ch| matches!(ch, b'<' | b'>' | b'&' | b'\'' | b'\"'))
 }
 
 /// Should only be used for escaping text content. In xml text content, it is allowed
@@ -82,15 +73,7 @@ pub fn escape(raw: &[u8]) -> Cow<[u8]> {
 /// This function escapes a `&[u8]` and replaces xml special characters (<, >, &) with
 /// their corresponding xml escaped value, but does not escape quote characters.
 pub fn partial_escape(raw: &[u8]) -> Cow<[u8]> {
-    #[inline]
-    fn to_escape(b: u8) -> bool {
-        match b {
-            b'<' | b'>' | b'&' => true,
-            _ => false,
-        }
-    }
-
-    _escape(raw, to_escape)
+    _escape(raw, |ch| matches!(ch, b'<' | b'>' | b'&'))
 }
 
 /// Escapes a `&[u8]` and replaces a subset of xml special characters (<, >, &, ', ") with their
@@ -130,32 +113,23 @@ fn _escape<F: Fn(u8) -> bool>(raw: &[u8], escape_chars: F) -> Cow<[u8]> {
 /// Unescape a `&[u8]` and replaces all xml escaped characters ('&...;') into their corresponding
 /// value
 pub fn unescape(raw: &[u8]) -> Result<Cow<[u8]>, EscapeError> {
-    do_unescape(raw, None)
+    unescape_with(raw, |_| None)
 }
 
 /// Unescape a `&[u8]` and replaces all xml escaped characters ('&...;') into their corresponding
-/// value, using a dictionnary of custom entities.
+/// value, using a resolver function for custom entities.
 ///
 /// # Pre-condition
 ///
-/// The keys and values of `custom_entities`, if any, must be valid UTF-8.
-pub fn unescape_with<'a>(
-    raw: &'a [u8],
-    custom_entities: &HashMap<Vec<u8>, Vec<u8>>,
-) -> Result<Cow<'a, [u8]>, EscapeError> {
-    do_unescape(raw, Some(custom_entities))
-}
-
-/// Unescape a `&[u8]` and replaces all xml escaped characters ('&...;') into their corresponding
-/// value, using an optional dictionary of custom entities.
-///
-/// # Pre-condition
-///
-/// The keys and values of `custom_entities`, if any, must be valid UTF-8.
-pub fn do_unescape<'a>(
-    raw: &'a [u8],
-    custom_entities: Option<&HashMap<Vec<u8>, Vec<u8>>>,
-) -> Result<Cow<'a, [u8]>, EscapeError> {
+/// The implementation of `resolve_entity` is expected to operate over UTF-8 inputs.
+pub fn unescape_with<'input, 'entity, F>(
+    raw: &'input [u8],
+    resolve_entity: F,
+) -> Result<Cow<'input, [u8]>, EscapeError>
+where
+    // the lifetime of the output comes from a capture or is `'static`
+    F: Fn(&[u8]) -> Option<&'entity str>,
+{
     let mut unescaped = None;
     let mut last_end = 0;
     let mut iter = memchr::memchr2_iter(b'&', b';', raw);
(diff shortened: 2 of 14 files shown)


dralley:
@Mingun At the moment, this is not working for the `custom_entities.rs` example because the lifetimes get rather complex.

src/escapei.rs
```diff
@@ -130,32 +113,22 @@ fn _escape<F: Fn(u8) -> bool>(raw: &[u8], escape_chars: F) -> Cow<[u8]> {
...
-/// value, using an optional dictionary of custom entities.
-///
-/// # Pre-condition
-///
-/// The keys and values of `custom_entities`, if any, must be valid UTF-8.
-pub fn do_unescape<'a>(
-    raw: &'a [u8],
-    custom_entities: Option<&HashMap<Vec<u8>, Vec<u8>>>,
-) -> Result<Cow<'a, [u8]>, EscapeError> {
+    lookup_custom_entity: impl Fn(&'b [u8]) -> Option<&'b str>,
```
Mingun, line 127: You need to use [Higher-Rank Trait Bounds](https://doc.rust-lang.org/nomicon/hrtb.html#higher-rank-trait-bounds-hrtbs) to fix lifetime issues.

Also it is better to name parameter `resolve_entity`

dralley: Seems like there's still a lot of compile errors.  I cannot tell, if it is trying to get me to annotate this signature on top of the closure as well, but if so that would be a very user-unfriendly API.  Or maybe I'm just missing something (relatively) simple to fix.

dralley: It looks like some of the errors are actually Rust issues.. https://github.com/rust-lang/rust/issues/75791

dralley: Not all of them though.  Seems like Rust also has a difficult time recognizing that in each individual iteration of the loop only a single mutable or non-mutable reference can exist.

Mingun: Corrent implementation requires a bit more time than I expected, so I will postpone the task for a while. This requires a big work for review of encoding support, so not before 0.25.0

dralley:
>It is better to replace suffix _with_custom_entities to _with -- this is more similar how other Rust APIs is named

I agree but when it comes to `unescape_and_decode_with()` the name sounds misleading - it sounds like the "with" refers to the decoding portion instead of the unescaping portion.

As with the corrections you've made recently, the more appropriate name would be `decode_and_unescape_with()`.  Thoughts?

src/events/attributes.rs
```diff
@@ -57,18 +57,11 @@ impl<'a> Attribute<'a> {
     /// # Pre-condition
     ///
     /// The keys and values of `custom_entities`, if any, must be valid UTF-8.
```
Mingun, line 59: Do not forget to update documentation on all methods

examples/custom_entities.rs
```diff
@@ -27,35 +28,41 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let entity_re = Regex::new(r#"<!ENTITY\s+([^ \t\r\n]+)\s+"([^"]*)"\s*>"#)?;
 
     loop {
+        let lookup_custom_entity = |ent: &_| custom_entities.get(ent).map(|s| s.as_str());
+
         match reader.read_event_into(&mut buf) {
             Ok(Event::DocType(ref e)) => {
                 for cap in entity_re.captures_iter(&e) {
-                    custom_entities.insert(cap[1].to_vec(), cap[2].to_vec());
+                    custom_entities.insert(cap[1].to_vec(), String::from_utf8(cap[2].to_vec())?);
```
Mingun, line 40: It is better to use `reader.decoder().decode()` for decoding. This is an example, I think, it is better to show the right way to get strings
```suggestion

                    let value = reader.decoder().decode(&cap[2])?.into_owned();
                    custom_entities.insert(cap[1].to_vec(), value);
```

examples/custom_entities.rs
```diff
@@ -27,35 +28,41 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     reader.trim_text(true);
 
     let mut buf = Vec::new();
-    let mut custom_entities = HashMap::new();
+    let mut custom_entities: HashMap<Vec<u8>, String> = HashMap::new();
     let entity_re = Regex::new(r#"<!ENTITY\s+([^ \t\r\n]+)\s+"([^"]*)"\s*>"#)?;
 
     loop {
+        let lookup_custom_entity = |ent: &_| custom_entities.get(ent).map(|s| s.as_str());
```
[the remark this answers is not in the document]
dralley: At the source level though, it's still an annoying bit of code to ask the user to write constantly.  And it is long enough to start breaking line limits if you write it in the invocation point.

src/events/attributes.rs
dralley: Right, it's just a draft, looking mostly for technical feedback

dralley:
@Mingun I'm not 100% sure if this makes sense, obviously you've put more thought into decoding, but:

Given that encoding is a property of the source file, and not something that changes on an element-by-element basis, does it really make sense for the complexity of decoding to be distributed throughout all the different structs?

I feel like the reader ought to just decode everything into the buffer as utf-8, and thereafter `Event*` and `Attribute*` and `*escape*` can just assume that it's working with UTF-8.  Instead of reading the bytes directly and trying to make all the machinery work with it.  And likewise on the writing side, working with UTF-8 and then converting it to UTF-16 as it gets written seems much simpler.

From a performance perspective, this is better anyway, since encoding / decoding huge blocks of data is easy to accelerate with SIMD (which it looks like encoding_rs tries to do when possible).

Mingun:
I would be glad to decode the input at time of get them from the underlying reader, even at cost of some minor performance penalty. But need to make this with caution, I trying to keep the spirit of @tafia's implementation -- do not do work if it is unrequired

dralley:
(BTW, @tafia , feel free to weigh in).

The way I see this is that if the user actually does want / need to support different encodings, the current approach will actually require more work (both human and computer) to be done.  

Because using the `unescape_and_decode()` functions means allocating a large number of small individual strings, which is a cost they need to pay regardless of which encoding they're using (even if they're using utf-8, they have to pay for all those allocations now), on top of the penalty of decoding small strings instead of large blocks.

dralley:
Also, the more I think about it, the more implementing parsing for each encoding sounds like a nightmare as well as being very difficult to test.

examples/custom_entities.rs
```diff
@@ -27,35 +28,41 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                }
                 _ => (),
             },
             Ok(Event::Text(ref e)) => {
                 println!(
                     "text value: {}",
-                    e.unescape_and_decode_with_custom_entities(&reader, &custom_entities)
+                    e.decode_and_unescape_with(&reader, |ent| custom_entities
+                        .get(ent)
+                        .map(|s| s.as_str()))
```
dralley, line 65: Open question: since this API is slightly more difficult to work with, should we provide a helper that provides the facade of the older API

Mingun: I think that this is not required

src/events/attributes.rs
```diff
@@ -37,39 +37,33 @@ impl<'a> Attribute<'a> {
...
-        &self,
-        custom_entities: &HashMap<Vec<u8>, Vec<u8>>,
-    ) -> XmlResult<Cow<[u8]>> {
-        self.make_unescaped_value(Some(custom_entities))
+    /// The implementation of `resolve_entity` is expected to operate over UTF-8 inputs.
+    pub fn unescaped_value_with<'s, 'entity>(
+        &'s self,
+        resolve_entity: impl Fn(&[u8]) -> Option<&'entity str>,
+    ) -> XmlResult<Cow<'s, [u8]>> {
+        unescape_with(&*self.value, resolve_entity).map_err(Error::EscapeError)
```
Mingun, line 63: I think it will be more consistently with other codebase

```suggestion
    #[cfg(not(feature = "encoding"))]
    pub fn unescape_value_with<'s, 'entity>(
        &'s self,
        resolve_entity: impl Fn(&[u8]) -> Option<&'entity str>,
    ) -> XmlResult<Cow<'s, [u8]>> {
        Ok(unescape_with(&*self.value, resolve_entity)?)
```

The added feature gate prevents incorrect usage. All usages in tests should be replaced with `decode_*` variant. Another way to solve that -- add a decoder to the `Attribute` / `Event` and keep only variant without `reader` parameter, but I think that this may be a task for another PR, focused on encoding.

src/events/mod.rs
```diff
@@ -740,83 +739,71 @@ impl<'a> BytesText<'a> {
...
-    }
-
-    fn make_unescaped<'s>(
+    pub fn unescaped_with<'s, 'entity>(
         &'s self,
-        custom_entities: Option<&HashMap<Vec<u8>, Vec<u8>>>,
+        resolve_entity: impl Fn(&[u8]) -> Option<&'entity str>,
     ) -> Result<Cow<'s, [u8]>> {
-        do_unescape(self, custom_entities).map_err(Error::EscapeError)
+        unescape_with(self, resolve_entity).map_err(Error::EscapeError)
```
Mingun, line 776:
```suggestion
        Ok(unescape_with(self, resolve_entity)?)
```

Mingun (changes requested):
It is almost good, only few small nits are better to solve:
- change `*ed` method names to `*e` names for further consistency
- add feature gate for methods that assumes UTF-8 encoding, which should restrict their incorrect usage

src/events/attributes.rs
dralley: I would rather push all the decoding changes into a different PR.  At the moment all of the parsing code is based on searches for single ascii bytes, which means it is effectively a UTF-8 only library, even if the detection of encodings works properly.  https://github.com/tafia/quick-xml/issues/322

So there's a lot of work needed to make the `encoding` feature actually useful.

Mingun: Yes, I understand and support. That is why I suggest just disable dangerous methods when `"encoding"` feature is enabled.

> So there's a lot of work needed to make the `encoding` feature actually useful.

Actually, it is useful for one-byte encodings. Bearing in mind, that
- we use `encoding_rs`
- [`encoding_rs` is not extensible by design](https://docs.rs/encoding_rs/latest/encoding_rs/index.html#no-extensibility-by-design)
- as the result, the list of encodings is fixed
- the only encodings, that is not XML compatible (because is not ASCII compatible, which is more strict restriction) is (generated by [this snippet](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=28602c63d99403b1edb96db8c615875d)):
  - `ISO-2022-JP`
  - `replacement` (is not a real encoding, actually)
  - `UTF-16BE`
  - `UTF-16LE`

So the `encoding` feature actually may be useful even in the current state.

----
The other suggestion is to make namings is even more consistently, compare:
- `unescaped`
- `decode_and_unescape`

Because the difference only in the `decode` operation applied in the second function, it is logically, that stripping out the `decode_and_` prefix we should get a name of the method that works without decoding

dralley: You are right about the other one-byte encodings, I just don't know how common they actually are, certainly compared to UTF-16.  But probably what I should have said is that I don't really want to go back and forth over functions that it might make sense to remove entirely at some point in the future.

dralley: I will go ahead and make the changes anyway (tomorrow, since I really need to go to sleep now)

dralley:
I moved all the thoughts about future encoding strategy to the appropriate issue https://github.com/tafia/quick-xml/issues/158, and did a bit of background research on performance

src/events/attributes.rs
dralley: @Mingun I wont forget about this, but I am going to put this in a separate PR.  I want to finish running some experiments first re: bulk decode vs. decoding individual elements and the outcome of that might influence what we want the API to look like.

Mingun: I'm totally agree that encoding part should be in a separate PR, but can you polish namings in this PR? Only one simple change:
- `unescaped()` -> `unescape()`
- `unescaped_with()` -> `unescape_with()`
- `unescaped_value()` -> `unescape_value()`
- `unescaped_value_with()` -> `unescape_value_with()`

As you already changed some of that names it is quite logically to finish with that here, just to keep history clean :). And after that it can be merged

dralley: OK :)  but, please take a look at https://github.com/tafia/quick-xml/issues/158#issuecomment-1185081402

I'm going to draft this for now because I'm 100% convinced we should change the APIs to pure `&str` / `String` ones.

Mingun: > I'm going to draft this for now because I'm 100% convinced we should change the APIs to pure `&str` / `String` ones.

I agree, I already work in this direction on top of this PR, namely changing the order of unescape and decode operations in the serde deserializer. I think I'm going to merge it now and continue evolving API in the another PRs. At least I want to make the new closure-based API accepts `&str` instead of `&[u8]` before the release.

src/events/attributes.rs
```diff
@@ -37,83 +37,67 @@ impl<'a> Attribute<'a> {
...
     ///
     /// This allocates a `String` in all cases. For performance reasons it might be a better idea to
     /// instead use one of:
     ///
     /// * [`Reader::decoder().decode()`], as it only allocates when the decoding can't be performed otherwise.
-    /// * [`unescaped_value()`], as it doesn't allocate when no escape sequences are used.
+    /// * [`unescape_value()`], as it doesn't allocate when no escape sequences are used.
     ///
-    /// [`unescaped_value()`]: Self::unescaped_value
+    /// [`unescape_value()`]: Self::unescape_value
```
dralley, line 74: To be honest, I think it reads a little worse, even if it's more consistent.  I don't care enough to argue any more (lol) but I'll just say it for what it's worth

* `.value` / `raw_value()`  ->  ... (nothing)
* `unescaped_value()` -> `unescape_value()
* `normalized_value()` -> `normalize_value()`

The former fields weird when compared to the latter
* `

dralley: Also, name conflict needs to be resolved https://github.com/tafia/quick-xml/blob/master/src/events/mod.rs#L738-L757=

Mingun: Just rename private function to something, I'll remove it in my PR tonight.

Mingun:
Thanks for you hard work!

PR #416, opened 2022-07-10
Add `<event>::borrow()` and some minor stuff

Mingun:
This PR makes some refactoring that I do on my path to the implementing namespaces support.

Currently I need to fix lifetime issues and I plan split reader and writer events into separate structs, which requires some preparing work. In order to not make that PR too big I made this PR with this work.

I fixed some intradoc links, move tests to appropriate place and introduced two improvements:
- `BytesStart` and `BytesEnd` structs now implements `From<QName>` that allows to create them easely (useful for writing)
- `Writer::write_event` now consumes event using an `Into` trait. `AsRef` trait, used before, does not allow to accept two unrelated structs which `reader::Event` and `writer::Event` will be. On the contrary, `writer::Event` will implement `From<reader::Event>` which would allow to write reader events as before. In order to not lose the ownership, the `Event::borrow()` method is introduced, as a counterpart to the `Event::into_owned()` method, so `Writer::write_event` could consume only borrowed event

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,8 @@
 - [#395]: Add support for XML Schema `xs:list`
 - [#324]: `Reader::from_str` / `Deserializer::from_str` / `from_str` now ignore
   the XML declared encoding and always use UTF-8
+- [#416]: Add `borrow()` methods in all event structs which allows to get
+  a borrowed version of any event
 
 ### Bug Fixes
 
@@ -108,6 +110,9 @@
   |`read_to_end_unbuffered` |`read_to_end`
 - [#412]: Change `read_to_end*` and `read_text_into` to accept `QName` instead of `AsRef<[u8]>`
 
+- [#416]: `BytesStart::to_borrowed` renamed to `BytesStart::borrow`, the same method
+  added to all events
+
 ### New Tests
 
 - [#9]: Added tests for incorrect nested tags in input
@@ -131,6 +136,7 @@
 [#403]: https://github.com/tafia/quick-xml/pull/403
 [#407]: https://github.com/tafia/quick-xml/pull/407
 [#412]: https://github.com/tafia/quick-xml/pull/412
+[#416]: https://github.com/tafia/quick-xml/pull/416
 
 ## 0.23.0 -- 2022-05-08
 
--- a/compare/benches/bench.rs
+++ b/compare/benches/bench.rs
@@ -25,7 +25,10 @@ fn low_level_comparison(c: &mut Criterion) {
                 }
                 buf.clear();
             }
-            assert_eq!(count, 1550, "Overall tag count in ./tests/documents/sample_rss.xml");
+            assert_eq!(
+                count, 1550,
+                "Overall tag count in ./tests/documents/sample_rss.xml"
+            );
         })
     });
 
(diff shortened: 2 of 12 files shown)


dralley:
> I plan split reader and writer events into separate structs

Why?

Mingun:
To solve lifetime issues, for example #332. Currently reader events can hypothetically contain an owned data, but in reality they never have. That, however, has negative implications on lifetimes.

I'll plan to make a reader event to always borrow (from buffer for buffered reader or from input for a borrowing reader). Reader event will implement an `Into` to the writer event, which is the current `events::Event`, so you will able to make transcoding as before.

You can wait with your opinion until I'll finish that work and make a PR for it

src/de/mod.rs
```diff
@@ -1106,27 +1097,24 @@ mod tests {
             assert_eq!(
                 de.read,
                 vec![
-                    Start(BytesStart::borrowed_name(b"inner")),
+                    Start(QName(b"inner").into()),
                     Text(BytesText::from_escaped_str("text")),
```
dralley, line 1101: I'm mildly negative on this because it makes using the API less consistent.  That's not to say that the implementation shouldn't be there, because it makes sense to have the implementation, but in practice it flows awkwardly.

Not a blocker, just FWIW

src/de/mod.rs
```diff
@@ -1054,26 +1054,20 @@ mod tests {
...
             assert_eq!(
                 de.write,
                 vec![
-                    Start(BytesStart::borrowed_name(b"inner")),
+                    Start(QName(b"inner").into()),
                     Text(BytesText::from_escaped_str("text")),
-                    Start(BytesStart::borrowed_name(b"inner")),
-                    End(BytesEnd::borrowed(b"inner")),
-                    End(BytesEnd::borrowed(b"inner")),
+                    Start(QName(b"inner").into()),
```
dralley, line 1068: I'm mildly negative about this because it makes using the API less consistent. That's not to say that the implementation shouldn't be there, because it makes sense to have the implementation, but in practice when used like this it flows more awkwardly when some tags use `QName` and others use `BytesText::from_*`, `BytesCdata::from_*`, etc.

Not a blocker, just FWIW

dralley:
@Mingun Would you mind splitting the 7 "less interesting" commits off into a different PR, keeping 7280390e4ca and b55aac61f here, and then then continuing your work on this one?  

I feel like it would be useful to evaluate them alongside the additional changes you have planned.

src/de/mod.rs
Mingun: I've made this change in order to reduce diff in one of following commits, that splits common `Event` into `reader::Event` and `writer::Event`. `reader::Event` will not have explicit public methods to construct a value (this is unnecessary for the user, if user need an event he could use `writer::Event`, reader events is read-only), so it will impossible to construct it in the doctests. I can move that change there

Mingun:
I think I'll just strip out controversial commits from this PR -- it is simpler.

dralley (approved):

SergeyKasmy:
Hey. Has there been any progress on splitting events into separate `ReadEvent` and `WriteEvent`? I've stumbled upon the need for them while writing my own (de)serializer and I'd like to try to take the task of doing that. Is that okay?

Mingun:
Actually I have a branch with this split since those days and I think, we should move in that direction. Many XML libraries that I've seen, have this distinction. I plan to make a release of 0.31 soon and make a PR for that in 0.32 and I hope it will be merged.

PR #419, opened 2022-07-15
Fix benchmark result saving w/ Criterion

dralley:
libtest and Criterion don't get along so well.
https://github.com/rust-lang/rust/issues/47241

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,15 @@ regex = "1"
 serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
 
+[lib]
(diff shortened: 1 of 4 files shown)


Mingun (approved):
I did not really understand why the fix is exactly this, but if it works for you then fine. Can merge

dralley:
Yes, all that works fine.  The problem is when you try to pass command line arguments to the benchmark binary directly, e.g.

`cargo bench -- --save-baseline old`.  Somehow libtest gets ahold of the command line arguments first and rejects them unless everything is configured exactly right.

PR #420, opened 2022-07-15
[work-in-progress]: Perform all decoding internally

dralley:
- [ ] The data being streamed from the reader would be converted from whichever encoding it is currently in to UTF-8 when copied to a buffer, or else validated as UTF-8 if that is the appropriate encoding, using streaming APIs
- [ ] UTF-8 would be the only encoding used within the library itself
- [x] All decoding functions would be removed from the API
- [ ] quick-xml would never have to allocate during parsing except from resizing a buffer or doing (un)escaping
- [ ] All API functions would be changed to return `&str` / `String` / `Cow<str>` rather than raw bytes
- [ ] Known character boundaries can be used to ensure the safety of `unsafe fn std::str::from_utf8_unchecked()`, which would not incur any overhead

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -103,6 +103,9 @@
 
 - [#416]: `BytesStart::to_borrowed` renamed to `BytesStart::borrow`, the same method
   added to all events
+- [#420]: `BytesText::from_escaped_str()` and `BytesText::from_plain_str()` renamed to
+  `BytesText::from_escaped()` and `BytesText::from_plain()`, respectively. The previous functions
+  with those names (which operated over `&[u8]`) have been removed.
 
 - [#421]: Removed ability to deserialize byte arrays from serde deserializer.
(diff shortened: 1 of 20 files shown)


dralley:
@Mingun seriously I need some feedback on the plan.

Mingun:
I'm going to sleep now, I'll give my thoughts later. General note: I prefer moving by small steps, first preparing the necessary infrastructure. To begin with, I suggest looking at #421 first

dralley:
I would be fine with merging the commits incrementally.

dralley:
@Mingun These 5 commits are "ready", if you would like to review them as-is.  But it is up to you whether you would like to merge it now, or as part of the complete patch series.

dralley:
I can split this apart better, one sec...

dralley:
Changing approach, working on top of the first half of the async PR.

PR #422, opened 2022-07-17
Finish updating escape functions to use &str and update BytesText constructors

dralley:


Mingun (commented):
I think, that in the current state this PR changes too much at once (it wouldn't be that bad if it was divided into separate commits that change one thing). So I suggest to rebase it over #423, which is focused only on escape functions.

Also, I was thinking about another way to get away from `Decoder` in method parameters, namely, about saving `Decoder` inside an event. With that approach we can change code more granularly without necessary to comment out not-yet ported code and made transition more smoothly

PR #423, opened 2022-07-17
Enforce escape functions to operate over UTF-8

Mingun:
They are always operate over UTF-8 input, but that was not reflected in types.

This set of changes is more isolated that one in #422, and it also does not enforce owning of data in some cases where that is not required.

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -681,15 +681,6 @@ impl<'a> BytesText<'a> {
         }
     }
 
-    /// Creates a new `BytesText` from a byte sequence. The byte sequence is
-    /// expected not to be escaped.
-    #[inline]
-    pub fn from_plain(content: &'a [u8]) -> Self {
-        Self {
-            content: escape(content),
-        }
-    }
-
     /// Creates a new `BytesText` from an escaped string.
     #[inline]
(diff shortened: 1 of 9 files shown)


src/events/mod.rs
```diff
@@ -901,11 +897,13 @@ impl<'a> BytesCData<'a> {
     /// | `&`       | `&amp;`
     /// | `'`       | `&apos;`
     /// | `"`       | `&quot;`
-    pub fn escape(self) -> BytesText<'a> {
-        BytesText::from_escaped(match escape(&self.content) {
+    pub fn escape(self, decoder: Decoder) -> Result<BytesText<'a>> {
```
Mingun, line 900: Because `BytesText::escape` and `BytesText::partial_escape` introduced only in new 0.24 release, the changes in their signatures are not reflected in the changelog. Moreover, the next PRs will change that signature back

dralley: +1

adrian17:
FYI, pretty sure that https://github.com/tafia/quick-xml/pull/423/commits/1a1f4bfea6790c04e008a7ab2ee32147884d0f97 doesn't work, as equality comparison on Cow ignores the variant. You probably need to use `matches` instead?

Mingun:
Yes, you're right. As far as I remember, there are a few more places in tests where we rely on such behavior

PR #424, opened 2022-07-17
Changed BytesText constructors to use utf8 types

dralley:


tests/unit_tests.rs
```diff
@@ -615,9 +615,9 @@ fn test_read_write_roundtrip_escape() -> Result<()> {
         match reader.read_event_into(&mut buf)? {
             Eof => break,
             Text(e) => {
-                let t = e.escape();
+                let t = reader.decoder().decode(e.escape())?;
```
dralley, line 618: Logically the order of operations here is wrong, but as the data is already stored escaped, `.escape()` is a no-op, thus it doesn't really matter - and this will be cleaned up soon anyway.

PR #428, opened 2022-07-18
Change all event and Attributes constructors to accept strings

Mingun:
This PR is extended version of #424. It changes all [events'](https://docs.rs/quick-xml/latest/quick_xml/events/index.html#structs) constructors to accept `&str` / `String` and removes all public constructors that accepts byte arrays. Internally these constructors are still used, so all tests still passed. On now we have our hands untied in changing the internal representation of events (well, almost, `BytesStart::set_name` still accepts `&[u8]`).

I did not include name changes in this PR, because I want to make more radical renames instead of just removing `_str` suffix. I'll made corresponding PR tomorrow.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -150,6 +150,8 @@
   but since now escaping works on strings. Use `BytesText::from_plain_str` instead
 
 - [#428]: Removed `BytesText::escaped()`. Use `.as_ref()` provided by `Deref` impl instead.
+- [#428]: Removed `BytesText::from_escaped()`. Use constructors from strings instead,
+  because writer anyway works in UTF-8 only
 
 ### New Tests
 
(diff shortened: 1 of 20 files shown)


dralley:
Approved: but please allow https://github.com/tafia/quick-xml/pull/425 to be merged first

Mingun:
Actually, I would to merge #425 as late as possible, because I want to play with the possible improvements next week. I see that here we have a consensus, so it seems that this could be merged sooner

dralley:
I would prefer for it to be merged as soon as possible, for the same reason : /  Everything which touches those files is effectively blocked on that PR, and the alternative is forcing 999eagle to rebase it over and over and over again, which doesn't feel like a great way to encourage further contributions.

dralley:
Well, the crux of the issue is the file split (which makes practically every change a breaking one), and your namespace PR includes it, so for the sake of making at least some forwards progress on this issue, we can go with that one first....

dralley (approved):

Issue #153, opened 2019-04-09
borrowed_name should be unsafe


max-ym:
Function `borrowed_name` (in quick_xml::events::BytesStart) as is stated in documentation - can be used to create BytesStart from a given name. But if you pass (by mistake or intentionally) attribute data in the string alongside with the name this function does not parse those attributes nor verify if actually passed data is a valid name. It will treat any kind of data as a name even if it is not valid! This can lead to hard to find bugs as it is possible to confuse `fn borrowed_name` with `fn borrowed` which seem to do similar things but lead to different internal state of BytesStart.

As this function accepts invalid data it should be marked unsafe. The documentation must be more specific about how to use this function and how not. For example, it must be stated that this function does not verify passed data and if user intents to pass attributes too he/she should use `fn borrowed`.

dralley:
`borrowed_name()` doesn't exist anymore (in the current unreleased master branch)

Mingun:
I don't think that this is done. Yes, the `borrowed_name` identifier does not exist anymore, but we still have a functions, that should be unsafe in order to not break invariants, like
https://github.com/tafia/quick-xml/blob/ad57bc29e8d9b10309748ce362fffc3580322cf5/src/events/mod.rs#L158-L171

PR #431, opened 2022-07-19
Use consistent naming for event constructors

Mingun:
Continuation of #428 (includes it changes) that unifies all event constructor names. I'm not sure about `BytesText::new`, however. While this name is very aligned and consistent with other names, I afraid, it can be error-prone. Would it be better to use `new` or leave `from_escaped`?

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -845,7 +845,7 @@ pub struct BytesCData<'a> {
 impl<'a> BytesCData<'a> {
     /// Creates a new `BytesCData` from a byte sequence.
     #[inline]
-    pub fn new<C: Into<Cow<'a, [u8]>>>(content: C) -> Self {
+    pub(crate) fn wrap<C: Into<Cow<'a, [u8]>>>(content: C) -> Self {
         Self {
             content: content.into(),
         }
@@ -854,7 +854,7 @@ impl<'a> BytesCData<'a> {
     /// Creates a new `BytesCData` from a string
     #[inline]
     pub fn from_str(content: &'a str) -> Self {
-        Self::new(content.as_bytes())
+        Self::wrap(content.as_bytes())
     }
 
     /// Ensures that all data is owned to extend the object's lifetime if
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -606,18 +606,8 @@ impl<'a> BytesEnd<'a> {
 
     /// Creates a new `BytesEnd` borrowing a slice
     #[inline]
-    pub fn borrowed(name: &'a str) -> BytesEnd<'a> {
-        BytesEnd {
-            name: Cow::Borrowed(name.as_bytes()),
-        }
-    }
-
-    /// Creates a new `BytesEnd` owning its name
-    #[inline]
-    pub fn owned(name: String) -> BytesEnd<'static> {
-        BytesEnd {
-            name: Cow::Owned(name.into_bytes()),
-        }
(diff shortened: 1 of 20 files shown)


dralley:
> Would it be better to use new or leave from_escaped?

I think it would be OK to leave `from_escaped()` or `new_from_escaped()`, but we should make it more clear in which circumstances it is safe to use.  Instead of saying

> Creates a new `BytesText` from a string. The string is expected not to be escaped.

We should say something like

> Creates a new `BytesText` without escaping the input.  

> Warning: This is to be used only when the input is pre-escaped, or known to not contain any characters requiring escape sequences (e.g. `<`, `>`, `&`).

src/events/mod.rs
```diff
@@ -683,27 +683,15 @@ impl<'a> BytesText<'a> {
...
     #[inline]
-    pub fn from_plain_str(content: &'a str) -> Self {
-        Self {
-            content: match escape(content) {
-                Cow::Borrowed(s) => Cow::Borrowed(s.as_bytes()),
-                Cow::Owned(s) => Cow::Owned(s.into_bytes()),
-            },
-            decoder: Decoder::utf8(),
-        }
+    pub fn from_unescaped(content: &'a str) -> Self {
```
dralley, line 693: This is backwards from what I think would be most intuitive.  The default should be to escape the input in `new()`, with an option to provide escaped input if you already have it (or know that it cannot be escaped)`.

Mingun: Yes, I had the same doubts. I've changed names: `from_escaped` and `new`

src/events/mod.rs
```diff
@@ -138,52 +138,33 @@ impl<'a> BytesStart<'a> {
...
-    /// `name` is not checked to be a valid name
-    #[inline]
-    pub fn borrowed_name(name: &'a str) -> BytesStart<'a> {
-        Self::borrowed(name, name.len())
-    }
-
-    /// Creates a new `BytesStart` from the given content (name + attributes)
-    ///
-    /// Owns its contents.
+    /// `&content[..name_len]` is not checked to be a valid name
```
dralley, line 159: The attributes are also not checked to be valid.

dralley: To be honest I'm not sure I see the use case for this function for the majority of users, ironically the one use which is clear is simplifying integration tests by making it easy to create events from invalid data.

I'm not saying that we should remove it, but perhaps it can be discouraged more actively?

Mingun: Actually, we should mark that functions `unsafe` (#153) and provide a safe versions returning `Result` and which will check for a validity. But first we need to implement that validity checks (#371 and others) :). So I think it is a task for another PR, probably the one that will add these checks

dralley: Sure, sounds fine.  Let's still do the documentation part though.

Mingun: Could you write a suggestion?

dralley: How about 

> `&content[..name_len]` must be a valid name, and the remainder of `content` must be valid attributes. Neither are checked, it is possible to generate invalid XML if `content` or `name_len` are incorrect.

Mingun: Done + added similar warnings to all other constructors where there were none

dralley:
Approved: but please allow https://github.com/tafia/quick-xml/pull/425 to be merged first

dralley:
Alright, well, go ahead and fix up the conflicts here and on the other PR.  Might as well go ahead and merge them.

dralley (approved):

PR #434, opened 2022-07-21
Remove excessive serde tests and fix a bug

Mingun:
This PR removes tests which have better alternatives in `seq` submodule since #387.

Also, while working on namespaces support, I accidently found a bug in a deserializer, so I've added a test and a fix for it.

### DIFF
--- a/tests/serde-de.rs
+++ b/tests/serde-de.rs
@@ -110,48 +110,6 @@ fn nested_collection() {
     );
 }
 
-#[test]
-fn collection_of_enums() {
-    #[derive(Debug, Deserialize, PartialEq)]
-    enum MyEnum {
-        A(String),
-        B { name: String, flag: bool },
-        C,
-    }
-
-    #[derive(Debug, Deserialize, PartialEq)]
-    struct MyEnums {
-        // TODO: This should be #[serde(flatten)], but right now serde don't support flattening of sequences
-        // See https://github.com/serde-rs/serde/issues/1905
-        #[serde(rename = "$value")]
-        items: Vec<MyEnum>,
-    }
-
-    let s = r#"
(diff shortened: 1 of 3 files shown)


tests/serde-de.rs
```diff
@@ -110,48 +110,6 @@ fn nested_collection() {
...
-        C,
-    }
-
-    #[derive(Debug, Deserialize, PartialEq)]
-    struct MyEnums {
-        // TODO: This should be #[serde(flatten)], but right now serde don't support flattening of sequences
-        // See https://github.com/serde-rs/serde/issues/1905
-        #[serde(rename = "$value")]
-        items: Vec<MyEnum>,
-    }
```
dralley, line 128: This is a much more complex test than the one that replaces it (https://github.com/tafia/quick-xml/blob/d34b77972ccaad505c3348761382fe7d01ca49ce/tests/serde-de.rs#L1261=), is it really a full replacement?

Mingun: Actually, yes. The mapping should be independent and composable, so we shouldn't bother of internals of the `MyEnum` type (or `Item` type in `Project`). For now we don't have a tests that ensures that mapping is really additive in all cases (which is why #435 is happened), but
- that check should be added explicitly and for all combinations
- in that test mapping is independent and composable

I'll plan to add necessary tests that would ensure that mapping is always composable (may be in form of doctests in #369)

dralley (approved):

dralley:
I used the merge conflict cleanup feature just to try it out - you can squash that commit if you want.

Mingun:
The history becomes too confusing because of the back-and-forth merges. I'd rather do a rebase.

PR #436, opened 2022-07-23
Remove the ability to use SliceReader with raw bytes

dralley:
In the near future, decoding will be performed automatically as the input is read. If the input has an unknown encoding, it must be decoded first, necessitating a buffer. Therefore only the buffered implementation can be used for `Reader::from_bytes()`

If the encoding of the bytes is known up-front, you can decode them up-front and subsequently use `Reader::from_str()` for borrowing behavior if desired.

Made some utilities such as `detect_encoding()`, `decode()`, and `decode_with_bom_removal()` available as standalone functions so that they can be used on user-provided data.


dralley:
Pending #425

src/reader/buffered_reader.rs
```diff
@@ -652,17 +660,63 @@ mod test {
...
     }
 
     #[allow(dead_code)]
-    fn reader_from_bytes(s: &[u8]) -> Reader<BufferedReader<&[u8]>> {
-        Reader::from_reader_internal(BufferedReader(s))
+    fn reader_from_bytes(bytes: &[u8]) -> Reader<BufferedReader<&[u8]>> {
+        Reader::from_reader_internal(BufferedReader(bytes))
+    }
+
+    #[cfg(feature = "encoding")]
```
dralley, line 676: Moving these because it doesn't make sense to check these for both when `SliceReader` can't work with encoded bytes.

PR #437, opened 2022-07-23
Split out namespace aware reader to a struct that manage the internal namespaces buffer

Mingun:
This is PR for low-level namespaces support in the quick-xml, on which high-level serde support will be implemented.

The goals of this PR is to make a safe interface and leave the option of not paying for namespace support if it is not needed. For that the buffer with readed namespaces now managed by the dedicated `NsReader` reader, which should be used instead of `Reader` when you need namespaces. Because now buffer is stored inside the reader, the user cannot accidently clear it and get a panic or wrong results later, when it would try to resolve namespace. In addition, it is impossible now to mix read methods with and without namespace awareness, which could lead to hard-to-catch bugs.

Originally, I've planned to include the serde part into this PR, but number of commits for support serde grows and I've decided to made a dedicated PR for that. At least, the serde part slightly depends on #434, because those changes will break some removed tests.

-----
I would like to merge that before #425, because:
- it includes the first commit from #425
- it makes unnecessary the second commit from #425
- I want to try to improve #425 next week


src/reader/mod.rs
```diff
@@ -77,6 +77,19 @@ macro_rules! builder_methods {
 
         /// Changes whether mismatched closing tag names should be detected.
         ///
+        /// Note, that start and end tags [should match literally][spec], they cannot
+        /// have different prefixes even if both prefixes resolves to the same namespace.
+        /// The
```
dralley, line 82: Should probably be "The XML ... is not valid, even though sematically the start tag is the same as the end tag."

"The reason is that namespaces _are_..."

src/reader/ns_reader.rs
```diff
@@ -121,55 +121,121 @@ impl<R> NsReader<R> {
 
 /// Read methods
 impl<R: BufRead> NsReader<R> {
-    /// Reads the next event and resolves its namespace (if applicable).
+    /// Reads the next event into given buffer.
+    ///
+    /// This method manages namespaces but not resolves them automatically.
```
dralley, line 285: but doesn't resolve them automatically

src/reader/ns_reader.rs
```diff
@@ -0,0 +1,739 @@
...
+    /// [`BytesStart::to_end()`]: crate::events::BytesStart::to_end
+    #[inline]
+    pub fn read_to_end(&mut self, end: QName) -> Result<()> {
+        // According to the https://www.w3.org/TR/xml11/#dt-etag, end name should
+        // match literally the start name. See `Self::check_end_names` documentation
+        self.reader.read_to_end(end)
+    }
+}
+
+impl<R> From<Reader<R>> for NsReader<R> {
```
dralley, line 720: If the user were to create a normal reader, read some events, then call `NsReader::from(reader)` as this would allow, is that going to result in incorrect internal state?

src/reader/ns_reader.rs
```diff
@@ -189,6 +255,121 @@ impl<'i> NsReader<&'i [u8]> {
...
+    /// assert_eq!(count, 3);
+    /// assert_eq!(txt, vec!["Test".to_string(), "Test 2".to_string()]);
+    /// ```
+    ///
+    /// [`Start`]: Event::Start
+    /// [`Empty`]: Event::Empty
+    /// [`End`]: Event::End
+    /// [`read_event()`]: Self::read_event
+    #[inline]
+    pub fn read_event_ns(&mut self) -> Result<(ResolveResult, Event<'i>)> {
```
dralley, line 632: I'm not sure this naming is an improvment, it feels too brief.  Perhaps `read_event_with_ns()` would be better.

src/reader/mod.rs
```diff
@@ -11,6 +11,109 @@ use crate::events::{BytesCData, BytesDecl, BytesEnd, BytesStart, BytesText, Even
 
 use memchr;
 
+macro_rules! builder_methods {
```
dralley, line 14: I don't see why this needs to be a macro, and the commit message doesn't provide any wisdom.  Probably should be separate commits...

src/reader/ns_reader.rs
Mingun: Yes, you are right. It is better to remove it

src/reader/ns_reader.rs
Mingun: I agree, not a better name. May be `read_resolved_event`?

src/reader/mod.rs
```diff
@@ -1,18 +1,139 @@
...
 #[cfg(feature = "encoding")]
 use encoding_rs::{Encoding, UTF_16BE, UTF_16LE, UTF_8};
 
 use crate::errors::{Error, Result};
 use crate::events::{BytesCData, BytesDecl, BytesEnd, BytesStart, BytesText, Event};
-use crate::name::{LocalName, NamespaceResolver, QName, ResolveResult};
 
 use memchr;
 
+macro_rules! builder_methods {
```
Mingun, line 14: @dralley , you've asked, why I use macro here. Although it looks like you have deleted it, I will still answer.

The change in which this macro is introduced, removes `impl DerefMut<Target=Reader> for NsReader`. That means, that all `Reader`'s methods that accept `&mut self` won't be available for the `NsReader`, including the ones that implemented by this macro.

Because we need a way to configure our reader, we need to reimplement necessary methods that accept `&mut self`. In order to not duplicate the code and the doc of this configure methods, I've put them into the macro and embed to both readers.

dralley: Yes, I figured that part out after a few minutes.

src/reader/ns_reader.rs
Mingun: Changed to `read_resolved_event` / `read_resolved_event_into`

src/reader/mod.rs
Mingun: Added that explanation to the commit message

dralley (approved):
I think the PR still needs to be pushed, changes aren't live yet.  `read_resolved_event()` still doesn't feel like the best name - but I have no suggestions that feel clearly better.

src/reader/ns_reader.rs
Mingun: Removed (replaced by private `new` method)

src/reader/mod.rs
```diff
@@ -77,6 +77,19 @@ macro_rules! configure_methods {
 
         /// Changes whether mismatched closing tag names should be detected.
         ///
+        /// Note, that start and end tags [should match literally][spec], they cannot
+        /// have different prefixes even if both prefixes resolves to the same namespace.
+        /// The
```
dralley, line 82: This should be either "The XML" or "This XML" or "This:"

dralley: Also "both prefixes resolves" -> "both prefixes resolve"

Mingun: Done ("The XML")!

PR #439, opened 2022-07-24
Encoding module

dralley:
Move the core functionality for encoding / decoding to a new module and provide some freestanding utilities for decoding buffers.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -1,0 +1,172 @@
+//! A module for wrappers that encode / decode data.
+
+use std::borrow::Cow;
+
+#[cfg(feature = "encoding")]
+use encoding_rs::{Encoding, UTF_16BE, UTF_16LE, UTF_8};
+
+use crate::{Error, Result};
+
+/// Decoder of byte slices to the strings. This is lightweight object that can be copied.
+///
+/// If feature `encoding` is enabled, this encoding taken from the `"encoding"`
+/// XML declaration or assumes UTF-8, if XML has no <?xml ?> declaration, encoding
+/// key is not defined or contains unknown encoding.
+///
+/// The library supports any UTF-8 compatible encodings that crate `encoding_rs`
+/// is supported. [*UTF-16 is not supported at the present*][utf16].
+///
+/// If feature `encoding` is disabled, the decoder is always UTF-8 decoder:
+/// any XML declarations are ignored.
+///
+/// [utf16]: https://github.com/tafia/quick-xml/issues/158
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Decoder {
+    #[cfg(feature = "encoding")]
+    pub(crate) encoding: &'static Encoding,
+}
+
+impl Decoder {
+    pub(crate) fn utf8() -> Self {
+        Decoder {
+            #[cfg(feature = "encoding")]
+            encoding: UTF_8,
+        }
+    }
+
+    #[cfg(all(test, feature = "encoding", feature = "serialize"))]
+    pub(crate) fn utf16() -> Self {
+        Decoder { encoding: UTF_16LE }
+    }
+}
+
+#[cfg(not(feature = "encoding"))]
+impl Decoder {
+    /// Decodes a UTF8 slice regardless of XML declaration and ignoring BOM if
+    /// it is present in the `bytes`.
+    ///
+    /// Returns an error in case of malformed sequences in the `bytes`.
+    ///
+    /// If you instead want to use XML declared encoding, use the `encoding` feature
+    #[inline]
+    pub fn decode<'b>(&self, bytes: &'b [u8]) -> Result<Cow<'b, str>> {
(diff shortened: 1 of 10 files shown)


src/encoding.rs
```diff
@@ -60,7 +60,7 @@ impl Decoder {
     ///
     /// If you instead want to use XML declared encoding, use the `encoding` feature
     pub fn decode_with_bom_removal<'b>(&self, bytes: &'b [u8]) -> Result<Cow<'b, str>> {
-        let bytes = if bytes.starts_with(b"\xEF\xBB\xBF") {
+        let bytes = if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
```
dralley, line 63: Aesthetic preference, I feel like this is easier to read and better communicates the non-text-ness.

src/encoding.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+#[cfg(feature = "encoding")]
+fn remove_bom<'b>(bytes: &'b [u8], encoding: &'static Encoding) -> &'b [u8] {
+    let (_, bytes) = split_at_bom(bytes, encoding);
+    bytes
+}
+
+/// Automatic encoding detection of XML files based using the [recommended algorithm]
+/// (https://www.w3.org/TR/xml11/#sec-guessing)
```
Mingun, line 151: It seems that markdown don't allow to break `](` sequence, rustdoc complains:
```
> cargo doc --all-features       
 Documenting quick-xml v0.23.0 (D:\Projects\quick-xml)
warning: this URL is not a hyperlink
   --> src\encoding.rs:151:6
    |
151 | /// (https://www.w3.org/TR/xml11/#sec-guessing)
    |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<https://www.w3.org/TR/xml11/#sec-guessing>`
    |
    = note: `#[warn(rustdoc::bare_urls)]` on by default
    = note: bare URLs are not automatically turned into clickable links
```

src/encoding.rs
```diff
@@ -0,0 +1,197 @@
...
+///
+/// The algorithm suggests examine up to the first 4 bytes to determine encoding
+/// according to the following table:
+///
+/// | Bytes       |Detected encoding
+/// |-------------|------------------------------------------
+/// |`00 00 FE FF`|UCS-4, big-endian machine (1234 order)
```
Mingun, line 158: Because now the function is public, maybe more explicitly mark rows, where autodetection is supported?

src/encoding.rs
```diff
@@ -0,0 +1,197 @@
...
+        // without BOM
+        _ if bytes.starts_with(&[0x00, b'<', 0x00, b'?']) => Some(UTF_16BE), // Some BE encoding, for example, UTF-16 or ISO-10646-UCS-2
+        _ if bytes.starts_with(&[b'<', 0x00, b'?', 0x00]) => Some(UTF_16LE), // Some LE encoding, for example, UTF-16 or ISO-10646-UCS-2
+        _ if bytes.starts_with(&[b'<', b'?', b'x', b'm']) => Some(UTF_8), // Some ASCII compatible
+
+        _ => None,
+    }
+}
+
+// TODO: add some tests for functions
```
Mingun, line 197: Don't mind to add tests before merge?

src/encoding.rs
```diff
@@ -0,0 +1,197 @@
...
+    if let Some(encoding) = detect_encoding(bytes) {
+        let bytes = remove_bom(bytes, encoding);
+        decode(bytes, encoding)
+    } else {
+        decode(bytes, UTF_8)
+    }
+}
+
+#[cfg(feature = "encoding")]
+fn split_at_bom<'b>(bytes: &'b [u8], encoding: &'static Encoding) -> (&'b [u8], &'b [u8]) {
```
Mingun, line 132: This seems redundant, because first part is not used anywhere. Why just cut off the beginning, as was before, is not enough?

Mingun (approved):
Approved, fix the markdown link

src/encoding.rs
dralley: I've just removed them entirely.  It is unlikely that support for those encodings will ever be available, at least using this library.

src/encoding.rs
dralley: I considered it but I figured it will be easier to do a big testing push at the end.  We've only got a few sample documents and entering the data manually would be painful.

dralley: Of course, it needs to happen eventually - it would just be helpful to have the full picture in mind while doing that work.

src/encoding.rs
dralley: You're right, but I was thinking we may want to add some variants that return the BOM the same way that we provide `StartText`.  Or at the very least that it would be useful for testing.

On the other hand, I kinda feel like both `StartText` and returning the BOM have limited utility in practice.  But it feels like an open question.

I'll leave it as-is for now but I wouldn't be be upset if we end up removing it later.

PR #440, opened 2022-07-24
Remove the ability to use un-buffered functions with Reader<&[u8]>

dralley:


src/de/mod.rs
```diff
@@ -301,7 +301,7 @@ pub fn from_slice<'de, T>(s: &'de [u8]) -> Result<T, DeError>
 where
     T: Deserialize<'de>,
 {
-    let mut de = Deserializer::from_slice(s);
+    let mut de = Deserializer::from_reader(s);
```
Mingun, line 304: Just remove this function completely

src/reader/ns_reader.rs
```diff
@@ -525,7 +525,7 @@ impl<'i> NsReader<&'i [u8]> {
     /// Creates an XML reader from a slice of bytes.
     #[inline]
     pub fn from_bytes(bytes: &'i [u8]) -> Self {
-        Self::new(Reader::from_bytes(bytes))
+        Self::new(Reader::from_reader(bytes))
```
Mingun, line 528: Just remove this function completely

Mingun (approved):
Approved, only remove mentioned functions -- their name assumes that thay can borrow, but that's not true anymore, so I think there's no need to keep them. We have a `from_reader` for that

src/de/mod.rs
```diff
@@ -296,15 +296,6 @@ where
...
 
-/// Deserialize an instance of type `T` from bytes of XML text.
-pub fn from_slice<'de, T>(s: &'de [u8]) -> Result<T, DeError>
-where
-    T: Deserialize<'de>,
-{
-    let mut de = Deserializer::from_slice(s);
-    T::deserialize(&mut de)
-}
-
```
Mingun, line 307: Still, the constructors were probably deleted in vain. What do they interfere with? It's still may be valuable to pass a byte array and allow quick_xml to detect encoding and parse it. If encoding is UTF-8, then borrowing is possible. @dralley, what do you think?

If it remains removed, then it is needed to fix dangling documentation links

dralley: re: documention, whoops, yes, I can fix that

re: constructors, IMO it is unnecessary, this is why we have the decoding functions as separate now.  It's only an extra line or two of code and if we made a constructor which included that line it would need to be falliable and return Result, unlike the others.

dralley: We should make the CI fail on documentation issues if possible ^^

Issue #360, opened 2022-02-17
Send events through channel


derolf:
Look at this broken piece of code:

```rust
  let mut xml = Reader::from_reader(...);
  let (tx, rx) = mpsc::channel::<Event>();
  let mut buf = Vec::new();
  
  loop {
      let ev = xml.read_event(&mut buf);
      tx.send(ev.unwrap());
  }
}
```

I'd like to decode the XML in one thread and send the events via a channel. Do you have a minimum example on how to do this?

Mingun:
This code is worked for me:
```rust
/// Regression test for https://github.com/tafia/quick-xml/issues/360
#[test]
fn issue360() {
    let mut r = Reader::from_str("<tag1 attr1='line 1\nline 2'></tag1>");
    let (tx, rx) = std::sync::mpsc::channel::<Event>();

    loop {
        match r.read_event() {
            Ok(Event::Eof) => break,
            ev => tx.send(ev.unwrap()).unwrap(),
        }
    }
    std::thread::spawn(move || {
        // Ok(Start(BytesStart { buf: Borrowed("tag1 attr1='line 10xAline 2'"), name_len: 4 }))
        println!("{:?}", rx.recv());
    }).join().unwrap();
}
```
@derolf, It that is not what you want, please add a comment

PR #441, opened 2022-07-25
[work-in-progress] Decoding BufReader implementation

dralley:
Very, very work-in-progress.  Doesn't compile yet.  Only posting it for transparency.


Mingun:
Did you consider to use [encoding_rs_io](https://lib.rs/crates/encoding_rs_io)?

dralley:
Never came across it, no.  I'll take a look.

dralley:
This library looks nice, there are a few limitations to think about how to work around (or not), but it does enough of the right things that I'll use it for now and we can figure out the rest later.

* getting a raw BOM back from the reader - you can ignore it entirely in which case it will be returned but not used for sniffing, or you can use it for sniffing in which case it will be removed from the stream (because otherwise the output wouldn't be valid)
* manually setting the decoder to a particular encoding variant mid-stream based on the `encoding` attribute. It does its own sniffing at runtime but doesn't provide a way to change it manually after building the decoder.

However I'll repeat that I don't see any great ways to accomplish this without having separate structs as in the other PR.  `Reader` is generic over `R`, and `R` needs to be decoded and buffered, which pushes the burden of doing so onto the user, because `Reader` can't wrap the inner `R` in a decoder or buffer without adding overhead when working with slices.

Mingun:
> `Reader` is generic over `R`, and `R` needs to be decoded and buffered

But after this PR we should perform decoding only for buffered case (because borrowed case will read from `&str`, right?) and we already require the user to provide it's own buffer for event data:
https://github.com/tafia/quick-xml/blob/ad57bc29e8d9b10309748ce362fffc3580322cf5/src/reader/buffered_reader.rs#L62
We should just change the type of that buffer to a `String` and add a small internal buffer for holding bytes if we stop in the middle of a char. Maybe even that buffer already managed by `encoding_rs` crate.

Even if we would allow to read from `&[u8]` using borrowed methods, we could just return an error if we couldn't read an event without allocation (for decoding or unescaping), as serde does in similar cases.

dralley:
>     But after this PR we should perform decoding only for buffered case (because borrowed case will read from &str, right?)

Yes but that's not really the issue I'm describing, basically when you have `Reader<&str>` on one hand but also `Reader<BufReader<DecodeReaderBytes<File>>>` on the other, constructors that are working with `Reader<R>` become a mess, particularly `NsReader`.  It may yet still work, I'm just trying to untangle the mess.

> and we already require the user to provide it's own buffer for event data: We should just change the type of that buffer to a String 

That's the first step, yes, although I'm not sure if it's the ideal end point - the buffer could just as well be internal rather than user-provided.  I'm not sure of the lifetime implications of that yet but it would be nice to just use `read_event()` everywhere, have it borrow from the buffer inside the reader.

> and add a small internal buffer for holding bytes if we stop in the middle of a char. Maybe even that buffer already managed by encoding_rs crate.

It's managed by the `encoding_rs_io` crate, we don't need to be concerned about it if we're using that. `read()` will only ever return UTF-8.

dralley:
@Mingun I have the next 4 days free to work on this.  Question for you. 

How do you want to handle BOMs?  There's no way to get it back from the `DecodeBytesReader` provided by `encoding_rs_io`.  I think that's fine, I don't see a lot of value in providing the raw BOM bytes and it makes internal representations a lot more difficult, but I'm open to counterargument.

There's also no way to see whether a BOM was detected or not, or to manually override the encoding.

Mingun:
> constructors that are working for with `Reader<R>` become a mess because they're no longer creating `Reader<R>` but rather `Reader<BufReader<DecodeReaderBytes<R>>>`.

Hmm, why they should? If you implement constructor for `Reader<BufReader<DecodeReaderBytes<R>>>`, they will not exist for `Reader<&str>`. If generic constructors in `Reader<R>` becomes useless, then make them private:

```rust
impl<R> Reader<R> {
  // former from_reader()
  fn new(reader: R) -> Self {...}
}

impl<'i> Reader<&'i str> {
  pub fn from_str(string: &'i str) -> Self {...}
}

impl<R: BufRead> Reader<BufReader<DecodeReaderBytes<R>>> {
  pub fn from_reader(reader: R) -> Self {...}
}
```

> I'm not sure of the lifetime implications of that yet but it would be nice to just use `read_event()` everywhere, have it borrow from the buffer inside the reader, and avoid forcing the user to manage the buffer by calling `.clear()` manually in the right place.

But @tafia designed that interface specially for to not store buffer inside and allow the end user to manage the buffer in a way that he needs. I think, we should preserve this behavior. This design is still good because in the end the event is not tied to the `Reader` that produced it. Probably that design should help with solving some usage problems, such as #360, which could become painful if we start borrow from the reader.

Mingun:
Regarding `encoding_rs_io` problems -- I see several ways how we could handle them:
1. do not use `encoding_rs_io`, but write our own implementation
2. make a PR(s) in `encoding_rs_io` that would to help to accomplish our goals
3. determine encoding without `encoding_rs_io`, and when we read `Decl` event, switch to `encoding_rs_io` reading

The second choice seems the most promising for me.

dralley:
@Mingun You can review the first 3 commits, still working on the rest.

tests/encodings.rs
```diff
@@ -0,0 +1,87 @@
+use quick_xml::events::Event;
+use quick_xml::Reader;
+
+#[cfg(feature = "encoding")]
```
Mingun, line 4: You should instead of this flag mark these tests as requiring `encoding` feature in `Cargo.toml`, like that is done for `serde-de.rs` tests

Mingun (commented):
I like the idea to create a separate test file for encodings, so the commit 2 is good for me. The others lets hold until your finish your work. I'm not sure, that we should remove `StartText` event, because it still can be useful even if will not contain BOM. The text before the declaration is not considered as a part of XML model, so it should be treated differently, for example, the unescaping of that text has no meaning.

I suggest firstly to add tests for all possible encodings -- because the set of encodings is finite and not so big, we can create a bunch of documents in each encoding and test their parsing. The tests that I would like to see:
- detecting encoding (I found that we do not do that correctly, I'll plan to look at this before release, I already made some investigations and improvements, but not yet proved them with tests)
- reading events

The XMLs should looks like:
```xml
<?xml encoding="<encoding>"?>
<root attribute1=value1
      attribute2="value1"
      attribute3='value2'
>
  <?instruction?>
  <!--Comment-->
  text
  <ns:element ns:attribute="value1" xmlns:ns="namespace"/>
  <[[CDATA[[cdata]]>
</root>
```

Because almost all supported encodings are one-byte encodings, we could create XML files with all possible symbols (probably except that explicitly forbidden by the XML standard). The national symbols should be presented in all possible places (tag names, namespace names, attribute names and values, text and CDATA content, comments, processing instructions).

Only after that, I believe, we should made any improvements in encoding support.

dralley:
> I'm not sure, that we should remove StartText event, because it still can be useful even if will not contain BOM. The text before the declaration is not considered as a part of XML model, so it should be treated differently, for example, the unescaping of that text has no meaning.

Even in that case though, one potential mitigation path would be to ignore such text by default, which would be pretty easy to do as we already treat the first event differently.  And if we decided to make it configurable so that the starting text can remain visible, the user in that case should "know what they're doing" and accept the minor usage quirks in that case.  

I'm not sure it's worth being too concerned about since the number of documents with this problem - and not just whitespace or newlines but significant data - is probably quite small.  I feel like we should have a user asking and making a good case for it before we add the complexity.  Right now, the only complaint has been about the BOM causing trouble.

Does that sound reasonable?

dralley:
>Because almost all supported encodings are one-byte encodings, we could create XML files with all possible symbols (probably except that explicitly forbidden by the XML standard). The national symbols should be presented in all possible places (tag names, namespace names, attribute names and values, text and CDATA content, comments, processing instructions).

>Only after that, I believe, we should made any improvements in encoding support.

We should probably do that before we declare perfect support for all these encodings, but I disagree that this ought to block the ergonomic improvements.

Mostly we need to test encoding *detection*, and whatever custom wrappers for e.g. validation that we add, but the area of interest is more the regions around the end of the buffer, not the various parts of XML.  Because only UTF-8 should ever be parsed in the first place.

dralley:
>    detecting encoding (I found that we do not do that correctly, I'll plan to look at this before release, I already made some investigations and improvements, but not yet proved them with tests)

Details?

Mingun:
> Details?

[Here](https://github.com/tafia/quick-xml/pull/459#discussion_r948175950). I'll make a PR on a weekend

src/de/simple_type.rs
```diff
@@ -493,6 +492,16 @@ impl<'de, 'a> Deref for CowRef<'de, 'a> {
     }
 }
 
+impl<'de, 'a> From<CowRef<'de, 'a>> for Content<'de, 'a> {
```
dralley, line 495: Can one of these two structs be removed, after all of these simplifications?

Mingun: I'm not sure that this possible. `Content::Owned` stores data + offset to unconsumed data whereas `CowRef::Owned` only a content itself. `Input` and `Slice` are always store unconsumed data and updated when data is consumed

Cargo.toml
```diff
@@ -39,7 +39,7 @@ name = "macrobenches"
 harness = false
 
 [features]
-default = []
+default = ["encoding"]
```
dralley, line 43: I don't have very strong feelings about this, good arguments can be made either way.  The rationale is in the commit message.

I did some tests measuring binary size and performance with and without the feature and the differences were not especially large (around 5% for binary size, 10% for performance on my laptop).  That feels acceptable for a friendlier default especially considering those are "without including IO" numbers.

dralley:
Apart from a UTF-16 encoding detection bug I'm still working on, and async support, it's working

src/reader/mod.rs
```diff
@@ -459,73 +460,47 @@ pub struct Reader<R> {
...
+        }
+    }
 }
 
-/// Getters
+/// Public implementation-independent functionality
 impl<R> Reader<R> {
-    /// Consumes `Reader` returning the underlying reader
-    ///
-    /// Can be used to compute line and column of a parsing error position
```
dralley, line 478: I'm thinking we really need a better solution than this.  Either by tracking newlines during parsing or by automatically implementing something like this for `&str` and perhaps `File`.

dralley: I'm not sure about the overhead of the former approach, might be too expensive.

Mingun (commented):
Left some preliminary comments

Changelog.md
```diff
@@ -160,7 +169,6 @@
 - [#416]: `BytesStart::to_borrowed` renamed to `BytesStart::borrow`, the same method
   added to all events
 
-- [#421]: `decode_and_unescape*` methods now does one less allocation if unescaping is not required
```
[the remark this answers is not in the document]
dralley: Yup, leftover from previous.

src/reader/mod.rs
```diff
@@ -1612,7 +1588,7 @@ mod test {
                 /// character should be stripped for consistency
                 #[$test]
                 $($async)? fn bom_from_reader() {
-                    let mut reader = Reader::from_reader("\u{feff}\u{feff}".as_bytes());
+                    let mut reader = Reader::from_str("\u{feff}\u{feff}");
```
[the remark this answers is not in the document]
dralley: Then it needs to be removed, because this variant of this method cannot be implemented for `SliceReader`.

dralley:
And also https://github.com/hsivonen/encoding_rs/issues/88

I'm not sure how long I'm going to be blocked on that, so if there is a portion of these commits you might be willing to merge now, I'd appreciate that.  Otherwise I'll probably work on the attribute normalization PR.

dralley:
So, as a practical matter I now think we need to strip the BOM in all situations, no matter what.

Not because there's anything invalid about it at a purely technical level, but because buried deep down on page 157 out of 1054 of the Unicode standard it says that for UTF-16 the BOM "should not be considered part of the text".  It actually says *nothing* about that for UTF-8 and in fact implies the opposite, but `encoding_rs` will remove the BOM from the stream anyway even for UTF-8.

Realistically I don't think we should cut against the grain of what `encoding_rs` is doing, and we need to get the same output in both cases, so that means the BOM should never be present.

Mingun:
Do what you think right, just explain decisions in the comments in the code

dralley:
@mingun I'm just going to rip out any changes to the serde code and do that absolutely last, so I don't need to keep redoing it

It's been a month without any further progress on getting https://github.com/BurntSushi/encoding_rs_io/pull/11 merged, so I'm just going to vendor the code internally for now.  It's not that much code.

dralley:
@Mingun I will try to pick this back up again in the next month, but it will need to be handled in smaller steps that are merged more incrementally.  I think the initial MVP should be to just have a feature that does utf-8 validation on the input, no decoding.

My question for you is whether that could be made the default behavior if the `encoding` feature is not enabled, or if you'd rather it be opt-in for now.  And whether, long-term, there would still be a need to have a bypass to disable validation and access raw bytes.

PR #449, opened 2022-07-31
Split reader to a `Parser` and a `Reader`

Mingun:
This PR has two goals:
- reduce size of the `reader/mod.rs` file to keep it maintainable
- allow to replate the underlying reader and keep the state. Thus we can convert sync reader to async reader and vice versa if the underlying reader allow this

Also, I include in this PR some minor changes that simplifies async implementation.

_(If merge please do that with merge commit)_

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -142,35 +142,35 @@ pub use ns_reader::NsReader;
 ///   subgraph _
 ///     direction LR
 ///
-///     Init   -- "(no event)"\nStartText                              --> Opened
-///     Opened -- Decl, DocType, PI\nComment, CData\nStart, Empty, End --> Closed
-///     Closed -- "#lt;false#gt;\n(no event)"\nText                    --> Opened
+///     Init      -- "(no event)"\nStartText                              --> OpenedTag
+///     OpenedTag -- Decl, DocType, PI\nComment, CData\nStart, Empty, End --> ClosedTag
+///     ClosedTag -- "#lt;false#gt;\n(no event)"\nText                    --> OpenedTag
 ///   end
-///   Closed -- "#lt;true#gt;"\nStart --> Empty
-///   Empty  -- End                   --> Closed
+///   ClosedTag -- "#lt;true#gt;"\nStart --> Empty
+///   Empty     -- End                   --> ClosedTag
 ///   _ -. Eof .-> Exit
 /// ```
 #[derive(Clone)]
-enum TagState {
+enum ParseState {
     /// Initial state in which reader stay after creation. Transition from that
     /// state could produce a `StartText`, `Decl`, `Comment` or `Start` event.
-    /// The next state is always `Opened`. The reader will never return to this
-    /// state. The event emitted during transition to `Opened` is a `StartEvent`
+    /// The next state is always `OpenedTag`. The reader will never return to this
+    /// state. The event emitted during transition to `OpenedTag` is a `StartEvent`
     /// if the first symbol not `<`, otherwise no event are emitted.
     Init,
     /// State after seeing the `<` symbol. Depending on the next symbol all other
     /// events (except `StartText`) could be generated.
     ///
-    /// After generating ane event the reader moves to the `Closed` state.
-    Opened,
+    /// After generating ane event the reader moves to the `ClosedTag` state.
+    OpenedTag,
     /// State in which reader searches the `<` symbol of a markup. All bytes before
     /// that symbol will be returned in the [`Event::Text`] event. After that
-    /// the reader moves to the `Opened` state.
-    Closed,
+    /// the reader moves to the `OpenedTag` state.
+    ClosedTag,
     /// This state is used only if option `expand_empty_elements` is set to `true`.
-    /// Reader enters to this state when it is in a `Closed` state and emits an
+    /// Reader enters to this state when it is in a `ClosedTag` state and emits an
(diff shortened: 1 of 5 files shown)


src/reader/mod.rs
```diff
@@ -287,7 +287,7 @@ pub struct Reader<R> {
     /// reader
     reader: R,
     /// current buffer position, useful for debugging errors
-    buf_position: usize,
+    offset: usize,
```
dralley: It would be helpful to clarify in the comments whether this is the offset is referring to the user-provided buffer, or the offset since the beginning of the file.

src/reader/slice_reader.rs
```diff
@@ -24,7 +24,7 @@ impl<'a> Reader<&'a [u8]> {
         #[cfg(feature = "encoding")]
         {
             let mut reader = Self::from_reader(s.as_bytes());
-            reader.encoding = EncodingRef::Explicit(UTF_8);
+            reader.parser.encoding = EncodingRef::Explicit(UTF_8);
```
dralley, line 27: Encoding doesn't feel like it belongs on the parser.  It's primarily an aspect of reading but the parser can override the original default.  But as I mentioned on the other PR even the manual override isn't really something that encoding_rs nor encoding_rs_io will give you.

src/reader/mod.rs
```diff
@@ -127,6 +127,36 @@ macro_rules! configure_methods {
     };
 }
 
+/// Generalization of `read_to_end` method for buffered and borrowed readers
+macro_rules! read_to_end {
```
dralley, line 131: I agree in theory with deduplication, but it would be better to wait until everything is settled before doing it.  Please drop this commit for now, we can bring it back later if it still makes sense.

src/reader/slice_reader.rs
Mingun: Encoding is an interesting field. In can be part of configuration (when you create reader `from_str`) and it can be part of the parse state (when you create reader by other functions). Because I plan to use that in such manner:
```rust
/// Convert any synchronous reader to asynchronous one if inner reader support that
impl<R: AsyncBufRead + Unpin> From<Reader<R>> for Reader<TokioReader<R>> {
    fn from(reader: Reader<R>) -> Self {
        Self {
            reader: TokioReader(reader.reader),
            parser: reader.parser,
        }
    }
}

/// Convert any asynchronous reader to a synchronous one if inner reader support that
impl<R> From<Reader<TokioReader<R>>> for Reader<R> {
    fn from(reader: Reader<TokioReader<R>>) -> Self {
        Self {
            reader: reader.reader.0,
            parser: reader.parser,
        }
    }
}
```
it is convenient to leave it in the `Parser`. It will be possible to move it later if it is required

src/reader/mod.rs
```diff
@@ -460,203 +410,6 @@ impl<R> Reader<R> {
...
-    /// If `encoding` feature is enabled and no encoding is specified in declaration,
-    /// defaults to UTF-8.
-    fn decoder(&self) -> Decoder {
-        Decoder {
-            #[cfg(feature = "encoding")]
-            encoding: self.encoding.encoding(),
-        }
-    }
-}
-
```
dralley, line 637: The logical separation between `Reader` and `Parser` is not very clean, if I had to describe it it seems like `Reader` is parsing the boundaries of elements and `Parser` is parsing *within* elements, but even that isn't quite accurate.  Basically `Reader` is still doing parsing and `Parser` is still doing reading.

I understand there are mechanical reasons for the separation but architecturally these labels don't make much sense.

Mingun: I partially agree with that -- `Reader` is still doing parsing, but `Parser` does only parsing. It consumes already read byte arrays and build events on top of them.

src/reader/mod.rs
Mingun: Removed

src/reader/mod.rs
```diff
@@ -405,15 +405,15 @@ impl<R> Reader<R> {
...
     {
-        let event = match self.parser.tag_state {
-            TagState::Init => self.read_until_open(buf, true),
-            TagState::Closed => self.read_until_open(buf, false),
-            TagState::Opened => self.read_until_close(buf),
-            TagState::Empty => self.parser.close_expanded_empty(),
-            TagState::Exit => return Ok(Event::Eof),
+        let event = match self.parser.state {
+            ParseState::Init => self.read_until_open(buf, true),
+            ParseState::Closed => self.read_until_open(buf, false),
```
dralley, line 410: Might make sense to rename these variants as e.g. `ParseState::ClosedTag`,  `ParseState::OpenTag`

src/reader/mod.rs
Mingun: Clarified

src/reader/parser.rs
```diff
@@ -19,21 +19,21 @@ use memchr;
 /// get back produced [`Event`]s.
 #[derive(Clone)]
 pub(super) struct Parser {
-    /// current buffer position, useful for debugging errors
+    /// Number of bytes read from the source of data since the parser was created
```
dralley, line 22: Which is not going to work once we support multiple encodings, unfortunately.

Mingun: We can change it when we changes the working with encodings

dralley:
Why don't you like "rebase and merge"

Mingun:
Without merge commit it is hard to figure out in which PR work was done and where the one PR finished and the new started

Issue #257, opened 2021-02-04
Help deserialize mixed tags and string in body $value (html text formatting)


Rudo2204:
I'm trying to deserialize some dictionary defitnitions and came across this one which contains mixed multiple tags with normal string (html text formatting).
```xml
<div style="margin-left:2em"><b>1</b> 〔学業･技術などの能力判定〕 an examination; a test; 《口》 an exam; 《米》 a quiz 《<i>pl</i>. quizzes》.</div>
```
I looked around in [serde-xml-rs tests](https://github.com/RReverser/serde-xml-rs/blob/master/tests/test.rs) and tried this solution which seems to be close but it doesn't quite work
```rust
#[derive(Debug, Deserialize, PartialEq)]
struct DivDefinition {
    style: String,
    #[serde(rename = "$value")]
    definition: Vec<MyEnum>,
}

#[derive(Debug, Deserialize, PartialEq)]
enum MyEnum {
    b(String),
    #[serde(rename = "$value")]
    String,
    i(String),
}
```
The error I'm getting is:
```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("unknown variant `〔学業･技術などの能力判定〕 an examination; a test; 《口》 an exam; 《米》 a quiz 《`, expected one of `b`, `$value`, `i`")'
```
I can make it work for now by not using `MyEnum` and just use `definition: Vec<String>`, but then I wouldn't know which text is bold and which is italic.
How can I properly deserialize this?

dralley:
Whoever picks this up, consider starting from https://github.com/tafia/quick-xml/pull/511

lirkwood:
Has anybody found a workaround for this? I am having the same issue.

lirkwood:
You can close this. Don't know when it was fixed but the original example works now with minor edits:
```rust
#[derive(Debug, Deserialize, PartialEq)]
struct DivDefinition {
    #[serde(rename = "@style")]
    style: String,
    #[serde(rename = "$value")]
    definition: Vec<MyEnum>,
}

#[derive(Debug, Deserialize, PartialEq)]
enum MyEnum {
    b(String),
    #[serde(rename = "$text")]
    String,
    i(String),
}
```

enricozb:
Thoughts on this idea? https://github.com/enricozb/quick-xml/commit/7b4b3f851a50ae9dbb45d54edfdc7c2374ec59d0

[...]

Issue #154, opened 2019-04-16
skips events on failed `read_text`


dunmatt:
Hi, I'm getting behavior that's gotta be a bug.  My parsing code is pretty straitforward:

```rust
    fn parse_calibration(calibration: &str) -> Result<AxiaCalibration, ConfigurationError> {
        let mut result: AxiaCalibration = Default::default();

        let mut reader = Reader::from_str(calibration);
        reader.trim_text(true);

        // 350 here to be long enough to hold the <calmtx> line in EXAMPLE_CALIBRATION.
        let mut buf = Vec::with_capacity(350); // this buffer gets recycled
        let mut inner_buf = Vec::with_capacity(350); // this buffer gets recycled

        loop {
            match reader.read_event(&mut buf) {
                Ok(Event::Start(e)) => {
                    if let Ok(s) = reader.read_text(e.name(), &mut inner_buf) {
                        println!("found {:?}", String::from_utf8(e.name().to_vec()));
                        parse_calibration_field(&e.name(), &s, &mut result)?;
                    } else {
                        println!("nogo {:?}", String::from_utf8(e.name().to_vec()));
                    }
                }
                Ok(Event::Eof) => break,
                Err(e) => return Err(e.into()),
                a => println!("{:?}", a), // ignore comments, end tags, random text, etc
            }
        }
        Ok(result)
    }
```
but I'm finding that it skips whatever comes immediately after `<netftCalibration>` in

```rust
    const EXAMPLE_CALIBRATION: &'static str = r#"<?xml version="1.0"?>
    <netftCalibration>
    <prodname>Ethernet Axia</prodname>
    <!-- Calibration Data-->
    <calthk>0</calthk>
    <caldis>0</caldis>
    <calsn>FT27120</calsn>
    <calpn>SI-500-20</calpn>
    <calfam>NET</calfam>
    <caldt>2/22/2019</caldt>
    <calmtx>101.194;-91.0611;-8.59668;-9.58448;-98.2309;106.267;-47.0668;64.063;106.857;-108.174;-66.2965;50.6429;59.2169;62.9028;57.032;61.8539;56.9527;62.1059;-2.19935;-1.12881;1.15961;-1.17096;0.976857;2.36193;-0.12006;2.04773;-1.9022;-2.01133;2.05653;-0.0719982;2.66346;-2.69029;2.5692;-2.52372;2.785;-2.80579</calmtx>
    <calfu>1</calfu>
    <scalfu>N</scalfu>
[...]

PR #450, opened 2022-07-31
Async support

Mingun:
First of all, I would like to thank @999eagle for her work. It was very useful for quickly obtaining a working result and its iterative improvement.

After week of experimenting and prepare work I think, that the result suits me. I tried to focus on those points:
- maximum reuse the code. For that all key methods rewritten using macros, so the sync and async code is the same
- try to minimize intermediate wrappers, because it is painful to pass long named types to functions, like in that example: https://github.com/tafia/quick-xml/blob/9ccc68620fb4f525d0316abe8e3cfac624780b56/src/reader/mod.rs#L324-L326

I have considered several options
- original suggestion with three intermediate structs: `SliceReader`, `BufferedReader` and `AsyncReader`:
  ```rust
  impl                  Reader<SliceReader> { ... }// for &[u8]
  impl<R: BufRead>      Reader<BufferedReader<R>> { ... }
  impl<R: AsyncBufRead> Reader<AsyncReader<R>> { ... }
  ```
  I have realized that we have enough only two of them therefore...
- two intermediate newtype structs, `SyncReader` and `AsyncReader`
  ```rust
  impl<R>               Reader<SyncReader<R>> { ... }// for &[u8] and BufRead
  impl<R: AsyncBufRead> Reader<AsyncReader<R>> { ... }
  ```
  After further experiments, it became clear that we can do without `SyncReader`, so...
- _(this one)_ sync code stays the same. Async code is implemented using an adapter struct, which I've name `TokioAdapter` (former `AsyncReader`)
  ```rust
  impl                  Reader<&[u8]> { ... }
  impl<R: BufRead>      Reader<R> { ... }
  impl<R: AsyncBufRead> Reader<AsyncReader<R>> { ... }
  ```
- do not introduce intermediate structs and just implement everything for `Reader<R>`
  ```rust
  impl                  Reader<&[u8]> { ... }
  impl<R: BufRead>      Reader<R> { ... }
  impl<R: AsyncBufRead> Reader<R> { ... }
  ```
  In this option the conflicts of names for internal methods would appear. In principle, there is nothing unsolvable, but if we decide to realize support for other asynchronous library, it will be perhaps not really convenient. However, my attempts to implement support for [`futures`](https://lib.rs/crates/futures) were unsuccessful -- API differs a little. Besides, there is nothing difficult in principle, but so far it seems that it is possible to do without it as as there are [adapters](https://lib.rs/crates/async-compat)

As it seemed, all realization differs slightly. I especially like the 4th option, as it eliminates the need for a long type if you need to pass `Reader` to the function. I think, I'll make also another PR with that variant.

I decided to not implement `from_file` for async reader, because it is required additional tokio feature and it seems that it can be easely done outside the crate.

Also, I did not implement `read_text_into_async` API, because it available not on all readers and I think, it should be reworked to return all intermediate notes as text as required for #257 and also was discussed in #154.

I'm not sure, whether we have to `Send` for inner type, as was originally implemented by @999eagle in #417, currently I do not require that. It seems, that compiler should automatically guarantee that when it can, but how to check that?

@999eagle, could you also review that?

Closes #417 
Closes #425

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,13 +8,14 @@ documentation = "https://docs.rs/quick-xml"
 repository = "https://github.com/tafia/quick-xml"
 
 keywords = ["xml", "serde", "parser", "writer", "html"]
-categories = ["encoding", "parsing", "parser-implementations"]
+categories = ["asynchronous", "encoding", "parsing", "parser-implementations"]
 license = "MIT"
 
 [dependencies]
 document-features = { version = "0.2", optional = true }
 encoding_rs = { version = "0.8", optional = true }
 serde = { version = "1.0", optional = true }
+tokio = { version = "1.20", optional = true, default-features = false, features = ["io-util"] }
 memchr = "2.5"
 
 [dev-dependencies]
@@ -23,6 +24,8 @@ pretty_assertions = "1.2"
 regex = "1"
 serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
+tokio = { version = "1.20", default-features = false, features = ["macros", "rt"] }
+tokio-test = "0.4"
 
 [lib]
 bench = false
@@ -37,6 +40,13 @@ harness = false
 
 [features]
 default = []
+
+## Enables support for asynchronous reading from `tokio`'s IO-Traits by enabling
+## [reading events] from types implementing [`tokio::io::AsyncBufRead`].
+##
+## [reading events]: crate::reader::Reader::read_event_into_async
+async-tokio = ["tokio"]
+
 ## Enables support of non-UTF-8 encoded documents. Encoding will be inferred from
 ## the XML declaration if it will be found, otherwise UTF-8 is assumed.
 ##
@@ -123,3 +133,7 @@ required-features = ["serialize"]
 [[test]]
 name = "serde-migrated"
 required-features = ["serialize"]
+
+[[test]]
+name = "async-tokio"
+required-features = ["async-tokio"]
--- a/Changelog.md
+++ b/Changelog.md
@@ -39,6 +39,7 @@
   |`attribute_namespace`   |`resolve_attribute`
 - [#439]: Added utilities `detect_encoding()`, `decode()`, and `decode_with_bom_removal()`
   under the `quick-xml::encoding` namespace.
+- [#450]: Added support of asynchronous [tokio](https://tokio.rs/) readers
 
 
 ### Bug Fixes
@@ -218,6 +219,7 @@
 [#439]: https://github.com/tafia/quick-xml/pull/439
 [#440]: https://github.com/tafia/quick-xml/pull/440
 [#443]: https://github.com/tafia/quick-xml/pull/443
+[#450]: https://github.com/tafia/quick-xml/pull/450
 
 
 ## 0.23.0 -- 2022-05-08
--- a/src/reader/async_tokio.rs
+++ b/src/reader/async_tokio.rs
@@ -1,0 +1,388 @@
+//! This is an implementation of [`Reader`] for reading from a [`AsyncBufRead`]
+//! as underlying byte stream. This reader fully implements async/await so reading
+//! can use non-blocking I/O.
+
+use std::future::Future;
+use std::pin::Pin;
+use tokio::io::{self, AsyncBufRead, AsyncBufReadExt};
+
+use crate::events::Event;
+use crate::name::{QName, ResolveResult};
+use crate::reader::buffered_reader::impl_buffered_source;
+use crate::reader::{is_whitespace, BangType, NsReader, ParseState, ReadElementState, Reader};
+use crate::{Error, Result};
+
+/// A struct for read XML asynchronously from an [`AsyncBufRead`].
+///
+/// Having own struct allows us to implement anything without risk of name conflicts
+/// and does not suffer from the impossibility of having `async` in traits.
+struct TokioAdapter<'a, R>(&'a mut R);
+
+impl<'a, R: AsyncBufRead + Unpin> TokioAdapter<'a, R> {
+    impl_buffered_source!('b, 0, async, await);
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+impl<R: AsyncBufRead + Unpin> Reader<R> {
+    /// An asynchronous version of [`read_event_into()`]. Reads the next event into
+    /// given buffer.
+    ///
+    /// > This function should be defined as
+    /// > ```ignore
+    /// > pub async fn read_event_into_async<'b>(
+    /// >     &mut self,
+    /// >     buf: &'b mut Vec<u8>
+    /// > ) -> Result<Event<'b>>;
+    /// > ```
+    /// > but Rust does not allow to write that for recursive asynchronous function
+    ///
+    /// This is the main entry point for reading XML `Event`s when using an async reader.
+    ///
+    /// See the documentation of [`read_event_into()`] for more information.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # tokio_test::block_on(async {
+    /// # use pretty_assertions::assert_eq;
+    /// use quick_xml::events::Event;
+    /// use quick_xml::Reader;
+    ///
+    /// // This explicitly uses `from_reader("...".as_bytes())` to use a buffered
+    /// // reader instead of relying on the zero-copy optimizations for reading
+    /// // from byte slices, which is provides the sync interface anyway.
+    /// let mut reader = Reader::from_reader(r#"
+    ///     <tag1 att1 = "test">
+    ///        <tag2><!--Test comment-->Test</tag2>
+    ///        <tag2>Test 2</tag2>
+    ///     </tag1>
+    /// "#.as_bytes());
+    /// reader.trim_text(true);
+    ///
+    /// let mut count = 0;
+    /// let mut buf = Vec::new();
+    /// let mut txt = Vec::new();
+    /// loop {
+    ///     match reader.read_event_into_async(&mut buf).await {
+    ///         Ok(Event::Start(_)) => count += 1,
+    ///         Ok(Event::Text(e)) => txt.push(e.unescape().unwrap().into_owned()),
+    ///         Err(e) => panic!("Error at position {}: {:?}", reader.buffer_position(), e),
+    ///         Ok(Event::Eof) => break,
+    ///         _ => (),
+    ///     }
+    ///     buf.clear();
+    /// }
+    /// assert_eq!(count, 3);
+    /// assert_eq!(txt, vec!["Test".to_string(), "Test 2".to_string()]);
+    /// # }) // tokio_test::block_on
+    /// ```
+    ///
+    /// [`read_event_into()`]: Reader::read_event_into
+    pub fn read_event_into_async<'reader, 'b: 'reader>(
+        &'reader mut self,
+        buf: &'b mut Vec<u8>,
+    ) -> Pin<Box<dyn Future<Output = Result<Event<'b>>> + 'reader>> {
+        Box::pin(async move {
+            read_event_impl!(self, buf, read_until_open_async, read_until_close_async, await)
+        })
+    }
+
+    /// An asynchronous version of [`read_to_end_into()`].
+    /// Reads asynchronously until end element is found using provided buffer as
+    /// intermediate storage for events content. This function is supposed to be
+    /// called after you already read a [`Start`] event.
+    ///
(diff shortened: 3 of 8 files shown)


benches/macrobenches.rs
```diff
@@ -1,6 +1,6 @@
 use criterion::{self, criterion_group, criterion_main, Criterion, Throughput};
 use quick_xml::events::Event;
-use quick_xml::Reader;
```
dralley, line 3: Let's at least make them available in a `prelude::*`

Cargo.toml
```diff
@@ -23,6 +24,8 @@ pretty_assertions = "1.2"
 regex = "1"
 serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
+tokio = { version = "1.20", default-features = false, features = ["macros", "rt-multi-thread"] }
```
dralley, line 27: Is `rt-multi-thread` actually needed?  It's still fundamentally a sequential operation, unlike handling e.g. network requests.

src/reader/ns_reader.rs
```diff
@@ -538,6 +538,10 @@ impl<'i> NsReader<&'i [u8]> {
     /// You also can use [`read_resolved_event()`] instead if you want to resolve namespace
     /// as soon as you get an event.
     ///
+    /// There is no asynchronous `read_event_async()` version of this function,
+    /// because it is meaningless to have it. You already have a buffer in a memory,
+    /// so you not needed to wait something (synchronously or asynchronously).
```
dralley, line 543:
```suggestion
    /// so you do not need to perform IO (either synchronously or asynchronously).
```

dralley, line 543:
```
    /// There is no asynchronous `read_event_async()` version of this function
    /// because it is not necessary - the contents are already in memory and no IO
    /// is necessary.
```

src/reader/ns_reader.rs
```diff
@@ -595,6 +599,10 @@ impl<'i> NsReader<&'i [u8]> {
     /// If you are not interested in namespaces, you can use [`read_event()`]
     /// which will not automatically resolve namespaces for you.
     ///
+    /// There is no asynchronous `read_resolved_event_async()` version of this function,
+    /// because it is meaningless to have it. You already have a buffer in a memory,
+    /// so you not needed to wait something (synchronously or asynchronously).
+    ///
```
dralley, line 605: ^^

benches/macrobenches.rs
Mingun: Do you think that this will cause usability issues? I don't think that it is needed to get this for that -- then it is just better to drop that commit.

Cargo.toml
Mingun: This is a test dependency and by default Rust runs all test in parallel, so I think it is needed. But I didn't dig deep, just copied from previous PRs

adamreichold: > by default Rust runs all tests in parallel, so I think it is needed.

This is not relevant as far as I understand it as each test decorated with [`#[tokio::test]`](https://docs.rs/tokio/latest/tokio/attr.test.html) will run within its own runtime instance (which can be either multi- or single-threaded).

However, I think the multi-threaded flavour is the default IIRC and one would need to explicitly select the single-threaded flavour if only the `rt` feature is enabled, so I think that this might be simpler for a development dependency which does not burden downstream crates.

src/reader/async_tokio.rs
```diff
@@ -0,0 +1,462 @@
...
+use crate::events::Event;
+use crate::name::{QName, ResolveResult};
+use crate::reader::buffered_reader::impl_buffered_source;
+use crate::reader::{is_whitespace, BangType, NsReader, ParseState, ReadElementState, Reader};
+use crate::{Error, Result};
+
+/// A struct that adapts any [`AsyncBufRead`] reader to read an XML.
+/// See [`Reader::read_event_into_async`] for details.
+#[derive(Debug, Clone)]
+pub struct TokioAdapter<R>(pub R);
```
999eagle, line 19: This `Reader` can't really override the `into_inner` implementation of `impl<R> Reader<R>` in `reader/mod.rs`, so a `Reader<TokioAdapter<R>>` only returns the `TokioAdapter<R>` when calling `into_inner`. To get the `AsyncBufRead` back out, `TokioAdapter` should also implement an `into_inner`.

Maybe there's a better way to do this like I did in #425 with the `InnerReader` trait?

src/reader/mod.rs
```diff
@@ -127,11 +127,152 @@ macro_rules! configure_methods {
     };
 }
 
+macro_rules! read_event_impl {
```
999eagle, line 130: All of these macros should be moved into a separate module (see also my comment in `buffered_reader.rs`).

Cargo.toml
```diff
@@ -23,6 +24,8 @@ pretty_assertions = "1.2"
 regex = "1"
 serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
+tokio = { version = "1.20", default-features = false, features = ["macros", "rt"] }
```
999eagle, line 27: I know you didn't want to have support for `from_file` functions for the async reader, but I think we should still have at least one test which actually reads from a file, which would require the `fs` feature here. At least in `dev-dependencies` I think it's okay to have more features.

Changelog.md
```diff
@@ -62,6 +63,7 @@
   and a document encoding is not an UTF-8
 - [#434]: Fixed incorrect error generated in some cases by serde deserializer
 - [#445]: Use local name without namespace prefix when selecting enum variants based on element names
+  in a serde deserializer
```
999eagle, line 66: This should be part of #445, not this PR.

Changelog.md
```diff
@@ -216,6 +221,8 @@
 [#437]: https://github.com/tafia/quick-xml/pull/437
 [#439]: https://github.com/tafia/quick-xml/pull/439
 [#440]: https://github.com/tafia/quick-xml/pull/440
+[#445]: https://github.com/tafia/quick-xml/pull/445
```
999eagle, line 224: This should be part of #445, not this PR.

999eagle (commented):
I think most of this PR looks good, thank you very much for adapting my work!

I haven't yet compared the old and new code exactly to make 100% sure they match, but they seem to match. There are a few changes I'd like to see, most notably the macros being moved to their own module to improve readability and get the `async` implementation out of `buffered_reader.rs`.

src/reader/async_tokio.rs
Mingun: `TokioAdapter` has a public field. Anyway, I think I found a better way and be able to hide `TokioAdapter` as an implementation detail. I'll update PR tomorrow

benches/macrobenches.rs
dralley: I'd prefer to keep everything available from the root namespace, but you don't have to drop the entire commit.  It still makes sense to have `reader` and `writer` modules.

src/reader/async_tokio.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// assert_eq!(count, 3);
+    /// assert_eq!(txt, vec!["Test".to_string(), "Test 2".to_string()]);
+    /// # }) // tokio_test::block_on
+    /// ```
+    ///
+    /// [`read_event_into()`]: Reader::read_event_into
+    pub fn read_event_into_async<'r, 'b: 'r>(
+        &'r mut self,
+        buf: &'b mut Vec<u8>,
+    ) -> Pin<Box<dyn Future<Output = Result<Event<'b>>> + 'r>> {
```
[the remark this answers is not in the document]
Mingun: The `'_self` name actually does not clarify nothing. `'r` is abbreviation of `'reader` which seems too long for me and I don't see good abbreviation. So this method should read as: "returns `Event` which borrow data from `buf` and that buffer live longer than reader _self_"

src/reader/buffered_reader.rs
```diff
@@ -12,6 +12,216 @@ use crate::events::Event;
 use crate::name::QName;
 use crate::reader::{is_whitespace, BangType, ReadElementState, Reader, XmlSource};
 
+macro_rules! impl_buffered_source {
```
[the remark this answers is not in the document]
Mingun: I don't know... Having macro in this place first of all indicates that it implements a buffering reader. Because "buffering" trait of this macro seems to me more important, than "async" trait, it is defined here. Moving it into another module will break the explicit link with that usage place, as I think

src/reader/mod.rs
Mingun: The advantage of their definition there is that here is one usage place and those macros is very tied to functions where they used. I think it will be better to leave them here.

Cargo.toml
Mingun: Actually, I do not see why we need to test actual reading from a file. I doubt that this test could catch something that is not catched by the test that read from memory. For example, we also does not have tests for sync reader that reads from a file (we have only two examples for that).

Changelog.md
Mingun: Yes, I think I'll split up this to a dedicated PR with other small stuff

Cargo.toml
Mingun: Well, because both features are optional, I think we will use the least demanding feature, i.e. `"rt"`

benches/macrobenches.rs
Mingun: Ok, therefore, because after my updates it does not add any value to async PR I'll move it to another small PR with various stuff

src/reader/async_tokio.rs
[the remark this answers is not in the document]
999eagle: `'_self` was just an example because it's a name I like to use in my projects. Maybe we could use `'rdr` here for the reader and `'buf` for the buffer to have more descriptive names for lifetimes than `'r` and `'b`.

src/reader/buffered_reader.rs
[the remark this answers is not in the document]
999eagle: It's your choice in the end. Keeping it here would work too.

tests/async-tokio.rs
```diff
@@ -0,0 +1,20 @@
+use quick_xml::events::Event::*;
+use quick_xml::Reader;
+
+#[tokio::test]
+async fn test_sample() {
+    let src = include_str!("documents/sample_rss.xml");
```
dralley, line 6: This example should probably open a file instead of working with a `&str`

dralley:
I really wish we could try out the internal buffering idea (no more `read_event_into()`, `&str` returned in all cases) before introducing all these macros, as I think there may be an opportunity there for macro-less deduplication.  The macros do make it a lot harder to actually read and understand the code.  However I'm also tied up with work again for the next week and a half though so I don't have much time to spend prototyping.

So I would say I'm -0, I won't block it but I'm not thrilled about that particular change to be honest.

tests/async-tokio.rs
Mingun: We do not do that for `BufRead` implementation, so it is unclear why we should do that in that case

dralley: Sorry, I think I though this test was an example last night.  It's fine to do this in tests.

src/reader/async_tokio.rs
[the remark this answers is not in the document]
Mingun: In the end I've changed `'r` to `'reader`, but leave `'b` -- I think it is quite descriptive and is already used in many places to indicate the lifetime of the buffer. In addition, with this choice of names, `mut` are beautifully aligned in the code and documentation :innocent:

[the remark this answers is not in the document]
999eagle: Okay, sounds good :)

Mingun:
> I really wish we could try out the internal buffering idea

I think that the idea to allow the user to manage the buffer with the event content as he wish has they own benefits. Actually, that approach very similar to how `Read` trait in the standard library is defined: you provide a buffer that will filled with some content.

So we could investigate that idea, and maybe even provide a special buffering reader for users that do not want to manage the buffer himself, but we should keep the current API. That API allows to apply the reader as an operator to a buffer and create events over it, which does not tied to the reader itself, so it can be even destroyed if it is not required anymore.

----
Since there are no fundamental objections, I think I will merge it in a few hours.

dralley:
@999eagle Do you feel like this feature is living up to your expectations?  Any issues with it in practice?

999eagle:
Hi @dralley, I just now got around to doing so, but I just published my [crate](https://crates.io/crates/async-xml) using this feature to implement async xml deserialization with derive-macros.

The actual crate I'm working on with this is also coming along fine, so (at least for now) this is fulfilling my expectations.

PR #451, opened 2022-08-01
Extend macrobenchmark suite to cover namespaces, buffered vs. nocopy

dralley:


benches/macrobenches.rs
```diff
@@ -41,45 +67,141 @@ fn parse_document(doc: &[u8]) -> XmlResult<()> {
...
+        group.throughput(Throughput::Bytes(data.len() as u64));
+        group.bench_with_input(*id, *data, |b, input| {
+            b.iter(|| parse_document_from_str(input).unwrap())
+        });
+    }
+
+    group.finish();
+}
+
+/// Decode into a buffer, then parse
```
dralley, line 160: I'm aware that this is a premature description

benches/macrobenches.rs
```diff
@@ -41,45 +67,143 @@ fn parse_document(doc: &[u8]) -> XmlResult<()> {
...
+    }
+    Ok(())
+}
+
+// TODO: use fully normalized attribute values
+fn parse_document_from_str_with_namespaces(doc: &str) -> XmlResult<()> {
+    let mut r = NsReader::from_str(doc);
+    loop {
+        match r.read_resolved_event()? {
+            (_, Event::Start(e) | Event::Empty(e)) => {
```
dralley, line 101: I'm not certain if something more needs to be done to prevent the namespace code from being entirely optimized out.

benches/macrobenches.rs
```diff
@@ -1,25 +1,51 @@
...
+static LIBREOFFICE_DOCUMENT: &str = include_str!("../tests/documents/libreoffice_document.fodt");
+static DOCUMENT: &str = include_str!("../tests/documents/document.xml");
+static TEST_WRITER_INDENT: &str = include_str!("../tests/documents/test_writer_indent.xml");
+static SAMPLE_1: &str = include_str!("../tests/documents/sample_1.xml");
+static LINESCORE: &str = include_str!("../tests/documents/linescore.xml");
+static SAMPLE_RSS: &str = include_str!("../tests/documents/sample_rss.xml");
+static SAMPLE_NS: &str = include_str!("../tests/documents/sample_ns.xml");
+static PLAYERS: &str = include_str!("../tests/documents/players.xml");
+
+static INPUTS: &[(&str, &str)] = &[
```
dralley, line 19: Probably we do want to ask the question at some point - which benchmarks are strictly necessary to get coverage of the things we want to test.  Because otherwise it will take a very long time to run them all.

Mingun: I think, that documentation would be enough

benches/macrobenches.rs
Mingun: I think you need to send resolved namespace into `black_box`

Mingun (approved):

benches/macrobenches.rs
dralley: FYI the current API seems to have a slight practical issue:

```
error[E0502]: cannot borrow `r` as immutable because it is also borrowed as mutable
   --> benches/macrobenches.rs:103:74
    |
100 |         match criterion::black_box(r.read_resolved_event()?) {
    |                                    ----------------------- mutable borrow occurs here
...
103 |                     criterion::black_box(attr?.decode_and_unescape_value(&r)?);
    |                                                                          ^^ immutable borrow occurs here
104 |                 }
105 |                 criterion::black_box(resolved_ns);
    |                                      ----------- mutable borrow later used here
```

for 

```
    loop {
        match criterion::black_box(r.read_resolved_event()?) {
            (resolved_ns, Event::Start(e) | Event::Empty(e)) => {
                for attr in e.attributes() {
                    criterion::black_box(attr?.decode_and_unescape_value(&r)?);
                }
                criterion::black_box(resolved_ns);
            }
.....
```

Mingun: Yes, I known. That is the one reason why I've put `Decoder` into events and would prefer to take `Decoder` instead of `&Reader` in attributes until better solution would be developed.

In that case the situation could be resolved, if you consume `resolved_ns` first

PR #458, opened 2022-08-16
Strip BOM from the event data, add an encoding configuration option on the writer for writing the BOM

dralley:
We need to allow users to write a BOM, but encoding it in the event stream creates other issues.  Instead lets delegate it to an option on the `Writer`.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -133,7 +133,6 @@ impl<W: Write> Writer<W> {
     pub fn write_event<'a, E: AsRef<Event<'a>>>(&mut self, event: E) -> Result<()> {
         let mut next_should_line_break = true;
         let result = match *event.as_ref() {
-            Event::StartText(ref e) => self.write(&e),
             Event::Start(ref e) => {
                 let result = self.write_wrapped(b"<", e, b">");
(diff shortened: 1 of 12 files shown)


dralley:
Somewhat depends on your feelings about https://github.com/tafia/quick-xml/pull/441#issuecomment-1215449571

src/writer.rs
```diff
@@ -65,14 +88,34 @@ impl<W: Write> Writer<W> {
...
             writer: inner,
             indent: Some(Indentation::new(indent_char, indent_size)),
+            encoding: EncodingScheme::default(),
+            first_write: true,
+        }
+    }
+
+    /// Creates a Writer from a generic Write implementor with configured whitespace indents and a
+    /// specified encoding scheme.
+    pub fn new_with_indent_and_encoding(
```
dralley, line 108: This feels like it's right up at the line of perhaps justifying a builder.

dralley:
Sidenote: we should probably implement a `Writer::from_file()` that wraps the file in a `BufReader`.   It is easy to forget and would destroy performance.

dralley:
Rethinking the approach slightly.

Issue #285, opened 2021-04-17
Disable `trim_text` in Deserializer from_reader


woodworker:
Is there a easy way to set trim_text to false in the `Deserializer::from_str` when i use `quick_xml::de::from_str`?

https://github.com/tafia/quick-xml/blob/a4be48431a2c512683538b81d7ee037af70d5cce/src/de/mod.rs#L160-L167

ImJeremyHe:
Is there a way to determine whether set trim_text by looking if there is xml:space = "preserve"?
For example:
```
<t xml:space="preserve">Text </t>
```
The trailing space here should not be trimmed.

tafia:
There are very little customization on the serde deserializer so far. I don't think there is any major blocking point, someone just needs to write it.

Mingun:
In the coming release `Deserializer::new` would be public and you could create a deserializer from a `Reader` (but do not turn off `expand_empty_elements`! For now `Deserializer` is not prepared for that).

Processing of `xml:space` still waits its own PR

naumazeredo:
> In the coming release `Deserializer::new` would be public

Was this implemented already? `Deserializer::new` is public in the latest version, but it seems useless since `XmlRead` can't be implemented outside of quick_xml and `SliceReader` and `IoReader` can't be instantiated also. There's no way to use `Deserializer::new`, unless I'm missing something here.

Mingun:
Yes, this is oversight. So, currently this is still not possible, even in `master`. Need to think about better API. I also wold to provide an API to create a deserializer for a part of XML, so you can mix usual `Reader` usage with the `Deserializer` usage, for example, to support streaming deserialization.

Mingun:
According to the original use case -- I do not think that simply disabling `trim_text` would be usable -- it seems that you'll just break deserialization of pretty-printed XMLs at all with that setting

naumazeredo:
Yeah, that's exactly what happened. I didn't get why that option, even though internal, exists.
I've sadly moved back to serde_xml_rs since they give an option to not trim and I'm not willing to spend more time debugging xml deserialization right now. I'll be trying quick_xml in the future in case it gets more versatility

Mingun:
> The trimming of spaces within elements probably ought to be separated from the trimming of spaces between elements.

Yes, I think, we should move in that direction. A couple of thoughts:
- need to take into account a #383 problem. I think, it can be solved by introducing a method to read all content as a string regardless of the XML markup inside:
  ```rust
[...]

PR #459, opened 2022-08-17
Strip BOM from the event stream, add a method to Writer for writing BOM

dralley:


dralley:
What I actually want to do: provide a configuration option for either `trim_before_first_element` which would skip all non-text bytes, or a `trim_interelement_text` option which would trim all text events in places where they aren't supposed to be (which would address cover this as well as address https://github.com/tafia/quick-xml/issues/285#issuecomment-1199756671).  When the configuration is disabled it should be possible to fully reproduce the document from the events of the original reader. 

But after attempting to implement it I realized it requires more substantial parsing changes then I'm prepared to make right now.

src/reader/parser.rs
```diff
@@ -61,42 +59,39 @@ pub(super) struct Parser {
...
-            Event::Text(BytesText::wrap(content, self.decoder()))
-        })
+        if first {
+            #[cfg(feature = "encoding")]
+            if self.encoding.can_be_refined() {
+                if let Some(encoding) = encoding::detect_encoding(bytes) {
+                    self.encoding = EncodingRef::BomDetected(encoding);
+                    content = encoding::remove_bom(content, encoding);
+                }
+            }
```
Mingun, line 87: I plan to pull the detection from this method into `read_until_open!`. Here it is too late to do it -- for example, for UTF-16BE without BOM `bytes` will contain `[0x00]` instead of `[0x00, '<', 0x00, '?']` which we use for detection. I send a PR later when I'll add all test files

Mingun (commented):
I like the idea of the `trim_interelement_text` option (the name, probably, could be better), that's what I was thinking.

I'm not sure that we should remove `StartText` if we want
- return text only as an `str`
- return the BOM, which is not an `str` (UTF BOMs seems to be representable as special characters, but I'm not sure that this is a general case. Should we bother about other possible BOMs?)

`StartText` solves this problem by splitting out the ordinary text from the BOM bytes and other unexpected chars before the declaration, which are not according to the standard, but I think, could be used in the wild (the examples is our tests where it is often convenient to have leading spaces before a declaration). Actually, it should be named `StartBytes`, because this is not a text

I agree that using `StartText` event not the best way to write BOM, the writer should write it by himself. That is why I was thinking about splitting event into events for reading and events for writing, for which reason public `reader` and `writer` modules was introduced and `Event::borrow` method was implemented.

Having different events we could fix #332 (because reader event no longer will own the data and we could return right lifetime with existing `Attributes` struct).

Different events for reading and writing also has an advantage, that reader event can contain bytes internally and decoded only on demand (although we have a goal to step away from that, but is can have non-obvious implications) and a span, but writer event could 
- own their content
- store their content in UTF-8 only (and in fact right now you _should_ use only UTF-8 -- event generation in other encodings seems to be broken. serde serializer, for example, always writes in UTF-8)
- does not have a span which is anyway is useless for it

What exactly prevents `StartText`?

dralley:
> I'm not sure that we should remove StartText if we want

> *    return text only as an str
> *    return the BOM, which is not an str (UTF BOMs seems to be representable as special characters, but I'm not sure that this is a general case. Should we bother about other possible BOMs?)
>StartText solves this problem by splitting out the ordinary text from the BOM bytes and other unexpected chars before the declaration, which are not according to the standard, but I think, could be used in the wild (the examples is our tests where it is often convenient to have leading spaces before a declaration). Actually, it should be named StartBytes, because this is not a text


This is a misunderstanding of the BOM that I also shared until a day or two ago. 

The BOM is actually *not* a magic sequence of bytes.  It actually *is* just text.  The BOM is a unicode character, the "byte order mark" character, and its presence at the beginning of a document means you can look at the starting bytes of the document and determine the encoding based on whether these bytes are the UTF-8, UTF-16 LE, or UTF-16 BE representation of the BOM character.

This makes total sense in retrospect but for whatever reason, it took me forever to recognize the implications :shrug: 

Because of this, when you decode a UTF-16 LE document with the corresponding BOM, you get a UTF-8 document with a UTF-8 BOM.  Because it's just a character, not magic bytes.  You can actually just encode and decode it all day long because it's baked into Unicode.

Therefore, since it is truly just `Text`, we don't need a separate `StartText`.  The problem with the BOM is that it's annoying and only relevant to the decoder, but not really part of "the XML", even though it's just text - which is the same problem we have with un-trimmed whitespace between elements.

Changelog.md
```diff
@@ -186,11 +175,14 @@
...
   array cannot guarantee borrowing due to possible copying while decoding.
 
-- [#455]: Removed `Reader::read_text_into` which is only not a better wrapper over match on `Event::Text`
+- [#455]: Removed `Reader::read_text_into` which is just a thin wrapper over match on `Event::Text`
 
 - [#456]: Reader and writer stuff grouped under `reader` and `writer` modules.
   You still can use re-exported definitions from a crate root
 
+- [#458]: Made the `Writer::write()` method non-public as writing random bytes to a document is not generally useful or desirable.
+- [#458]: BOM bytes are no longer emitted as `Event::Text`. To write a BOM, use `Writer::write_bom()`.
```
Mingun, line 184: Actually, now it is #459 
```suggestion
- [#459]: Made the `Writer::write()` method non-public as writing random bytes to a document is not generally useful or desirable.
- [#459]: BOM bytes are no longer emitted as `Event::Text`. To write a BOM, use `Writer::write_bom()`.
```

Changelog.md
```diff
@@ -234,7 +226,7 @@
 [#450]: https://github.com/tafia/quick-xml/pull/450
 [#455]: https://github.com/tafia/quick-xml/pull/455
 [#456]: https://github.com/tafia/quick-xml/pull/456
-
+[#458]: https://github.com/tafia/quick-xml/pull/458
```
Mingun, line 229:
```suggestion
[#459]: https://github.com/tafia/quick-xml/pull/459
```

src/encoding.rs
```diff
@@ -9,6 +9,15 @@ use encoding_rs::{Encoding, UTF_16BE, UTF_16LE, UTF_8};
...
 
+/// Unicode "byte order mark" encoded as UTF-8
+pub(crate) const UTF8_BOM: &[u8] = &[0xEF, 0xBB, 0xBF];
+/// Unicode "byte order mark" encoded as UTF-16 with little-endian byte order
+#[allow(dead_code)]
+pub(crate) const UTF16_LE_BOM: &[u8] = &[0xFF, 0xFE];
+/// Unicode "byte order mark" encoded as UTF-16 with big-endian byte order
+#[allow(dead_code)]
+pub(crate) const UTF16_BE_BOM: &[u8] = &[0xFE, 0xFF];
+
```
Mingun, line 20:
```suggestion
/// Unicode "byte order mark" (\u{FEFF}) encoded as UTF-8.
/// See <https://unicode.org/faq/utf_bom.html#bom1>
pub(crate) const UTF8_BOM: &[u8] = &[0xEF, 0xBB, 0xBF];
/// Unicode "byte order mark" (\u{FEFF}) encoded as UTF-16 with little-endian byte order.
/// See <https://unicode.org/faq/utf_bom.html#bom1>
#[cfg(feature = "encoding")]
pub(crate) const UTF16_LE_BOM: &[u8] = &[0xFF, 0xFE];
/// Unicode "byte order mark" (\u{FEFF}) encoded as UTF-16 with big-endian byte order.
/// See <https://unicode.org/faq/utf_bom.html#bom1>
#[cfg(feature = "encoding")]
pub(crate) const UTF16_BE_BOM: &[u8] = &[0xFE, 0xFF];

```

src/reader/mod.rs
```diff
@@ -553,8 +553,6 @@ impl<R> Reader<R> {
     }
 
     /// Read until '<' is found and moves reader to an `OpenedTag` state.
-    ///
-    /// Return a `StartText` event if `first` is `true` and a `Text` event otherwise
```
Mingun, line 557: It is better to leave the
```
Return a `Text` event
```
part

src/reader/mod.rs
```diff
@@ -1573,16 +1571,6 @@ mod test {
...
-                #[$test]
-                $($async)? fn start_text() {
-                    let mut reader = Reader::from_str("bom");
-
-                    assert_eq!(
-                        reader.$read_event($buf) $(.$await)? .unwrap(),
-                        Event::StartText(BytesText::from_escaped("bom").into())
-                    );
-                }
-
```
Mingun, line 1585: Need to replace the `text` test content with the body of this test (with replacing of `StartText` to `Text` and `"bom"` to `"text"` of course).

Also, it would be worth to add a test for a text with only BOM:
```rust
let mut reader = Reader::from_str("\u{feff}");
assert_eq!(
    reader.$read_event($buf) $(.$await)? .unwrap(),
    // what to expect?
);
```

src/writer.rs
```diff
@@ -86,11 +88,43 @@ impl<W: Write> Writer<W> {
         &mut self.writer
     }
 
+    /// Write a Byte-Order-Mark character to the document.
+    ///
+    /// # Example
+    ///
+    /// ```rust
```
Mingun, line 95:
```suggestion
    /// Write a [Byte-Order-Mark] character to the document.
    ///
    /// # Example
    ///
    /// ```
```

src/writer.rs
```diff
@@ -86,11 +88,43 @@ impl<W: Write> Writer<W> {
...
+    ///     .write_empty()
+    ///     .expect("failure");
+    ///
+    /// assert_eq!(
+    ///     std::str::from_utf8(&buffer).unwrap(),
+    ///     "\u{FEFF}<empty attr1=\"value1\"/>"
+    /// );
+    /// # Ok(())
+    /// # }
+    /// ```
```
Mingun, line 119:
```suggestion
    /// ```
    /// [Byte-Order-Mark]: https://unicode.org/faq/utf_bom.html#BOM
```

Mingun (approved):
> It's not that it's able to be represented as a special character but that it is _literally just a character_, with an assigned Unicode character number, which can be encoded in different encodings just like any other Unicode character - and if you put this character at the beginning of the document, and you know how different encodings represent that character, then you can just look at the first few bytes and see if they match how a particular encoding would encode the character U+FEFF.

OK, the official FAQ about that: https://unicode.org/faq/utf_bom.html#bom1

Also I've made some experiments:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4f7773f8164bd228c4f35329b915f38b

Generally approved, but please made the small improvements in doc & tests

dralley:
@Mingun re: test, I've marked it as an expected failure until we can address the same issue you pointed out.  We can strip out the BOM here but it's too late to avoid emitting the text event entirely.

But we had the same issue before with `*_with_bom_removal()`, so it's not a regression.

PR #462, opened 2022-08-22
Minor attribute performance related changes

dralley:
I think this (at least `with_checks(false)`) got lost in the mess of rebasing a few months ago.

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -226,7 +226,7 @@ impl<'a> BytesStart<'a> {
         bytes.push(b' ');
         bytes.extend_from_slice(a.key.as_ref());
         bytes.extend_from_slice(b"=\"");
-        bytes.extend_from_slice(&*a.value);
+        bytes.extend_from_slice(a.value.as_ref());
         bytes.push(b'"');
     }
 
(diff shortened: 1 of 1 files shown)


src/events/mod.rs
```diff
@@ -226,7 +226,7 @@ impl<'a> BytesStart<'a> {
         bytes.push(b' ');
         bytes.extend_from_slice(a.key.as_ref());
         bytes.extend_from_slice(b"=\"");
-        bytes.extend_from_slice(&*a.value);
+        bytes.extend_from_slice(a.value.as_ref());
```
dralley, line 229: Just for consistency.

src/events/mod.rs
```diff
@@ -258,7 +258,7 @@ impl<'a> BytesStart<'a> {
         &'a self,
         attr_name: N,
     ) -> Result<Option<Attribute<'a>>> {
-        for a in self.attributes() {
+        for a in self.attributes().with_checks(false) {
```
dralley, line 261: IMO, if you're asking for the first occurrence of one particular attribute, you probably don't care much about about getting an error back on finding some duplicate attribute.  It's almost twice as fast without the check.

Issue #461, opened 2022-08-18
There should be a configuration option for skipping non-meaningful text events in-between XML tags


dralley:
We should provide an option which would address https://github.com/tafia/quick-xml/issues/285#issuecomment-1199756671 by trimming away `Text` events that do not have meaning to XML (e.g. whitespace between elements, or text prior to the first XML open tag which may include a BOM character)

To a first approximation this option essentially filters out things which aren't relevant to the information payload of an XML document, such as formatting.  This should replace the existing BOM stripping behavior.

Related to https://github.com/tafia/quick-xml/issues/285 and https://github.com/tafia/quick-xml/issues/414

PR #465, opened 2022-08-26
Correctly detect encoding even without BOM

Mingun:
This PR correctly processes input and allow to detect UTF-16 encoding even without BOM, using [the algorithm, recommended by W3C](https://www.w3.org/TR/xml11/#sec-guessing). Previously it is not worked because we ran it too late, when the `<` (0x3C) byte was already recognized as start of a tag and the "start text" does not contained it.

There is one thin thing there: when we should strip BOM? Currently I chosen the consistent behavior across `encoding` feature enabled / disabled and parsing from bytes / from `&str`, but probably it is not the best choice.

The available variants can be summarized in that table:
|                 |`from_reader`|`from_str`|
|----------- |--------------|-----------|
|`--features encoding`|auto-detected, BOM stripped|UTF-8, is it needed to strip possible BOM?
|no `--features encoding`|UTF-8, is it needed to strip possible BOM?|UTF-8, is it needed to strip possible BOM?|

I doubt about stripping BOM from `&str` arguments, because it is highly likely that BOM already was stripped when you've got the string, and it it the other BOM in the beginning, it is probably the content part (because BOM is ordinary character, U+FEFF, and can be used in any place like any other character).

However, when `encoding` feature is not enabled, the `from_reader` and `from_str` cases are the same -- input should be in UTF-8. Should we strip BOM in `from_reader` case then?

----
This PR also adds a set of files in all supported encodings most of which contains all characters that corresponding encoding can handle. The utility project for generation is in the `test-gen` folder


src/encoding.rs
```diff
@@ -111,9 +103,20 @@ impl Decoder {
     ///
     /// If XML declaration is absent in the XML, UTF-8 is used.
     ///
+    /// ----
     /// Returns an error in case of malformed sequences in the `bytes`.
     pub fn decode_with_bom_removal<'b>(&self, bytes: &'b [u8]) -> Result<Cow<'b, str>> {
```
Mingun, line 108: I'm not sure, should we keep this method? It is not used internally, maybe remove it?

Cargo.toml
```diff
@@ -118,6 +118,10 @@ all-features = true
 # See https://stackoverflow.com/questions/61417452
 rustdoc-args = ["--cfg", "docs_rs"]
 
+[[test]]
+name = "encodings"
+required-features = ["encoding"]
```
dralley, line 168: There are some tests here (which perhaps can be moved to another file) which we still want to run without the encoding feature, e.g. to verify that non-utf8 encodings fail.

However I have been wondering whether it's worth having an encoding feature to begin with going forwards.  Since we have to at least validate that the bytes being read are UTF-8 no matter what, and `encoding_rs` makes that somewhat easier (less brand new code needs to be written and we can delete all this conditional code).

src/encoding.rs
dralley: Delete it

src/encoding.rs
```diff
@@ -173,25 +180,26 @@ pub(crate) fn remove_bom<'b>(bytes: &'b [u8], encoding: &'static Encoding) -> &'
 ///
 /// | Bytes       |Detected encoding
 /// |-------------|------------------------------------------
-/// |`FE FF ## ##`|UTF-16, big-endian
+/// | **BOM**
+/// |`FE_FF_##_##`|UTF-16, big-endian
```
Mingun, line 184: Underscores added because otherwise long text in lower cells shrink the first column and leads to text wrapping, which looks not good

Cargo.toml
Mingun: Currently all tests in that file under `#[cfg(feature = "encoding")]`. I suppose, that you'll plan to add more later? Or you forget to remove feature gates from some tests?

dralley: That is currently true, but presuming we do want to have non-decoding reader support that just does validation, we'll want some verified-failure tests for parsing non-UTF-8 encodings.

src/encoding.rs
Mingun: The free function `decode_with_bom_removal` also is not used currently. Do you plan to use it in one of you PRs? Otherwise I suggest to remove it also. It seems that in any case, the user will not have to take care of the BOM, it will be handled by quick-xml internally and new access methods that could be introduced in #461 probably also will not require that function

Cargo.toml
Mingun: I think, I'll change it now as I've did that, and in the future we always can switch again when we will need that. I'll also move `fuzz` tests to a dedicated file

src/encoding.rs
dralley: I don't, we can remove it.  As you say, the options provided to `Reader` ought to be able to handle this eventually.

Mingun:
What about BOM stripping? Should we strip BOM when we parse from `&str`?

dralley:
I don't think so personally - https://github.com/tafia/quick-xml/issues/461 should handle it so that the input itself never needs to be transformed.  Unless you think we need it would be useful in the interim before that issue is dealt with.

Mingun:
Ok, then for now I leave it as implemented:

|                 |`from_reader`|`from_str`|
|----------- |--------------|-----------|
|`--features encoding`|auto-detected, BOM stripped if present|UTF-8, BOM stripped if present
|no `--features encoding`|UTF-8, BOM stripped if present|UTF-8, BOM stripped if present|

PR is ready for review

dralley:
Thanks.  I've been waiting for this to avoid overlapping work as I needed to make a few similar changes - will try to proceed with the decoding work this weekend.

Cargo.toml
```diff
@@ -54,9 +54,54 @@ async-tokio = ["tokio"]
 ## UTF-16 will not work (therefore, `quick-xml` is not [standard compliant]).
 ##
 ## List of supported encodings includes all encodings supported by [`encoding_rs`]
-## crate, that satisfied the restriction above.
+## crate, that satisfied the restriction above. So, the following encodings are
+## **not supported**:
```
dralley, line 58: There's a lot of redundancy between this, the lines above, and the lines below that can probably be cleaned up.

dralley (changes requested):

dralley (approved):
Changing it to approved so you can merge once the minor comments are addressed.

Cargo.toml
```diff
@@ -54,9 +54,54 @@ async-tokio = ["tokio"]
...
+## - [UTF-16BE]
+## - [UTF-16LE]
+## - [ISO-2022-JP]
+##
+## You should stop to process document when one of that encoding will be detected,
```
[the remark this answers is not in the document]
Mingun: Yes, I just want to explicitly warn users about that UTF-16 and ISO-2022-JP are not supported for now. For example, parsing documents with some Chinese characters that represented as `[ASCII byte, some byte]` in UTF-16BE or `[some byte, ASCII byte]` in UTF16LE can confuse the parser. That can be avoided if you stop to processing such documents in the very beginning.

I want make this change because I want to cut release this weekend, and it is still unclear when the correct solution will be ready, so it is best to avoid using problematic encodings for now.

I hope that in the next release this note will be removed. It should not break anything -- once correct support will be implemented, users will be updated and remove their guard code.

Cargo.toml
Mingun: I tried to remove redundancy, if you have something else in mind, please make a suggestion.

Cargo.toml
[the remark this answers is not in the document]
Mingun: I added a remark, that restriction is temporary and will be eliminated once #158 is fixed.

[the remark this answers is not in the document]
dralley: Thanks.  I do think finishing #158 this weekend is plausible but I won't promise it.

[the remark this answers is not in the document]
dralley: Well, an intermediate stage of it, in any case.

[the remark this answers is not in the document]
dralley: Or it would be, except for async.  Goddammit.  Keep forgetting about that.

Issue #218, opened 2020-06-22
Struct namespaces with Serde


Richterrettich:
Hi,
it would be really cool to have a feature to define common namespaces for structs using serde. Something like this:

```rust
#[derive(Serialize, Deserialize)]
#[namespace(F,foourn)]
struct Foo {
  id: String,
  #[serde(flatten)]
  #[namespace(B,barurn)]
  bar: Bar
}

#[derive(Serialize, Deserialize)]
struct Bar {
  name: String,
  desc: String
}
```
resulting in:

```xml
<F:foo xmlns:B="foourn" xmlns:F="barurn">
      <F:id>123</F:id>
      <B:name>asdf</B:name>
      <B:desc>foobar </B:desc>
</F:foo>
```

### Background

I sometimes face (old) API's that requrie XML to be structured in a way similar to the given example. They use namespaces to distinguish between entities to form some sort of inheritance tree.   
A feature like this would make interacting with these kind of API's very easy.

mlevkov:
I do not know how to handle this here, but I've used yaserde crate for such @Richterrettich

tafia:
This would be a nice feature indeed. Unfortunately I don't have much time but I'd be happy to integrate it if someone finds the time to do it.

WhyNotHugo:
I'd love to see this happen, but I've no idea where to even start. What kind of macro would `namespace` be in order for its data to somehow be accessible in the serde (de)serialization stage?

meghfossa:
Is there any workaround for this, or `:` is not supported at all. I don't mind if the workaround approach is verbose or unintuitive.

WhyNotHugo:
No workaround if you want serde. I wrote my own parser with https://docs.rs/quick-xml/latest/quick_xml/reader/struct.NsReader.html

Mingun:
If you don't want to write deserialization code manually, you could look at [xmlserde](https://lib.rs/crates/xmlserde). Support for namespaces for serde is not an easy task, unfortunately.

dralley:
Whoever picks this up, consider starting from https://github.com/tafia/quick-xml/pull/466

jespersm:
I want to give this one a try.

I'm aiming for the following feature set:

 * Easy specification of namespace information into serde derive attributes
 * Ability to ensure space-efficient serialization (i.e. by explicitly pushing the namespace definitions to the start of the serialized XML)
 * Ser/de for `xml:lang` and `xml:space`
[...]

PR #466, opened 2022-08-26
Proposal for serializing namespaces in serde addressing #218

JOSEPHGILBY:
I'm interested in forwarding the design / implementing a solution to the problem of serializing namespaces in Serde.

## Proposal from issue #218
The current proposal is to implement a 'namespace' attribute macro where the first argument is the prefix and the second argument is the namespace. Additionally, the proposal indicates that any child field will automatically take the defined prefix unless the field is shadowed by an additional namespace macro + serde flatten helper attribute.

Input: 
```rust
#[namespace(F,foourn)]
struct Foo {
  id: String,
  #[serde(flatten)]
  #[namespace(B,barurn)]
  bar: Bar
}

struct Bar {
  name: String,
  desc: String
}
```

output: 
```xml
<F:foo xmlns:B="foourn" xmlns:F="barurn">
      <F:id>123</F:id>
      <B:name>asdf</B:name>
      <B:desc>foobar </B:desc>
</F:foo>
```

### Potential downside
One downside I see with #218 is that the second xmlns statement will have to be 'hoisted' up into the foo tag. Additionally, multiple namespaces will **always** have to be used with the serde-flatten macro. The downside of using the flatten-macro is that quick-xml doesn't allow attributes to be directly defined (as currently implemented) in the parent container. As seen in the example above, the "id" field gets serialized into `<id>123</id>` and not `id="123"`. Thus, it doesn't look like it would be possible for id to be an xml attribute, which could be limiting for serializing xml documents.

## Alternative idea
Despite the proposal above, I want to explore a little bit on some other design ideas and get feedback on it. Playing around a little bit with attribute macros, I was able to get the following to process using `syn::parse2`. I haven't gotten it to fully serialize into xml quite yet, but I can indicate the expected xml.

### The proposal
Create a new derive macro called `QuickXml` with two helper attributes `xmlns` and `xmlpre`. What Serde does is that the `Serialize` and `Deserialize` macros are derive macros, but they list `serde` as a helper attribute. Thus, anytime you see `#[serde(...)]`, it isn't actually a separate macro which expands. Instead, the `Serialize` and `Deserialize` macros will be fed the location and parameters to the `#[serde(...)]` statements and use that for their expansion.

In a similar vein, we could have `QuickXml` be the derive macro which runs, but is fed `xmlns` and `xmlpre` helper attributes to expand-out namespace and prefix information for the struct. In contrast to the above proposal, there won't be a need for 'hoisting up' additional namespaces into the containing tag. Additionally, without being forced to use `serde(flatten)`, it's possible to both have multiple namespaces for a single tag, and, define attributes and elements at the same time for the tag. This would give a good amount of flexibility to the programmer for representing xml files as serde structs.

Another point is that `xmlns` and `xlmpre` would be one-to-one with the xml document in contrast to the above proposal. That is, as the programmer is reading through an example xml output which they would like to represent in rust using quick-xml and serde, everytime they see an 'xmlns:prefix=" and "prefix:attr-or-element-name" statement in the xml document, they would put an `#[xmlns:prefix="..."]` and `#[xmlpre:prefix]attr_or_element_name: type` into the rust serde struct. The downside is of course typing 'xmlpre' for each field in the struct and struct name which has a prefix on it. However, one-to-oneness makes it easier to visually verify the correctness between an example xml document and the rust serde struct the programmer is trying to make.

Finally, the format `xmlns:prefix="namespace"` or `xmlns="default-namespace"` used inside the attribute helper might be somewhat non-standard. However, this format allows the programmer to directly copy-and-paste xmlns statements from an example xml document, which provides a usability advantage.

### Example
Input: 
```rust
#[derive(Serialize, Deserialize, QuickXml)]
#[xmlns:F="foourn"]
#[xmlns:B="barurn"]
#[xmlns="http://this-is-a-default-namespace"]
#[xmlpre:B]
#[serde(rename = foo)]
struct Foo {

    // attributes

    #[xmlpre:F]
    id: String,

    mar: String

    // elements

    #[xmlpre:F]
    #[serde(rename = "$unflatten=element")]
    element: String,

    #[xmlpre:B]
    bar: Bar,
}

#[derive(Serialize, Deserialize, QuickXml)]
#[xmlns="http://this-is-another-default-namespace-shadowing-the-previous"]
#[serde(rename = "bar")]
struct Bar {

    // attributes

    #[xmlpre:F]
    name: String,

    #[xmlpre:F]
    desc: String,

    #[serde(rename = "default-namespace-field"]
    default_namespace_field: String

    // elements

    #[serde(rename = "$value")]
    pub body: String,
}
```
Intended output:
```xml
<B:foo xmlns:F="foourn" xmlns:B="barurn" xmlns="http://this-is-a-default-namespace" F:id="123" mar="asdf">
    <element>this is an element</element>
    <B:bar xmlns="http://this-is-another-default-namespace-shadowing-the-previous" F:name="a name" F:desc="asdf" default-namespace-field="another filled field">this is a value</B:bar>
</B:foo>
```

### Remaining questions
1. One possible issue from this proposal is multiply-defined prefixes for fields. That is, if a parent struct Parent has a field `#[xmlpre:A]child: Child`, but, Child is defined by `#[xmlpre:B] struct Child`, what should the outcome be?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ encoding_rs = { version = "0.8", optional = true }
 serde = { version = "1.0", optional = true }
 tokio = { version = "1.20", optional = true, default-features = false, features = ["io-util"] }
 memchr = "2.5"
+quick-xml-derive = { path = "quick-xml-derive" }
 
 [dev-dependencies]
 criterion = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ encoding_rs = { version = "0.8", optional = true }
 serde = { version = "1.0", optional = true }
 tokio = { version = "1.20", optional = true, default-features = false, features = ["io-util"] }
 memchr = "2.5"
-quick-xml-derive = { path = "quick-xml-derive" }
+quick-xml-derived = { path = "quick-xml-derived" }
 
 [dev-dependencies]
 criterion = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,8 @@ encoding_rs = { version = "0.8", optional = true }
 serde = { version = "1.0", optional = true }
 tokio = { version = "1.20", optional = true, default-features = false, features = ["io-util"] }
 memchr = "2.5"
-quick-xml-derived = { path = "quick-xml-derived" }
+quick_xml_derived = { path = "quick-xml-derived" }
+q_meta = { path = "quick-xml-derived/q_meta"}
 
 [dev-dependencies]
 criterion = "0.3"
@@ -27,6 +28,9 @@ serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
 tokio = { version = "1.20", default-features = false, features = ["macros", "rt"] }
 tokio-test = "0.4"
+phf = { version = "0.11.1", features = ["macros"] }
+q_meta = { path = "quick-xml-derived/q_meta"}
+
 
 [lib]
 bench = false
@@ -114,7 +118,7 @@ serialize = ["serde"]
 
 [package.metadata.docs.rs]
 # document all features
-all-features = true
+all-features = false
 # defines the configuration attribute `docs_rs` to enable feature requirements
 # See https://stackoverflow.com/questions/61417452
 rustdoc-args = ["--cfg", "docs_rs"]
@@ -133,6 +137,10 @@ required-features = ["serialize"]
 
 [[test]]
 name = "serde-migrated"
+required-features = ["serialize"]
+
+[[test]]
+name = "serde-namespace-se"
 required-features = ["serialize"]
 
 [[test]]
--- a/quick-xml-derived/Cargo.toml
+++ b/quick-xml-derived/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "quick-xml-derived"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+quote = "1.0.21"
+syn = { version = "1.0.99", features = ["full"] }
+
+[lib]
+proc-macro = true
--- a/quick-xml-derived/Cargo.toml
+++ b/quick-xml-derived/Cargo.toml
@@ -1,13 +1,20 @@
 [package]
-name = "quick-xml-derived"
+name = "quick_xml_derived"
 version = "0.1.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+proc-macro2 = "1.0.43"
 quote = "1.0.21"
 syn = { version = "1.0.99", features = ["full"] }
+serde = { version = "1.0", optional = true }
+phf = { version = "0.11.1", features = ["macros"] }
+
+[dev-dependencies]
+serde = { version = "1.0", features = ["derive"] }
+macrotest = "1.0.9"
 
 [lib]
 proc-macro = true
--- a/quick-xml-derived/Cargo.toml
+++ b/quick-xml-derived/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "quick_xml_derived"
 version = "0.1.0"
-edition = "2021"
+edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
@@ -14,7 +14,10 @@ phf = { version = "0.11.1", features = ["macros"] }
 
 [dev-dependencies]
 serde = { version = "1.0", features = ["derive"] }
+phf = { version = "0.11.1", features = ["macros"] }
 macrotest = "1.0.9"
+q_meta = { path = "q_meta"}
+trybuild = { version = "1.0.66", features = ["diff"] }
 
 [lib]
 proc-macro = true
--- a/quick-xml-derived/q_meta/Cargo.toml
+++ b/quick-xml-derived/q_meta/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "q_meta"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+phf = { version = "0.11.1", features = ["macros"] }
+syllogism = "0.1.3"
--- a/quick-xml-derived/q_meta/src/lib.rs
+++ b/quick-xml-derived/q_meta/src/lib.rs
@@ -1,0 +1,80 @@
+use phf::phf_map;
+
+#[derive(Debug)]
+pub struct QuickXmlMeta {
+    pub namespace_declarations: &'static [(&'static str, &'static str)],
+    pub identifier_prefix_map: phf::Map<&'static str, &'static str>
+}
+pub trait CurrentItemVisitorQXmlMeta {
+    fn visit_current_item_as_self<T>(
+        &self,
+        contained_visitor: &mut T,
+        ident_in_parent: Option<&'static str>,
+        parent_meta: Option<&'static QuickXmlMeta>,
+    ) -> &'static QuickXmlMeta
+    where
+        T: ContainedItemVisitorQXmlMeta;
+}
+
+impl CurrentItemVisitorQXmlMeta for String {
+    fn visit_current_item_as_self<T>(
+        &self,
+        contained_visitor: &mut T,
+        ident_in_parent: Option<&'static str>,
+        parent_meta: Option<&'static QuickXmlMeta>,
+    ) -> &'static QuickXmlMeta
+    where
+        T: ContainedItemVisitorQXmlMeta
+    {
+        &QuickXmlMeta {
+            namespace_declarations: &[],
+            identifier_prefix_map: phf_map! {},
+        }
+    }
+}
+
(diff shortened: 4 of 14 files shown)


Mingun:
Actually, there are already exists a library on top of quick-xml, that introduces custom traits & derive macro: [xmlserde](https://lib.rs/crates/xmlserde). After quick look it seems that it does not support namespaces nowadays. So maybe it would be better to contribute to that library.

----

About overall design. It should not be necessary to define prefixes in the type. Prefix is an entity of a concrete XML instance, not of an XML Schema, which is defined by Rust types. You can have an XML with multiple prefixes mapped to the one namespace and you should be able to deserialize type from any of them (even when part of fields uses one prefix, and the other part is another).

I think, it would be better to leave special trait and macro to third-party library (i.e. the second part of your proposal), and quick-xml itself will provide a serde serializer & deserializer as best as it could.

----

Your opening example then could be rewritten as:
```rust
#[xml(namespace = "foourn")]
struct Foo {
  id: String,   // QName: {foourn}id
  name: String, // QName: {foourn}name
  #[serde(flatten)]
  bar: Bar,     // QName: {foourn}bar
}

#[xml(namespace = "barurn")]
struct Bar {
  name: String, // QName: {barurn}name
  desc: String, // QName: {barurn}desc
}
```
Note, how different prefixes `B1` and `B2` mapped to the same namespace forming a valid XML, and how that is also work for a default namespace too:
```xml
<F:foo xmlns="foourn" xmlns:F="foourn" xmlns:B1="barurn" xmlns:B2="barurn">
      <F:id>123</F:id>
      <name>asdf</name>
      <B1:name>asdf</B1:name>
      <B2:desc>foobar</B2:desc>
</F:foo>
```

Here `#[xml(namespace = "...")]` attribute applied to a type, as in the ordinary XSD, where you define types in some namespace and all their elements are automatically defined in that namespace.

Strictly speaking, root element `<foo>` not necessary should be defined in the `"foourn"` namespace, the namespace of a root element is also explicitly defined in XSD by having `<xs:element>` under `<xs::schema>`. In quick-xml we believe, that when you call `from_str` (and other deserializer methods), you work with that namespace that is defined. So this will also be parsed:
```xml
<any-ns:any-tag xmlns="foourn" xmlns:F="foourn" xmlns:any-ns="some-other-ns" xmlns:B1="barurn" xmlns:B2="barurn">
      <F:id>123</F:id>
      <name>asdf</name>
      <B1:name>asdf</B1:name>
      <B2:desc>foobar</B2:desc>
</any-ns:any-tag>
```
Right now deserializer deserializes _type_, but root element is a part of _document_ (if you take any mature XML library in any language, you will find that conception everywhere).

----

Using only a vanilla serde, we cannot use `#[xml(namespace = "...")]` attribute directly. serde at all does not allow us to send any type attributes from type to (de)serializer. It could only tell us a name of a struct, and because we does not use that name in any case, we could encode namespace information in it:
```rust
//#[xml(namespace = "barurn")]
#[serde(rename = "{barurn}Bar")] // form a special name with a namespace information
struct Bar {
  name: String, // QName: {barurn}name
  desc: String, // QName: {barurn}desc
}
```

We can, however, try one trick. We could write a macro, that will take our type definition and rewrite it to serde format. So, having a macro `quick_xml::xml`, we could transform
```rust
// We should put the macro before the derive, so it runs first... I hope
#[xml(namespace = "barurn")] // or in full form: #[quick_xml::xml(namespace = "barurn")]
struct Bar {
  name: String, // QName: {barurn}name
  desc: String, // QName: {barurn}desc
}
```
to
```rust
#[serde(rename = "{barurn}Bar")] // form a special name with a namespace information
struct Bar {
  name: String, // QName: {barurn}name
  desc: String, // QName: {barurn}desc
}
```

I already had starting investigate this opportunity and it probably conceptually possible. I didn't work in that direction deeply, however, but I plan. One thing, that I can note here -- it is better to not try to write macro by hand. There is an excellent [darling](https://github.com/TedDriggs/darling) library which could process all boilerplate.

JOSEPHGILBY:
@Mingun 

I'll try to summarize your points before responding to them. Let me know if I misunderstand anything you said.

1. Somebody already made a custom macro library built-on top of quick-xml: xmlserde. It does not have namespaces so why not contribute there?
2. quick-xml's deserializer takes an xml-document (an instance) and maps the document's elements to the struct/enum/etc.'s fields based off of the resolved QName for the field. There may be a multitude of valid xml-documents which could satisfy this mapping. As an example 
```xml
<foo xmlns="foourn" xmlns:A="foourn" xmlns:B1="barurn" xmlns:B2="barurn">
    <A:id>123</A:id>
    <name>asdf</name>
    <B1:name>asdf</B1:name>
    <B2:desc>foobar</B2:desc>
</foo>
```
and 
```xml
<foo xmlns="foourn" xmlns:A="foourn" xmlns:B1="barurn" xmlns:B2="barurn">
    <id>123</id>
    <A:name>asdf</A:name>
    <B2:name>asdf</B2:name>
    <B1:desc>foobar</B1:desc>
</foo>
```
and 
```xml
<foo xmlns="foourn" xmlns:A="foourn" xmlns:B1="barurn" xmlns:B2="barurn">
    <id>123</id>
    <name>asdf</name>
    <B1:name>asdf</B1:name>
    <B1:desc>foobar</B1:desc>
</foo>
```
should all be able to map into a struct/enum/etc. where the QName is associated for each field
```rust
struct Foo {
    id: String,    // QName: {foorun}id     |-----------------------------------------------------|
    name1: String, // QName: {foourn}name <- name conflict which is resolved through QName in XML |
    name2: String, // QName: {barurn}name <- name conflict in rust can be resolved through        |
                   //                       | serde(rename) or separate structs/enums or          |
                   //                       | whatever else we decide in design                   |
                   //                       |-----------------------------------------------------|
    desc: String   // QName: {barurn}desc
}
```
3. A conceptual solution to point 2 would be for the namespace to be associated with the struct/enum/etc. declaration and use serde(flatten) whenever there is a mixture of namespaces. The following example is what you provide:

```rust
#[xml(namespace = "foourn")]
struct Foo {
  id: String,   // QName: {foourn}id
  name: String, // QName: {foourn}name
  #[serde(flatten)]
  bar: Bar,     // QName: {foourn}bar
}

#[xml(namespace = "barurn")]
struct Bar {
  name: String, // QName: {barurn}name
  desc: String, // QName: {barurn}desc
}
```

The main point is that the `{foourn}` namespace is associated with all fields in Foo, but, `{foorun}bar` ceases to exist and instead pulls-in `{barurn}name` and `{barurn}desc` when `#[serde(flatten)]` is run.
4. Root elements are a special case for the deserializer. That is, the QName of the element does not have to match the struct/enum/etc.'s name+namespace association. As long as the QNames inside the xml element match the inside of the struct/enum/etc's  fields, it will be a valid deserialization. Given the example in point 3, if we consider `struct Foo` to associate with the root of an xml-document, the following should deserialize:

```xml
<this-can-be-any-namespace:this-can-be-any-tag xmlns:this-can-be-any-namespace="some-other-ns" xmlns="foourn" xmlns:F="foourn" xmlns:B1="barurn" xmlns:B2="barurn">
      <F:id>123</F:id>
      <name>asdf</name>
      <B1:name>asdf</B1:name>
      <B2:desc>foobar</B2:desc>
</this-can-be-any-namespace:this-can-be-any-tag>
```
This is a convention used by most mature xml libraries so we should follow it as well.
5. Sending extra metadata into the serde hooks (`fn serialize_struct` and `fn deserialize_struct`) like the expected QName is difficult because the function signatures take in the "name" which is the name after serde(rename) is applied. Sending metadata into the serde hooks would require either the user manually putting the metadata in `serde(rename)` which is something already done in quick-xml like `serde(rename = "@attribute")`, or, having a macro execute which could modify / add `serde(rename)` to these fields. Hopefully, this macro executes before `Serialize`/`Deserialize`.

Mingun:
@JOSEPHGILBY, sorry, missed your answer. Yes, you understood me perfectly.

One note about point 3: I realize, that unfortunately, assigning namespace information to a structs (rather that their fields) will not work with `#[serde(flatten)]`, because in that case the struct would be serialized using [`serialize_map`](https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_map) and we loose struct name with namespace info. Also, `#[serde(flatten)]` in general work poorly with non self-describing formats (i.e. formats that cannot say what type the content has in [`deserialize_any`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_any), and XML cannot). So it is better to avoid it usage at all.

riverar:
Just throwing in a random comment, apologies if this is out of place.

This proposal makes use of xml namespace prefixes. I'd like to caution that these namespace prefixes are not enforceable via xml schema and can be randomized.

In other words, the following will not work:

```rust
#[derive(Deserialize)]
#[namespace(prefixA, "https://schemas.example.com/foo/v1")]
struct Foo {
  id: String,
}
```

```xml
<root xmlns:prefixB="https://schemas.example.com/foo/v1">
  <prefixB:foo>
    <id>123</id>
  </prefixB:foo>
</root>
```

Instead, I propose the namespace prefixes be mapped/handled internally, letting users instead specify a namespace:

```rust
#[derive(Deserialize)]
#[namespace("https://schemas.example.com/foo/v1")]
struct Foo {
  id: String,
}
```

dralley:
@JOSEPHGILBY Do you intend to continue this work?

JOSEPHGILBY:
Unfortunately, I don't have the time.

PR #468, opened 2022-09-01
Start rewrite serde serialization tests

Mingun:
Our serde serializer a far from ideal, but rewriting it requires a very long commit history, as I think (I already have 20 commits and I suspect that I'll double that count...). Because I'm not a big fan of long PRs, I propose that changes, that rewrites tests without changing behavior. Next tests that I plan to add will test new corrected behavior, that is why this PR could look slightly unfinished.

Added a commit that ensures, that key name is always produces valid XML name for an element


dralley:
I'd like to try to not be rewriting the serde code in parallel - if there are relevant pieces of https://github.com/tafia/quick-xml/pull/441 which you would be willing to merge separately, let's try to do that.

Mingun:
For now I see that we are practically do not intersect -- it seems that serializer implementation does not depend much on the encoding, because it always writes UTF-8 strings. I rewriting serializer completely, so it will use `std::fmt::Write` (instead of `std::io::Write`) and not deal with encodings at all

Changelog.md
```diff
@@ -17,6 +17,7 @@
 ### Misc Changes
 
 - [#468]: Content of `DeError::Unsupported` changed from `&'static str` to `Cow<'static, str>`
+- [#468]: Ensure, that map keys could only types that serialized as primitives only
```
dralley, line 20: Ensure that map keys are restricted to types that can be serialized as primitives

dralley (approved):

PR #471, opened 2022-09-06
Fix #469 - parsing from buffered reader

### LINKED ISSUE #469: Problems with reading XML from buffering reader
Here is an example
```rust
fn main() {
    use quick_xml::events::Event;
    use std::io::BufReader;

    let xml = r###"<torrents>
    <torrent id="2" registred_at="2009.08.19 00:13:00" size="3">
    <title>Заголовок</title>
    <torrent hash="7096DB05CC2612B30079B26C94823DD8CA2A8156" tracker_id="3"/>
    <forum id="1"><![CDATA[<fo>]]></forum>
    <del/>
    <content><![CDATA[
        some content
    ]]></content>
    </torrent>
    </torrents>"###;

    let br = BufReader::with_capacity(32, xml.as_bytes());
    let mut reader = Reader::from_reader(br);
    reader.trim_text(true);

    let mut buf = Vec::new();
    let mut read_title = false;
    loop {
            Ok(Event::Start(e)) if e.name().as_ref() == b"title" => {
                read_title = true;
            }
            Ok(Event::Text(e)) => {
                if read_title {
                    read_title = false;
                    let escaped = &e.into_inner();
                    let text = reader.decoder().decode(escaped).unwrap_or_default();
                    let title = String::from(text);
                    print!("{title}");
                }
            }
            Ok(Event::Eof) => break, // exits the loop when reaching end of file
            Err(e) => panic!("Error at position {}: {:?}", reader.buffer_position(), e),
            _ => (), // There are several other `Event`s we do not consider here
        }
        buf.clear();
    }
}
```
when you run it you will crash with something like this
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/qxerr`
thread 'main' panicked at 'Error at position 299: EndEventMismatch { expected: "forum", found: "content" }', src/main.rs:39:23
Заголовок%         
```
if buffer capacity increase to 512 (see  let br = BufReader::with_capacity(32, xml.as_bytes());) the problem will gone. Replacing non Latin1 chars to latin1 ones will also solves the problem.

## ADDITIONAL INFO
It was not a problem before v0.23, so v0.22 parsed such files correctly

Mingun:
Fixes #469

The bug was introduced in f2b99f0966aa24febd3e58aaf8f2db1315ea47eb.

### DIFF
--- a/src/reader/async_tokio.rs
+++ b/src/reader/async_tokio.rs
@@ -377,7 +377,7 @@ impl<R: AsyncBufRead + Unpin> NsReader<R> {
 #[cfg(test)]
 mod test {
     use super::TokioAdapter;
-    use crate::reader::test::check;
+    use crate::reader::test::{check, small_buffers};
 
     check!(
         #[tokio::test]
@@ -385,6 +385,12 @@ mod test {
         read_until_close_async,
         TokioAdapter,
         &mut Vec::new(),
+        async, await
+    );
+
+    small_buffers!(
+        #[tokio::test]
+        read_event_into_async: tokio::io::BufReader<_>,
         async, await
     );
 }
--- a/src/reader/buffered_reader.rs
+++ b/src/reader/buffered_reader.rs
@@ -117,7 +117,7 @@ macro_rules! impl_buffered_source {
                     // somewhere sane rather than at the EOF
                     Ok(n) if n.is_empty() => return Err(bang_type.to_err()),
                     Ok(available) => {
-                        if let Some((consumed, used)) = bang_type.parse(available, read) {
+                        if let Some((consumed, used)) = bang_type.parse(buf, available) {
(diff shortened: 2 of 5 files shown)


Mingun:
I'm going to merge this soon and release 0.25.0 with the fix

dralley:
That's fine.  I had a brief look at it but hadn't did a comprehensive review yet.  My amount of free time varies week to week.

Some automated fuzzing would probably be effective at rooting out issues like this.

dralley:
It would be good to set guidelines for which bugs are severe enough to trigger a yank.   I'm not sure that it was worthwhile in this case

Mingun:
The problem in that that this bug can hide some content from reader if XML will be carefully constructed, which can have security implications. That is why I decided to yank that versions

PR #481, opened 2022-09-18
Remove use of `const fn`, add CI tests for MSRV

dralley:


Cargo.toml
```diff
@@ -10,13 +10,14 @@ repository = "https://github.com/tafia/quick-xml"
...
 license = "MIT"
+rust = "1.54"
 
 [dependencies]
 document-features = { version = "0.2", optional = true }
 encoding_rs = { version = "0.8", optional = true }
-serde = { version = "1.0", optional = true }
-tokio = { version = "1.21", optional = true, default-features = false, features = ["io-util"] }
-memchr = "2.5"
+serde = { version = "^1.0.100", optional = true }
```
adamreichold, line 18: As far as I know, [caret requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) are the default and I would suggest omitting the carets here for consistency with the other dependencies, otherwise it looks like there is a difference when there actually is not.

.github/workflows/rust.yml
```diff
@@ -10,6 +10,13 @@ jobs:
     - name: Check fmt
       run: cargo fmt -- --check
 
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@1.54.0
+    - run: cargo check --all-features
```
adamreichold, line 18: I think most crates make their MSRV promise for the default build or at least only for the default dependencies. `--all-features` will enable all optional dependencies which might be a harder promise to keep?

Cargo.toml
```diff
@@ -10,13 +10,14 @@ repository = "https://github.com/tafia/quick-xml"
 keywords = ["xml", "serde", "parser", "writer", "html"]
 categories = ["asynchronous", "encoding", "parsing", "parser-implementations"]
 license = "MIT"
+rust = "1.54"
```
adamreichold, line 13: The key is called spelled `rust-version`, c.f. https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

adamreichold (commented):
Thanks for going the extra mile and making things work even for 1.54 instead of 1.56!

.github/workflows/rust.yml
Mingun: Agree, I think, we should also test with `cargo check` and set that version as MSRV.

Cargo.toml
Mingun: Agree. Also, why `1.0.100`? Something required for us in that version was introduced? If so, then maybe `>=1.0.100`?

adamreichold: > Agree. Also, why `1.0.100`? Something required for us in that version was introduced? If so, then maybe `>=1.0.100`?

The default and caret requirements already mean `>=1.0.100, <2`.

On why `1.0.100`, serde is known for introducing new functionality in patch releases which suggests to specify a minimum patch release one has actually tested (or ideally even figuring out the true minimum patch release), c.f. https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

Mingun (changes requested):
So I think, that changes should be done:
- fix key in `Cargo.toml`
- set MSRV according to `cargo check`
- add a badge to readme

Also, maybe add a test for minimal versions? It seem that proper way to test it is using https://lib.rs/crates/cargo-minimal-versions.

Specifying dependencies with caret seems unnecessary, actually, their are already specified as caret dependencies. I think, that this part can be reverted if that will not fail the build.

Also, it seems that actually not all `const`s are unsupported on old `rustc`, maybe leave some of them? I don't insist, though.

Cargo.toml
dralley: Whoops, I was using the RFC, they must have changed it https://rust-lang.github.io/rfcs/2495-min-rust-version.html

Cargo.toml
dralley: >As far as I know, [caret requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) are the default and I would suggest omitting the carets here for consistency with the other dependencies, otherwise it looks like there is a difference when there actually is not.

The "crates" extension for VSCode was telling me that the unadorned version would prefer the oldest, whereas `^$version` selects the newest.  However, it is entirely possible that this is incorrect or at least misleading.  If it's documented that they are equivalent I have no reason to doubt that.

>Agree. Also, why 1.0.100? Something required for us in that version was introduced? If so, then maybe >=1.0.100?

I tried with serde 1.0 but it seemed to be using some nightly features that no longer exist.  There's nothing special about .100 other than that it is reasonably old and worked, and I didn't feel like bisecting 144 different versions at 1 AM.  I can do that now if we feel strongly about it.

Mingun: >  I can do that now if we feel strongly about it.

No, it's fine, but I would prefer if you add this consideration in comment in `Cargo.toml` for future readers

dralley: >There's nothing special about .100 other than that it is reasonably old and worked, and I didn't feel like bisecting 144 different versions at 1 AM. I can do that now if we feel strongly about it.

It turns out that no, out of all 144 versions, .100 **is** actually the minimum that works.  So my random selection was the actual right answer.  I should go buy a lottery ticket..

dralley:
>set MSRV according to cargo check

I assume by this you mean: repeatedly downgrade the toolchain until `cargo check` doesn't work anymore?  Or did you mean something else?

>add a badge to readme

I looked around at other projects to see how it is typically done, and it actually seems rather uncommon.  `serde`, `tokio`, `hyper`, `axum`, `kube-rs`, `rustls`, `thiserror`, `opencv-rust`, etc... none of them really use it.

Mingun:
> I assume by this you mean: repeatedly downgrade the toolchain until `cargo check` doesn't work anymore? Or did you mean something else?

I mean, do the same steps that you did to find MSRV as 1.54, but run `cargo check` instead of `cargo check --all-features` as, it seems, you did. You can also try https://github.com/foresterre/cargo-msrv

> I did find one for the [`regex`](https://raw.githubusercontent.com/rust-lang/regex/) crate, but looking at how the badge is configured compared to the others, I'm not sure that it's worth it.

`regex`, `serde`, `actix` just output static version of badge (i. e. MSRV is encoded in the link) generated by shields.io. This is fine for us as well.

dralley:
`cargo-msrv` said the minimum was 1.43.1, but I was having some issues with the MSRV check with that version due to the `document-features` dependency.  I'm not quite sure what was going on with that.

But in any case it works fine if I bump it up a bit higher to 1.46.0.

Mingun (approved):

vojtechkral:
Hey guys, I [ran](https://github.com/vojtechkral/bard/actions/runs/3154417560/jobs/5131987996) into this, can I trouble you to tag a release with this fix? :pray:

jonhoo:
Gentle nudge on this @Mingun — could we get a new release with this fix too?

Mingun:
Yes, I'll make a release soon. Actually, I plan to make a new release after merging #490, but because it introduces some significant changes, it will be better to make a intermediate release with features, wanted by community now. After all, releases does not happen much

Issue #483, opened 2022-09-26
Getting the entire text content as a String between two tags


econaxis:
In version 0.22.0, there was the function `read_text` that will return the entire text content as a String. Now, there is `read_to_end_into`. However, `read_to_end_into` cannot be used to recover the original text if we are using a BufReader. 

We only get the indices of the text, but there is no way to get the original text because BufRead API doesn't support that.

Any way that `read_text` could be added back to the latest version?

Mingun:
The `read_text` from 0.22 was never work as you expected, it only returned the first `Text` event (it even did not respect `CData` event) and expected `End` event just after it. That is why it was reworked and removed from buffered reader, because currently it is hard to implement it for it. I personally not much interested to implement it in near time, so feel free to submit a PR.

The current `read_text` _truly_ returns you _any text between tags_ (including comments and any other markup).

econaxis:
Ah ok. I can submit a PR with the function `read_text_to_string` that behaves like `read_to_end_into` but returns an actual String instead of just the range of indices. Is this both what we are thinking of?

PhilipDaniels:
Another vote here for this feature. I was about 99% done with my file parser when I ran into this :-( I see from the linked issues that impementing it would be non-trivial. In the meantime I think I will have to try and workaround it by reading the whole file into RAM first (an interesting afternoon awaits :-)). For my typical file sizes that should be OK, but it would be nice to keep the streaming API.

Thanks for the library, it is working well apart from this.

As an aside, I wonder if it is possible to improve the generated docs so that they mention that `read_text` is only available on slice-based readers?

PR #484, opened 2022-09-27
Implement read_to_end_to_string() to return the text content between two tags as a string

econaxis:
Implements the feature request from #483 .

### DIFF
--- a/src/reader/buffered_reader.rs
+++ b/src/reader/buffered_reader.rs
@@ -248,6 +248,55 @@ impl<'b, R: BufRead> XmlSource<'b, &'b mut Vec<u8>> for R {
     impl_buffered_source!();
 }
 
+use std::io::{Read};
+struct StoredReader<R> {
+    inner: R,
+    initial_offset: usize,
+    stored: Vec<u8>
+}
+
+impl<R: Read> Read for StoredReader<R> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        let result = self.inner.read(buf);
+        if let Ok(size) = &result {
+            self.stored.extend_from_slice(&buf[0..*size]);
+        }
+        result
+    }
+}
+impl<R> StoredReader<R> {
+    fn new(inner: R, initial_offset: usize) -> Self {
+        Self {
+            inner, initial_offset, stored: Vec::new()
+        }
+    }
+    fn get_stored_value(mut self, start: usize, end: usize) -> Vec<u8> {
+        // Pray that the start is 0 so we don't have to allocate
+        let start = start - self.initial_offset;
+        let end = end - self.initial_offset;
+
+        if start == 0 {
+            self.stored.truncate(end);
+            self.stored
+        } else {
+            unimplemented!()
+        }
+    }
+}
+impl<R: BufRead> BufRead for StoredReader<R> {
+    fn fill_buf(&mut self) -> io::Result<&[u8]> {
+        let result = self.inner.fill_buf();
+        if let Ok(x) = result {
+            self.stored.extend_from_slice(x);
+        };
+        result
+    }
+
+    fn consume(&mut self, amt: usize) {
+        self.inner.consume(amt)
+    }
+}
+
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
 /// This is an implementation for reading from a [`BufRead`] as underlying byte stream.
@@ -392,6 +441,21 @@ impl<R: BufRead> Reader<R> {
         Ok(read_to_end!(self, end, buf, read_event_impl, {
             buf.clear();
         }))
+    }
+
+    /// sample
+    pub fn read_text_to_string(&mut self, end: QName, buf: &mut Vec<u8>) -> Result<Vec<u8>> {
+        // Initialize a new XMLParser
+        let bp = self.buffer_position();
+        let reader = StoredReader ::new(
+        &mut self.reader, bp
+        );
+        let mut temp_reader: Reader<StoredReader<&mut R>> = Reader {
+            reader, parser: self.parser.clone()
+        };
+
+        let range = temp_reader.read_to_end_into(end, buf)?;
+        Ok(temp_reader.into_inner().get_stored_value(range.start, range.end))
     }
 }
 
--- a/src/reader/buffered_reader.rs
+++ b/src/reader/buffered_reader.rs
@@ -2,6 +2,8 @@
 //! underlying byte stream.
 
 use std::fs::File;
+use std::io::Read;
+
 use std::io::{self, BufRead, BufReader};
 use std::path::Path;
 
@@ -10,7 +12,7 @@ use memchr;
 use crate::errors::{Error, Result};
 use crate::events::Event;
 use crate::name::QName;
-use crate::reader::{is_whitespace, BangType, ReadElementState, Reader, Span, XmlSource};
+use crate::reader::{BangType, is_whitespace, read_to_string_impl, ReadElementState, Reader, Span, XmlSource};
 
 macro_rules! impl_buffered_source {
     ($($lf:lifetime, $reader:tt, $async:ident, $await:ident)?) => {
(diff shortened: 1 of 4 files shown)


src/reader/buffered_reader.rs
```diff
@@ -2,6 +2,8 @@
 //! underlying byte stream.
 
 use std::fs::File;
+use std::io::Read;
```
Mingun, line 5: Please put that in line 7

src/reader/buffered_reader.rs
```diff
@@ -241,6 +243,7 @@ macro_rules! impl_buffered_source {
 
 // Make it public for use in async implementations
 pub(super) use impl_buffered_source;
+use crate::reader::parser::Parser;
```
Mingun, line 246: That should be around line 15

src/reader/buffered_reader.rs
```diff
@@ -393,8 +396,14 @@ impl<R: BufRead> Reader<R> {
             buf.clear();
         }))
     }
+
+    /// Same as `read_to_end_into` but returns all the bytes read instead of the indices.
+    pub fn read_to_end_to_string(&mut self, end: QName, buf: &mut Vec<u8>) -> Result<Vec<u8>> {
```
Mingun, line 401: That function should be named `read_text_into` and have signature
```rust
pub fn read_text_into<'b>(&mut self, end: QName, buf: &'b mut Vec<u8>) -> Result<Cow<'b, str>>
```

I also expect the same set of tests as for `read_text` and implementation of `read_text_into_async`.

Both should have the similar documentation as `read_text` has. You can just copy existing documentation of `read_text` and edit the examples section. Also look at the differences in documentation of other paired functions and write the doc the same way.

src/reader/mod.rs
```diff
@@ -1930,3 +1933,68 @@ mod test {
...
+    fn get_stored_value(mut self, start: usize, end: usize) -> Vec<u8> {
+        // Pray that the start is 0 so we don't have to allocate
+        let start = start - self.initial_offset;
+        let end = end - self.initial_offset;
+
+        if start == 0 {
+            self.stored.truncate(end);
+            self.stored
+        } else {
+            unimplemented!()
```
Mingun, line 1983: That macros should not be used in the final code. If you have unreachable branches, you should use `unreachable!` macro and is OBLIGATORY mark it with the `// SAFETY` comment describing, why it is safe to call it

src/reader/mod.rs
```diff
@@ -1930,3 +1933,68 @@ mod test {
...
+        reader,
+        parser
+    };
+
+    let range = temp_reader.read_to_end_into(end, buf)?;
+    Ok(temp_reader.into_inner().get_stored_value(range.start, range.end))
+}
+
+/// Wraps an existing Reader and stores all the bytes read out of the `inner`.
+struct StoredReader<R> {
```
Mingun, line 1952: This struct seems a reimplementation of `BufReader` with an additional `initial_offset` field. Is it possible to replace it by
```rust
struct StoredReader<R> {
    inner: BufReader<R>,
    initial_offset: usize,
}
```
?

Also, I would like to see some internal tests of just this reader that it work as expected. For example, that it will not consume more than required.

src/reader/ns_reader.rs
```diff
@@ -526,6 +526,13 @@ impl<R: BufRead> NsReader<R> {
         // match literally the start name. See `Self::check_end_names` documentation
         self.reader.read_to_end_into(end, buf)
     }
+
+    /// Same as `read_to_end_into` but returns all the bytes read instead of the indices.
+    pub fn read_to_end_to_string(&mut self, end: QName, buf: &mut Vec<u8>) -> Result<Vec<u8>> {
+        let parser = self.parser.clone();
```
Mingun, line 532: `Parser` is a quite big structure. I would prefer to see implementation that does not copy it.

src/reader/mod.rs
econaxis: Yes that's right, but we want to use a user-provided storage buffer. We can't do this with BufReader as BufReader will allocate it's own buffer.

src/reader/buffered_reader.rs
econaxis: OK. I think we actually need two buffers. One buffer is for temporary storage of event data, and one buffer is to store the actual output string that will be returned to the user. 

Also I'm not sure why we would need a `Cow<str>` as we already have a buffer we can use to put the str in. The return type should be `Result<&'b str>`

So the signature will be:
```rust
pub fn read_text_into<'b>(&mut self, end: QName, event_buf: &mut Vec<u8>, output_buf: &'b mut Vec<u8>) -> Result<&'b str>
```

src/reader/mod.rs
```diff
@@ -1930,3 +1933,68 @@ mod test {
     pub(super) use check;
     pub(super) use small_buffers;
 }
+
+fn read_to_string_impl<R: BufRead>(starting_buffer_position: usize, reader: &mut R, parser: Parser, end: QName, buf: &mut Vec<u8>) -> Result<Vec<u8>> {
```
dralley, line 1937: What if anything is `buf` actually used for?  It seems that we copy the data but throw it away in favor of a buffer managed separately and internally, and without any way to reduce the size as the user-provided buffer allows.  

I'm not completely against the idea as I've had a similar one in the past, but I don't want to bolt on such a significant architectural element for this one particular feature without taking other aspects into consideration and thinking about whether it impacts the wisdom of our other API choices.

It also feels a bit like we're reimplementing https://gitlab.com/sequoia-pgp/sequoia/-/tree/main/buffered-reader

econaxis: Yes that's right. `buf` is entirely just to store the temporary contents and allow us to use `reader.read_event_into`. However, the same thing could be said about the current function `reader.read_to_end_into`. The `buf` parameter passed into that function is also just used for temporary holding.

The sequoia BufferedReader is similar to what we have. But the `StoredReader` that we have is much simpler as it just needs to store the read output into a user-provided buffer.

econaxis:
Implemented most of the changes. I will add the tests and documentation after we finalize the design and implementation of this feature.

src/reader/buffered_reader.rs
Mingun: `Cow` is required because it is return type of the `decode` method. You cannot use `from_utf8` to get a string. When feature `encoding` is activated, we can parse documents in almost all encodings supported by `encoding_rs` and we should decode bytes to string using XML encoding.

> One buffer is for temporary storage of event data

It is better to avoid this. That likely will require more work that just reuse existing parsing methods, but `output_buf` will contain all needed data for the events. This is why we need separate methods for buffered and borrowing readers and one way to implement that -- add a new method to `XmlSource`. But you're right, this is not necessarily required. You can try to find another way, the way how `async` is supported probably can give you some ideas

src/reader/mod.rs
```diff
@@ -1930,3 +1934,74 @@ mod test {
...
+            stored: buffer,
+        }
+    }
+    fn get_stored_value(mut self, start: usize, end: usize) -> Result<&'a str> {
+        // Pray that the start is 0 so we don't have to allocate
+        let end = end - self.initial_offset;
+
+        // Since read_to_end_impl will read all text from the current (initial) position to the end tag,
+        // the start must be equals to the initial position
+        assert_eq!(start, self.initial_offset);
```
Mingun, line 1988: `self.initial_offset` is not used at all, so it can be removed entirely

src/reader/mod.rs
```diff
@@ -1930,3 +1934,74 @@ mod test {
...
+        self.stored.truncate(end);
+
+        Ok(std::str::from_utf8(self.stored)?)
```
Mingun, line 1991: You should use `reader.decoder().decode(...)` here

src/reader/buffered_reader.rs
```diff
@@ -393,6 +397,26 @@ impl<R: BufRead> Reader<R> {
...
         }))
     }
+
+    /// Same as `read_to_end_into` but returns all the bytes read instead of the indices.
+    pub fn read_text_into<'b>(
+        &mut self,
+        end: QName,
+        buf: &mut Vec<u8>,
+        output_buf: &'b mut Vec<u8>,
+    ) -> Result<&'b str> {
```
Mingun, line 407: I would like to do without a temporary buffer, at least in the API, at best at all. Of course, that means that `read_to_end_into` cannot be used directly...

src/reader/mod.rs
```diff
@@ -1930,3 +1934,74 @@ mod test {
...
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        let result = self.inner.read(buf);
+        if let Ok(size) = &result {
+            self.stored.extend_from_slice(&buf[0..*size]);
+        }
+        result
+    }
+}
+
+impl<'a, R> StoredReader<'a, R> {
```
Mingun, line 1974: I think, that both methods can be inlined in their place of use. They a used only once and they both trivial

Mingun (changes requested):

econaxis:
I'm not sure if there's a way to implement this without building the parsing methods entirely from scratch, as it cannot use the built-in `read_to_end_into` function. I'm not too familiar with the codebase, but it does seem like a lot of work if we can't reuse the existing parsing structure to our advantage. 

Perhaps I can tackle this a bit later when I have more time, but I think I'll have to put my PR on hold for now.

PR #487, opened 2022-09-29
Micellaneous cleanups

dralley:


src/reader/mod.rs
```diff
@@ -458,8 +459,36 @@ pub struct Reader<R> {
     parser: Parser,
 }
 
-/// Builder methods
+/// Methods specialized to reading from a slice
+impl<'a> Reader<&'a [u8]> {
```
dralley, line 463: While it makes sense to have it in the other file, in practice it is somewhat annoying for these constructors to be in separate places.  Sticking them in the same file eliminates a couple of imports.

src/reader/mod.rs
```diff
@@ -1630,33 +1654,13 @@ mod test {
...
                 use pretty_assertions::assert_eq;
 
-                /// When `encoding` feature is enabled, encoding should be detected
-                /// from BOM (UTF-8) and BOM should be stripped.
-                ///
-                /// When `encoding` feature is disabled, UTF-8 is assumed and BOM
-                /// character should be stripped for consistency
-                #[$test]
-                $($async)? fn bom_from_reader() {
-                    let mut reader = Reader::from_reader("\u{feff}\u{feff}".as_bytes());
```
dralley, line 1640: `Reader::from_reader()` isn't going to be implemented for `Reader<&[u8]>`, but the tests in this macro get implemented for both.

dralley: To keep this test we would need to either specialize it fo each reader variant, or keep it as a unit-level test.

src/reader/mod.rs
```diff
@@ -458,8 +459,36 @@ pub struct Reader<R> {
...
+    /// Creates an XML reader from a string slice.
+    pub fn from_str(s: &'a str) -> Self {
+        // Rust strings are guaranteed to be UTF-8, so lock the encoding
+        #[cfg(feature = "encoding")]
+        {
+            let mut parser = Parser::default();
+            parser.encoding = EncodingRef::Explicit(UTF_8);
+            Self {
+                reader: s.as_bytes(),
+                parser: parser,
```
Mingun, line 473:
```suggestion
                parser,
```

src/reader/mod.rs
```diff
@@ -458,8 +459,36 @@ pub struct Reader<R> {
...
+                reader: s.as_bytes(),
+                parser: Parser::default(),
+            }
+        }
+    }
+}
+
+/// Public implementation-independent functionality
 impl<R> Reader<R> {
+    /// Configuration methods
```
Mingun, line 489: I'm not sure that doc comment is allowed here or has some meaning
```suggestion
    // Configuration methods
```

src/reader/mod.rs
Mingun: It seems to me more logically to include removal of this test to the PR which actually will remove `from_reader` from `Reader<&[u8]>`. For now we ensure some behavior via that test and before encoding will be reworked, it is better at least keep the behavior and verify it by tests.

Also, after removal of that test the comment for a `bom` test function become outdated

src/reader/slice_reader.rs
```diff
@@ -16,24 +14,10 @@ use crate::reader::{is_whitespace, BangType, ReadElementState, Reader, Span, Xml
 
 use memchr;
 
-/// This is an implementation for reading from a `&[u8]` as underlying byte stream.
-/// This implementation supports not using an intermediate buffer as the byte slice
-/// itself can be used to borrow from.
+/// This is an implementation of [`Reader`] for reading from a `&[u8]` as
```
Mingun, line 17: Actually, this is a docstring for a `Reader` struct and this reference would be to the same page. Not much useful as for me.

Mingun (changes requested):
I would like to see removal of tests alongside with the replacing tests or in the same PR that makes them unnecessary / useless. So I think it is better to include test removal to your PR about encodings, where it will be obvious why it is removed. Well, at least doc for other test should be updated accordingly.

About moving of constructor I agree with you -- I also found current situation somewhat uncomfortable

src/reader/mod.rs
dralley: I'm just trying to make that PR smaller and more approachable - both of these patches are from there so if you'd prefer to keep it in that PR then I'll just close this one entirely.

Mingun: Yes, I think then it is better to include them in that PR. I, too, struggled with the temptation to split  my next PR into pieces, but it seems to me that it will be better if they are all together, although there are 20 commits :)

Issue #368, opened 2022-03-11
Handling of control characters


dralley:
XML 1.0 disallows most kinds of control characters.  XML 1.1 allows some control characters, but only when they've been encoded as numeric character references.

There should probably be some way to encode these provided at minimum, or maybe prevent the writer from writing them.  I haven't gotten around to checking whether the reader checks for this yet, but libxml2 seems to throw a warning / error when it encounters raw illegal control characters.

https://en.wikipedia.org/wiki/Valid_characters_in_XML

Issue #354, opened 2022-01-29
Deserializing `$primitive=` unit variants doesn't work correctly yet


fwcd:
Currently, the deserializer doesn't recognize a primitive unit variant (i.e. one that uses the mechanism introduced in #304) correctly. Consider the following example (simplified from the unit tests):

```rust
enum Node {
  Unit,
  #[serde(rename = "$primitive=PrimitiveUnit")]
  PrimitiveUnit
}
```

Deserializing it with `from_str("PrimitiveUnit").unwrap()` throws
```
thread ... panicked at 'called `Result::unwrap()` on an `Err` value: Custom("unknown variant `PrimitiveUnit`, expected one of `Unit`, `$primitive=PrimitiveUnit`")`
```

Experimenting a bit with `EnumAccess`, tracking the primitive variants in a similar fashion to unflattened fields seems to do the trick on this test (see [this branch](https://github.com/fwcd/quick-xml/blob/unit-primitive-deserialization/src/de/var.rs#L52-L56)). However, it doesn't fix the error when the enum is nested in a struct, e.g. like this:

```rust
struct Wrapper {
  node: Node
}
```

In this case, deserializing `from_str(r#"<Wrapper node="PrimitiveUnit"/>"#)` yields us the same error. The problem is that the deserializer attempts to deserialize it as a string (apparently?) in the context of a `MapAccess`. I am not entirely sure how the Serde machinery works here, but we might have to keep track of primitive variants even outside of direct enum contexts, which could be quite a bit harder to do.

rrichardson:
@fwcd thanks for investigating this.   

I would like to have a go at fixing this.   Has anyone started researching it? Is there a simple fix, or would there likely have to be some major machinery changes? It seems like that a node would just have to "remember" its primitive designation, which, considering that there is a PrimitiveNodeType, that part should be done.  So is it simply a matter of dicing up the element differently for the PrimitiveNodeType?

rrichardson:
I found https://github.com/tafia/quick-xml/pull/304  

It appears that we just need to execute the opposite step for `de`, perhaps in a similar fashion as what `de` does for `$unflatten` in structs.   

I'll take a stab at implementing it.

Mingun:
`$primitive=` prefix will be removed in #490, so the problem disappears. The current course for mapping:
```rust
enum E {
  Unit
}
struct AnyName {
  element: E,

  #[serde(rename = "@attribute")] // prepend @
  attribute: E,

[...]

PR #490, opened 2022-10-01
Rewrite serializer

### LINKED ISSUE #252: Help serializing empty attributes
I have an externally defined XML standard I'm trying to write to which includes requiring empty attribute values.  As an example, I need to write the following:

```
<TIM A="" D="4">
```

Where my `A` attribute needs to be written with an empty value, and `D` needs a constant value.  Is it possible to achieve this with the crate using the serde serialize feature?  Ideally, I'd like to use an `Option` like this:

```
#[derive(Serialize, Debug)]
struct Tim {
   a: Option<String>,
   b: u8,
}
```
But when I do that, the `A` attribute isn't serialized at all.  I'm not sure that the crate would be able to decide whether or not to serialize the attribute, so I'm guessing I'll need `serialize_with`.  Also tried that, but haven't been able to get it to write out an empty attribute value.  The closest I've gotten is writing out `<TIM A=" " D="4"/>` but in my case, the space in the `A` attribute value is invalid.

### LINKED ISSUE #280: Vec<String> inside of struct serializes incorrectly with serde

### LINKED ISSUE #287: How to have tag on each item?
I have struct `Person`.

```rust
#[derive(serde::Serialize)]
pub struct Person {
    name: String,
    age: i32,
}
```
and I want to parse it to XML string:

```xml
<Person><name>Foo</name><age>20</age></Person>
```

But I can't find a way to do it in `quick-xml`. https://docs.rs/quick-xml/0.22.0/quick_xml/se/fn.to_string.html

```rust
#[derive(serde::Serialize)]
pub struct Person {
    name: String,
    age: i32,
}

fn main() {
    let person = Person {
        name: "Foo".to_string(),
        age: 20,
    };

    let data_xml = quick_xml::se::to_string(&person).unwrap(); // how to have open-closed tag ??
    println!("{}", data_xml);
}
```

My current workaround is to use `serde_xml_rs`. 

```rust
let data_xml = serde_xml_rs::to_string(&person).unwrap();
```
Is there a way to do this? So that I don't need multiple XML dependencies.

Thanks for quick-xml

### LINKED ISSUE #343: Serialization of Hashmap duplicate outer tag in the inners elements. Deserialization has no such problem and fails on serialized xml.
```toml
quick-xml = {version = "0.23.0-alpha2", features = ["serialize"]}
```

```rust
use std::collections::HashMap;

use quick_xml::de::from_str;
use quick_xml::se::to_string;

#[derive(Debug, Serialize, Deserialize)]
struct Users {
    users: HashMap<String, User>,
}
#[derive(Debug, Deserialize, Serialize, PartialEq)]
struct Max(u16);

#[derive(Debug, Deserialize, Serialize, PartialEq)]
struct User {
    max: Max,
}

fn main() {
    let xml = "<Users><users><roger><max>10</max></roger></users></Users";
    let users: Users = from_str(xml).unwrap();
    println!("des : {:?}", users);
    let xml2 = to_string(&users).unwrap();
    println!("res : {:?}", xml);
    println!("res : {:?}", xml2);
}
```

output :

```
des : Users { users: {"roger": User { max: Max(10) }} }
res : "<Users><users><roger><max>10</max></roger></users></Users"
res : "<Users><users><roger><users><max>10</max></users></roger></users></Users>"
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("missing field `max`")', src/main.rs:27:41
```

### LINKED ISSUE #346: Serialization of enums adds extra tags

### LINKED ISSUE #361: How to pretty-print with serde?
Hello,

I have the code below; when it runs, it results in the following output:
```xml
seralized:
<osm><node visible="true" timestamp="2022-02-17T13:55:16.184553291Z"><tag k="key1" v="value1"/><tag k="key2" v="value2"/></node></osm>
```

I would like to print it in a prettyfied way, i.e. with indention and line breaks following the XML structure. How can that be done?

```rust
use serde::Serialize;
use quick_xml::se;
use chrono::{ DateTime, Utc };

#[derive(Serialize, Clone, Debug)]
#[serde(rename = "tag")]
#[serde(rename_all = "lowercase")]
struct OsmTag {
    k: String,
    v: String,
}

#[derive(Serialize, Clone, Debug)]
#[serde(rename = "node")]
#[serde(rename_all = "lowercase")]
struct OsmNode {
    #[serde(skip_serializing_if = "Option::is_none")]
    id: Option<u64>,
    visible: bool,
    #[serde(skip_serializing_if = "Option::is_none")]
    version: Option<u64>,
    #[serde(skip_serializing_if = "Option::is_none")]
    changeset: Option<u64>,
    timestamp: DateTime<Utc>,
    #[serde(skip_serializing_if = "Option::is_none")]
    user: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none")]
    uid: Option<u64>,
    #[serde(skip_serializing_if = "Option::is_none")]
    lat: Option<f64>,
    #[serde(skip_serializing_if = "Option::is_none")]
    lon: Option<f64>,
    tag: Vec<OsmTag>,
}

impl Default for OsmNode {
    fn default() -> OsmNode {
        OsmNode {
            id: None,
            visible: true,
            version: None,
            changeset: None,
            timestamp: Utc::now(),
            user: None,
            uid: None,
            lat: None,
            lon: None,
            tag: Vec::new()
        }
    }
}

#[derive(Serialize, Debug)]
#[serde(rename = "osm")]
#[serde(rename_all = "lowercase")]
struct OsmNodeContainer {
    node: OsmNode,
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut on = OsmNode { ..Default::default() };
    let t1 = OsmTag { k: String::from("key1"), v: String::from("value1") };
    let t2 = OsmTag { k: String::from("key2"), v: String::from("value2") };
    on.tag.push(t1.clone());
    on.tag.push(t2.clone());
    let onc = OsmNodeContainer { node: on.clone(), };
    let ser = se::to_string(&onc)?;
    println!("seralized:");
    println!("{}", ser);

    Ok(())
}
```

### LINKED ISSUE #429: Serializing $value enum variant doesn't work
Cargo.toml:
```toml
[package]
name = "quick-xml-test"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
serde = { version = "1.0.139", features = ["derive"] }
quick-xml = { version = "0.23.0", features = ["serialize" ] }
```

src/main.rs:
```rust

#[derive(Debug, Deserialize, Serialize)]
enum State {
    A,
    B,
    C,
}

#[derive(Debug, Deserialize, Serialize)]
struct StateOuter {
    #[serde(rename = "$value")]
    state: State,
}

#[derive(Debug, Deserialize, Serialize)]
pub struct Root {
    state: StateOuter,
}

impl Root {
    pub fn new() -> Self {
        Self { state: StateOuter { state: State::B } }
    }
}

fn main() {
    let xml = "<root><state>B</state></root>";
    println!("Deserialized: {:?}", quick_xml::de::from_str::<Root>(&xml));

    let root = Root::new();
    println!("Serialized: {:?}", quick_xml::se::to_string(&root));
}
```

Output:
```
Deserialized: Ok(Root { state: StateOuter { state: B } })
Serialized: Ok("<Root><state><B/></state></Root>")
```

Expected output:
```
Deserialized: Ok(Root { state: StateOuter { state: B } })
Serialized: Ok("<Root><state>B</state></Root>")
```

As you can see deserialization works, but serialization does not.

### LINKED ISSUE #430: Use $value and $unflatten together
I have some XML that looks like:
```xml
<root>
  <state>B</state>
</root>
```

which I would like to represent in rust as:
```rust
enum State {
  A,
  B,
  C,
}

struct Root {
  state: State,
}
```

I have tried combining `$unflatten` and `$value` to make this happen, but it doesn't seem to work:

Cargo.toml:
```toml
[package]
name = "quick-xml-test"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
serde = { version = "1.0.139", features = ["derive"] }
quick-xml = { version = "0.23.0", features = ["serialize" ] }
```

src/main.rs:
```rust
mod working {

    #[derive(Debug, Deserialize, Serialize)]
    enum State {
        A,
        B,
        C,
    }

    #[derive(Debug, Deserialize, Serialize)]
    struct StateOuter {
        #[serde(rename = "$value")]
        state: State,
    }

    #[derive(Debug, Deserialize, Serialize)]
    pub struct Root {
        state: StateOuter,
    }
}

mod not_working {

    #[derive(Debug, Deserialize, Serialize)]
    #[serde(rename = "$value")]
    enum State {
        A,
        B,
        C,
    }

    #[derive(Debug, Deserialize, Serialize)]
    pub struct Root {
        #[serde(rename = "$unflatten=state")]
        state: State,
    }
}

fn main() {
    let xml = "<root><state>B</state></root>";
    println!("Deserialized (working): {:?}", quick_xml::de::from_str::<working::Root>(&xml));
    println!("Deserialized (not working): {:?}", quick_xml::de::from_str::<not_working::Root>(&xml));
}
```

Output:
```
Deserialized (working): Ok(Root { state: StateOuter { state: B } })
Deserialized (not working): Err(Custom("unknown variant `state`, expected one of `A`, `B`, `C`"))
```

Is it possible to do without introducing the extra `StateOuter` struct?

Mingun:
This PR rewrites serializer and fixes the following issues:

- Fixes #252
- Fixes #280
- Fixes #287
- Fixes #343
- Fixes #346
- Invalidates #354 (but it still should be checked that it works aligned with the serializer)
- Fixes #361
- Partially addresses #368
- Fixes #429
- Fixes #430

Notable changes:
- Removed `$unflatten=` prefix, as it not used anymore
- Removed `$primitive=` prefix, as it not used anymore
- `$value` special name split into two special names `#text` and ~`#any`~ `#content`. @dralley, please check especially, is the documentation clear about the differences?
- Change how attributes is defined. Now attribute should have a name started with `@` character. Without that fields serialized as elements and can be deserialized only from elements. Deserialization from element or from attribute into one field is not possible anymore. It's a weird wish anyway


src/errors.rs
```diff
@@ -188,6 +188,15 @@ pub mod serialize {
         /// An attempt to deserialize to a type, that is not supported by the XML
         /// store at current position, for example, attempt to deserialize `struct`
         /// from attribute or attempt to deserialize binary data.
+        ///
+        /// Serialized type cannot be represented in an XML due to violation of the
+        /// XML rules in the final XML document. For example, attempt to serialize
+        /// a `HashMap<{integer}, ...>` would cause this error because [XML name]
+        /// cannot start from a digit or a hyphen (minus sign). The same result
+        /// would occur if map key is a complex type that serialized not as
```
dralley, line 196: that serialized not as => that cannot be serialized as

src/se/key.rs
```diff
@@ -50,6 +50,18 @@ impl<W: Write> Serializer for QNameSerializer<W> {
         ))
     }
 
+    /// We cannot store anything, so the absence of a unit and presence of it
```
dralley, line 53: Do you mean, we cannot store nothing?

tests/test.rs
```diff
@@ -155,690 +155,6 @@ fn test_clone_reader() {
     assert!(matches!(cloned.read_event().unwrap(), End(_)));
 }
 
-#[cfg(feature = "serialize")]
```
dralley, line 158: The `seq` and `struct` tests feel rather abstract to me, because `seq` tests only look at the number of elements without checking the contents, and `struct` tests only check the serialization of one element at a time.

This is more of an integration test which covers a much more complex scenario involving multiple elements on the same struct and nesting, which the other tests don't appear to cover thoroughly.  

I'd personally be more comfortable if it was moved to `serde_roundtrip` but if there's a strong case that it doesn't add value I would listen

Changelog.md
```diff
@@ -14,6 +14,11 @@
 
 ### Bug Fixes
 
+- [#490]: Ensure, that serialization of map keys always produces valid XML names.
```
dralley, line 17: Comma is unnecessary

dralley:
>+6,885 −1,917 

I might need a few days for this

src/se/key.rs
Mingun: Yes, I mean that following consistency rules we cannot store any sign that some data is present here. The rule is "name of the type does not matter". We never use type name when serialize data (except for inferring top-level element name). Because this is a serializer for a element/attribute name, we must write something, and because there is no data to write, we return error here

tests/test.rs
Mingun: Yes, this is possible to add a special test for `seq + non-trivial struct` to existing `seq` tests to test an integration. I don't want to restore this specific test because it is too big and redundant to test such a scenario. If it falls, it will be difficult to find an error.

rrichardson:
How does one achieve the previous behavior of enums where the enum value would be wrapped in the enum type tag? 

```rust 
enum Wat {
    Foo,
    Bar,
    Baz
} 
struct Test {
    wat: Wat, 
} 
``` 
results in 
```xml
<Test>
    <Foo/> 
</Test> 
```
and if you add `rename=#text`  to `wat: Wat` then you get 
```xml
<Test>
    Foo
</Test> 
```

Is the intention that we wrap the enum in a wrapper struct if we want type tags for the enum?  

```xml 
<Test> 
    <Wat>Foo</Wat> 
</Test> 
```

I guess this makes it consistent with the way lists are handled, and also consistent with the philosophical standpoint that enum is simply one of N possible values,  and not a container. 

We should probably document this change, though.

rrichardson:
It appears that the `#text` and `#any` attributes are *required* if you want to successfully deserialize an enum from a struct.  I'm assuming that this is because the outer type tags of the enum are eaten, so we need to tell the deserializer exactly how to parse the contents. 

```rust 
mod try_enum {
    use quick_xml::{de::from_str, se::to_string};
    
    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
    enum WatEnum {
        #[serde(rename="FOO")]
        Foo,
        #[serde(rename="BAR")]
        Bar,
        #[serde(rename="BAZ")]
        Baz
    }

    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
    struct Wat {
        wat: WatEnum
    }

    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
    struct Wat2 {
        #[serde(rename="#any")]
        #[serde(rename="BAZ")]
        Baz
    }

    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
    struct Wat {
        wat: WatEnum
    }

    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
    struct Wat2 {
        #[serde(rename="#any")]
        wat: WatEnum
    }

    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
    struct Wat3 {
        #[serde(rename="#text")]
        wat: WatEnum
    }

    #[test]  // This fails
    fn round_trip_enum_none() {
        let tst = Wat { wat: WatEnum::Bar };
        let data = to_string(&tst).unwrap();
        println!("{:?}", &data);
        let tst_again: Wat = from_str(data.as_str()).unwrap();
        assert_eq!(&tst, &tst_again);
    }

    #[test]  // This works
    fn round_trip_enum_any() {
        let tst = Wat2 { wat: WatEnum::Bar };
        let data = to_string(&tst).unwrap();
        println!("{:?}", &data);
        let tst_again: Wat2 = from_str(data.as_str()).unwrap();
        assert_eq!(&tst, &tst_again);
    }
 
    #[test]  // this works
    fn round_trip_enum_text() {
        let tst = Wat3 { wat: WatEnum::Bar };
        let data = to_string(&tst).unwrap();
        println!("{:?}", &data);
        let tst_again: Wat3 = from_str(data.as_str()).unwrap();
        assert_eq!(&tst, &tst_again);
    }
}

File: src/escapei.rs [hunk 5]
```diff
@@ -112,7 +112,14 @@ fn _escape<F: Fn(u8) -> bool>(raw: &str, escape_chars: F) -> Cow<str> {
             b'\'' => escaped.extend_from_slice(b"&apos;"),
             b'&' => escaped.extend_from_slice(b"&amp;"),
             b'"' => escaped.extend_from_slice(b"&quot;"),
-            _ => unreachable!("Only '<', '>','\', '&' and '\"' are escaped"),
+
+            b'\t' => escaped.extend_from_slice(b"&#9;"),
+            b'\r' => escaped.extend_from_slice(b"&#10;"),
+            b'\n' => escaped.extend_from_slice(b"&#13;"),
+            b' ' => escaped.extend_from_slice(b"&#32;"),
```
dralley, line 119: Why escape these characters?

x09, x0D, x0A are all explicitly allowed as per https://en.wikipedia.org/wiki/Valid_characters_in_XML#Non-restricted_characters

And space is obviously OK

src/escapei.rs
```diff
@@ -112,7 +112,14 @@ fn _escape<F: Fn(u8) -> bool>(raw: &str, escape_chars: F) -> Cow<str> {
             b'\'' => escaped.extend_from_slice(b"&apos;"),
             b'&' => escaped.extend_from_slice(b"&amp;"),
             b'"' => escaped.extend_from_slice(b"&quot;"),
-            _ => unreachable!("Only '<', '>','\', '&' and '\"' are escaped"),
+
+            b'\t' => escaped.extend_from_slice(b"&#9;"),
```
dralley, line 116: Leave a comment about the reasoning behind why these characters (sometimes) need to be escaped as it isn't immediately obvious

tests/serde-se.rs
```diff
@@ -285,28 +285,41 @@ mod with_root {
...
+            </root>");
     serialize_as!(flatten_struct:
         FlattenStruct {
             nested: Nested { float: 42.0 },
             string: "answer",
         }
-        => r#"<root><float>42</float><string>answer</string></root>"#);
+        => "<root>\
+                <float>42</float>\
+                <string>answer</string>\
```
dralley, line 1080: What is the mechanism for having a different expectation of this test vs. the previous one when the structs are declared the same?

README.md
```diff
@@ -131,37 +131,6 @@ struct Foo {
 }
```
dralley, line 131: Lines 118 and 129 need to be fixed to remove `$value` and probably document the new way instead.

Line 120 needs to be removed as no longer relevant.

src/se/mod.rs
```diff
@@ -154,6 +147,16 @@ pub enum QuoteLevel {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+macro_rules! forward {
```
dralley, line 150: Some documentation here would be nice

dralley:
Have you run benchmarks before and after?

Mingun:
@dralley, no. The goal is to make a correct and consistent implementation first. There a no reason to worry about performance while the old implementation is just incorrect and inconsistent

Mingun:
> It appears that the `#text` or `#any` attributes are _required_ if you want to successfully deserialize an enum from a struct.

While in this PR this can be true, that is not desired behavior. This PR is just about serializer, I plan to continue work on deserializer part in next PRs. The final goal -- a consistent and composable implementation:
- _consistent_: the same types should be serialized and deserialized similarly independently of their location in the XML. The change of representation should depend only on explicitly added attributes in Rust code, not of a way, how types are composed
- _composable_: if type `A` includes in type `B`, then in XML representation markup for type `A` should be included in markup of type `B` following by predictable rules
- the most used patterns should be covered without (or with minimal) additional attributes. The such patterns are:
  - elements and attributes
  - XML `xs:choice` mapped to Rust `enum`
  - text content of nodes
  - deserialization with `deserialize_any` should give a result that close to result of converting XML to JSON (because internally `serde` uses JSON-like type model). This is important, because in many cases `serde` uses `deserialize_any` under the hood

dralley:
>There a no reason to worry about performance while the old implementation is just incorrect and inconsistent

I'm not worried about performance, correctness is of course more important, it would just be good to know if it deserves some attention afterwards

Changelog.md
```diff
@@ -23,6 +23,11 @@
 - [#490]: Removed `$unflatten=` special prefix for fields for serde (de)serializer, because:
   - it is useless for deserializer
   - serializer was rewritten and does not require it anymore
+- [#490]: Special name `$value` for a field has been replaced by two special names:
+  - `#text` if you want to map field to text content only. No markup is expected
+    (but text can represent a list as defined by `xs:list` type)
+  - `#any` that is close to the previous `$value` behavior. Use it when you
+    want to map field to any content of the field, text or markup
```
dralley, line 35: An example might also be nice - or a link to documentation where you can read more.

src/se/content.rs
```diff
@@ -0,0 +1,614 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/// A serializer used to serialize content of the element. It does not write
+/// surrounding tags.
+///
+/// This serializer does the following:
+/// - primitives (booleans, numbers and strings) serialized as naked strings
+/// - `None` does not write anything
+/// - sequences serialized without delimiters
```
dralley, line 30: An example would be nice, similar to the others

src/se/content.rs
```diff
@@ -0,0 +1,614 @@
...
+/// - units (`()`) and unit structs are not supported
+/// - structs and maps are not supported
+/// - unit variants serialized as self-closed `<${variant}/>`
+/// - tuple variants serialized as sequences where each is wrapped in
+///   `<${variant}>...</${variant}>`
+/// - struct variants serialized wrapped `<${variant}>...</${variant}>`
+///
+/// Difference of this serializer from [`SimpleTypeSerializer`] in how sequences
```
dralley, line 38: The difference between this serializer and

src/se/content.rs
```diff
@@ -0,0 +1,614 @@
...
+                    <val>42</val>\
+                </Struct>");
+
+        /// Special field name `#text` should be serialized as a text content
+        mod text {
+            use super::*;
+            use pretty_assertions::assert_eq;
+
+            err!(map: BTreeMap::from([("#text", 2), ("_3", 4)])
+                => Unsupported("serialization of map is not supported in `#any` field"));
```
dralley, line 608: of map types

src/se/element.rs
```diff
@@ -0,0 +1,1455 @@
...
+
+    #[inline]
+    fn end(self) -> Result<Self::Ok, Self::Error> {
+        <Self as SerializeTuple>::end(self)
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/// A serializer for struct variants, which serialized wrapped in `<${tag}>...</${tag}>`.
```
dralley, line 282: Suggestion => which serializes the struct contents inside of wrapping tags

src/se/element.rs
```diff
@@ -0,0 +1,1455 @@
...
+        serialize_as!(char_apos: '\'' => "<root>&apos;</root>");
+        serialize_as!(char_quot: '"' => "<root>&quot;</root>");
+
+        serialize_as!(str_non_escaped: "non-escaped string" => "<root>non-escaped string</root>");
+        serialize_as!(str_escaped: "<\"escaped & string'>" => "<root>&lt;&quot;escaped &amp; string&apos;&gt;</root>");
+
+        err!(bytes: Bytes(b"<\"escaped & bytes'>") => Unsupported("`serialize_bytes` not supported yet"));
+
+        serialize_as!(option_none: Option::<&str>::None => "<root/>");
+        serialize_as!(option_some: Some("non-escaped string") => "<root>non-escaped string</root>");
```
dralley, line 610: Consider testing serialization of `Option(&str)` with escapable input

src/se/element.rs
```diff
@@ -0,0 +1,1455 @@
...
+                any!(enum_unit: Enum::Unit => "<Unit/>");
+                err!(enum_unit_escaped:
+                    BTreeMap::from([("#any", Enum::UnitEscaped)])
+                    => Unsupported("character `<` is not allowed at the start of an XML name `<\"&'>`"));
+
+                any!(newtype: Newtype(42) => "42");
+                any!(enum_newtype: Enum::Newtype(42) => "<Newtype>42</Newtype>");
+
+                // Note that sequences of primitives serialized without delimiters!
+                any!(seq: vec![1, 2, 3] => "123");
```
dralley, line 1123: This seems rather dangerous and unintentional?  I'm missing something

src/se/element.rs
dralley: And similar variants

src/se/mod.rs
```diff
@@ -100,6 +101,56 @@ pub fn to_string<S: Serialize>(value: &S) -> Result<String, DeError> {
...
 
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/// Defines which characters would be quoted in [`Text`] events and attribute
+/// values.
+///
+/// [`Text`]: Event::Text
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum QuoteLevel {
+    /// Performs quoting, escape all characters that could have special meaning
```
dralley, line 108: Is "quoting" an official name for this?  I've only ever seen this referred to as escaping.

src/se/simple_type.rs
```diff
@@ -0,0 +1,396 @@
+//! Contains Serde `Serializer` for XML [simple types] [as defined] in the XML Schema.
+//!
+//! [simple types]: https://www.w3schools.com/xml/el_simpletype.asp
+//! [as defined]: https://www.w3.org/TR/xmlschema11-1/#Simple_Type_Definition
```
dralley, line 4: If any of this can sensibly be shared, it might make sense in `crate::escapei`

tests/serde-se.rs
```diff
@@ -547,6 +547,397 @@ mod without_root {
...
+    mod with_indent {
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        macro_rules! serialize_as {
+            ($name:ident: $data:expr => $expected:literal) => {
+                #[test]
+                fn $name() {
+                    let mut ser = Serializer::new(String::new());
+                    ser.indent(' ', 2);
```
dralley, line 560: Expected output for other tests is also indented, despite not having an indentation set?  One example: https://github.com/tafia/quick-xml/pull/490/files#diff-452d3372c13dbffe2706a379e9b64cb6d109a116360ca59ea6efe3ee8c403b08R411

I'm confused whether those tests ought to not expect any indentation.

dacut: FWIW: For the tests expecting proper indentation, I've found [indoc](https://docs.rs/indoc/latest/indoc/) to be useful. It strips leading spaces up to the indentation point but no further, e.g.:
```rust
indoc! {r#"
    <foo>
        <bar>42</bar>
    </foo>"# }
```

becomes
```rust
<foo>
    <bar>42</bar>
</foo>
```

src/se/content.rs
```diff
@@ -39,23 +39,36 @@ macro_rules! write_primitive {
...
+/// indent, sequence of strings become one big string with additional content
+/// and it would be impossible to distinguish between content of the original
+/// strings and inserted indent characters.
+pub struct ContentSerializer<'i, W: Write> {
     pub writer: W,
     /// Defines which XML characters need to be escaped in text content
     pub level: QuoteLevel,
+    /// Current indentation level
+    pub(super) indent: Indent<'i>,
+    /// If `true`, then current indent will be written before write the content,
```
dralley, line 52: before writing the content

dralley, line 52:
```suggestion
/// If `true`, then current indent will be written before writing the content,
```

src/se/content.rs
```diff
@@ -39,23 +39,36 @@ macro_rules! write_primitive {
...
+/// impossible to distinguish between them. Even worse, when serializing with
+/// indent, sequence of strings become one big string with additional content
+/// and it would be impossible to distinguish between content of the original
+/// strings and inserted indent characters.
+pub struct ContentSerializer<'i, W: Write> {
     pub writer: W,
     /// Defines which XML characters need to be escaped in text content
     pub level: QuoteLevel,
+    /// Current indentation level
+    pub(super) indent: Indent<'i>,
```
dralley, line 51: The documentation should probably make the distinction between `write_indent: false` and `indent: Indent::None` more clear

src/de/key.rs
```diff
@@ -0,0 +1,458 @@
...
+/// only UTF-8 compatible characters (usually only ASCII characters).
+#[inline]
+fn decode_name<'n>(name: QName<'n>, decoder: Decoder) -> Result<Cow<'n, str>, DeError> {
+    let local = name.local_name();
+    Ok(decoder.decode(local.into_inner())?)
+}
+
+/// A deserializer for xml names of elements and attributes.
+///
+/// Used for deserialize values from:
```
dralley, line 31:
```suggestion
/// Used for deserializing values from:
```

src/de/key.rs
```diff
@@ -0,0 +1,458 @@
...
+/// - attribute names (`<... name="..." ...>`)
+/// - element names (`<name>...</name>`)
+///
+/// Converts name to the identifier string by following rules:
```
dralley, line 35:
```suggestion
/// Converts a name to an identifier string using the following rules:
```

src/de/key.rs
```diff
@@ -0,0 +1,458 @@
...
+///
+/// - if that is an [`attribute`] name, put `@` to the identifier
```
dralley, line 37: /// - if it is an [`attribute`] name, put `@` in front of the identifier

src/de/key.rs
```diff
@@ -0,0 +1,458 @@
...
+/// - put the decoded [`local_name()`] of a name to the identifier
```
dralley, line 38: I'm not actually sure what you're trying to say here

src/de/key.rs
```diff
@@ -0,0 +1,458 @@
...
+/// The final identifier looks like `[@]local_name`
+/// (where `[]` means optional element).
+///
+/// The deserializer also supports deserializing names as other primitive types:
+/// - numbers
+/// - booleans
+/// - unit (`()`) and unit structs
+/// - unit variants of the enumerations
+///
+/// Because `serde` does not define, on which side type conversion should be
```
dralley, line 49: Comma not needed

src/de/key.rs
```diff
@@ -0,0 +1,458 @@
...
+/// performed and because [`Deserialize`] implementation for that primitives
```
dralley, line 50: Comma should go after "performed"

src/de/mod.rs
```diff
@@ -1,4 +1,84 @@
 //! Serde `Deserializer` module
+//!
+//! # Difference between `#text` and `#any` special names
+//!
+//! quick-xml supports two special names for fields -- `#text` and `#any`. Although
+//! they may seem the same, there is a distinction. Two different names required
```
dralley, line 6:
```suggestion
//! they may seem the same, there is a distinction. Two different names is required
```

src/de/mod.rs
```diff
@@ -1,4 +1,84 @@
...
+//! mostly for serialization, because quick-xml should know how you want to
+//! serialize certain constructs, that could be represented differently in the
```
dralley, line 8: which could be represented through XML in multiple different ways

dralley:
@Mingun Done

tests/test.rs
Mingun: Added new tests

src/se/key.rs
Mingun: Slightly reworded this comment to make it more obvious

src/escapei.rs
Mingun: Commend added

Changelog.md
Mingun: Added link to documentation

src/se/content.rs
Mingun: Added an example: `[1, 2, 3]` would be serialized as `123`

src/se/element.rs
Mingun: The point here is to test serialization of `Some` variant. Using a string just a most straightforward choice for a type. I think it is enough to test only the one variant of composition here, we should assume that others will work too, otherwise it is unclear where we should stop and believe that composition works

src/se/element.rs
Mingun: Unfortunately, that is. Yes, this can be unexpected, but this is consistent behaviour, and so it is preditable and can be explained. A possible solution here is to detect such a situation and return an error during serialization, I left the corresponding TODO for further improvements on line 55 in `content.rs`:
https://github.com/tafia/quick-xml/pull/490/files#diff-db7f64b5e8c9079f9ad5065b66f9c766bf2c32fed0d619ab2acc9e949248e166R55

src/se/simple_type.rs
Mingun: This is hard to say now which of them would be useful, can be re-exported later in follow-up PRs.

src/se/content.rs
Mingun: Explained:
> Current indentation level. Note, that `Indent::None` means that there is no indentation at all, but `write_indent == false` means only, that indent writing is disabled in this instantiation of `ContentSerializer`, but child serializers should have access to the actual state of indentation.

README.md
Mingun: Fixed

Changelog.md
```diff
@@ -23,6 +23,11 @@
 - [#490]: Removed `$unflatten=` special prefix for fields for serde (de)serializer, because:
   - it is useless for deserializer
   - serializer was rewritten and does not require it anymore
+- [#490]: Special name `$value` for a field has been replaced by two special names:
```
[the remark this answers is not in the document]
Mingun: Added

src/de/key.rs
Mingun: I try say here that XML name stored in struct's `name` field would be converted to the "identifier" (in terms of serde's `Deserializer::deserialize_identifier`) of the Rust field name using the described rules. So Rust (actually, serde) name of field for an attribute `name` should be `@name` and for element `<name>` should be `name`.

src/se/mod.rs
Mingun: I don't known. Replaced to "escaping"

Mingun:
@dralley , almost all your concerns is addressed, I'll make final check and push tomorrow

src/se/mod.rs
Mingun: Added:
> Implements serialization method by forwarding it to the serializer created by the helper method `Serializer::ser`.

Mingun:
I tried to explain all the subtle points, and also expanded the documentation on the differences between the special names, providing it with examples. In the process I realized, that `#any` is a better to rename to `#content`, because "any" can give a misconception, that it can handle any rust type, which is not true (`struct`s with fields cannot be handled). "any" here referred to the XML side. In any case, the new name seems better.

Please use <kbd>[Compare](https://github.com/tafia/quick-xml/compare/8511d6f7c948e8704e41a42721540fca3ab596e9..21d61615c65a7bda987c3dab429f2b78de7dcc31)</kbd> button to see the difference.

I think, it is almost good now. @dralley, could you please check that all language in the new docs are correct and clear.

dralley:
Changes look good, particularly `#any` -> `#content`, it's a much more intuitive name.

A few additional typos but nothing significant

src/de/mod.rs
```diff
@@ -1,4 +1,209 @@
...
+//! let obj = AnyName { field: vec![1, 2, 3] };
+//! let xml = to_string(&obj).unwrap();
+//! assert_eq!(xml, "<AnyName>1 2 3</AnyName>");
+//!
+//! let object: AnyName = from_str(&xml).unwrap();
+//! assert_eq!(object, obj);
+//! ```
+//!
+//! ## `#content`
+//! Representation of primitive types in `#content` does not differ from they representation
```
dralley, line 55:
```suggestion
//! Representation of primitive types in `#content` does not differ from their representation
```

src/de/mod.rs
```diff
@@ -1,4 +1,209 @@
...
+//! in `#text` field. The difference is how sequences are serialized. `#content` serializes
+//! each sequence item as a separate XML element. The name of that element is taken
+//! from serialized type, and because only `enum`s provide such name (their variant
+//! name), only them should be used in such field.
```
dralley, line 59:
```suggestion
//! name), only they should be used for such fields.
```

alternately: only enums should be used for such fields.

src/de/mod.rs
```diff
@@ -1,4 +1,209 @@
...
+//! let xml = to_string(&obj).unwrap();
+//! // Note, that types that does not produce their own tag are written as is!
+//! assert_eq!(xml, "<AnyName>123</AnyName>");
+//!
+//! let object: AnyName = from_str("<AnyName>123</AnyName>").unwrap();
+//! assert_eq!(object, AnyName { field: vec![123] });
+//!
+//! // `1 2 3` is mapped to a single `usize` element
+//! // It is impossible to deserialize list of primitives to such field
+//! from_str::<AnyName>("<AnyName>1 2 3</AnyName>").unwrap_err();
```
dralley, line 128: It might be worth calling this out as unintuitive behavior.

dralley (approved):
Small suggestions, otherwise looks good.

Mingun:
After carefully reading examples, that I wrote for **Difference between `#text` and `#content` special names** section I understand, that actually the behavior of `#content` field mostly replicates behavior of `$value` field. In order to keep compatibility with other XML serde crates, I ended up to continue to use `$value` instead of `#content`, although the latter name much more liking for me.

In just pushed set of changes the real changes are only in `de/mod.rs` and `Changelog.md`, the others contains technical renames (any -> value, to keep linkage between a special name and those names).

As usual, use the <kbd>[Compare](https://github.com/tafia/quick-xml/compare/21d61615c65a7bda987c3dab429f2b78de7dcc31..af6f35cff913210c6d5e17f4cd5f865d56e2f8f4)</kbd> button to see the difference.

I _really_ hope that this will be the last time where things changed so much. @dralley, please can you review once again?

dralley (approved):
Substitution looks fine.  Only question I have is whether `#text` ought to be `$text` considering that `#content` is now `$value`.

Mingun:
I choose name `#text` because, as I remember, I saw that many HTML/XML DOM engines uses it as an internal name for a text node, and even visualize it with that name. But well, consistence is better that ideal namings. So I rename it to `$text` and merge

seanpianka:
What is `#[serde(rename="#text")]` and why does it make enum deserialization work? Is there any pre-existing docs that explain what this is or why it's needed?

noguxun:
For anyone who needs a quick example of adding indentation when do serializing 

```
    use serde::Serialize;
    use quick_xml::se::Serializer;

    #[derive(Debug, PartialEq, Serialize)]
    struct Struct {
        question: String,
        answer: u32,
    }

    let data = Struct {
        question: "The Ultimate Question of Life, the Universe, and Everything".into(),
        answer: 42,
    };

    let mut buffer = String::new();
    let mut ser = Serializer::new(&mut buffer);
    ser.indent(' ', 2);

    data.serialize(ser).unwrap();

    println!("{buffer}");
```

PR #496, opened 2022-10-26
Allow `&#0;` values

sashka:
Do not return error on parsing preperly escaped zero values, which are perfectly valid. Add simple test to check it.

### DIFF
--- a/src/escapei.rs
+++ b/src/escapei.rs
@@ -10,8 +10,6 @@ use pretty_assertions::assert_eq;
 /// Error for XML escape / unescape.
 #[derive(Debug)]
 pub enum EscapeError {
-    /// Entity with Null character
-    EntityWithNull(Range<usize>),
     /// Unrecognized escape symbol
     UnrecognizedSymbol(Range<usize>, String),
     /// Cannot find `;` after `&`
@@ -31,11 +29,6 @@ pub enum EscapeError {
 impl std::fmt::Display for EscapeError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
-            EscapeError::EntityWithNull(e) => write!(
-                f,
-                "Error while escaping character at range {:?}: Null character entity not allowed",
-                e
-            ),
             EscapeError::UnrecognizedSymbol(rge, res) => write!(
                 f,
                 "Error while escaping character at range {:?}: Unrecognized escape symbol: {:?}",
@@ -183,7 +176,7 @@ where
                 let pat = &raw[start + 1..end];
                 if pat.starts_with('#') {
                     let entity = &pat[1..]; // starts after the #
-                    let codepoint = parse_number(entity, start..end)?;
+                    let codepoint = parse_number(entity)?;
                     unescaped.push_str(codepoint.encode_utf8(&mut [0u8; 4]));
                 } else if let Some(value) = named_entity(pat) {
                     unescaped.push_str(value);
@@ -1690,15 +1683,12 @@ fn named_entity(name: &str) -> Option<&str> {
     Some(s)
 }
 
-fn parse_number(bytes: &str, range: Range<usize>) -> Result<char, EscapeError> {
+fn parse_number(bytes: &str) -> Result<char, EscapeError> {
     let code = if bytes.starts_with('x') {
         parse_hexadecimal(&bytes[1..])
     } else {
         parse_decimal(bytes)
     }?;
-    if code == 0 {
-        return Err(EscapeError::EntityWithNull(range));
-    }
     match std::char::from_u32(code) {
         Some(c) => Ok(c),
         None => Err(EscapeError::InvalidCodepoint(code)),
@@ -1745,6 +1735,8 @@ fn test_unescape() {
     assert_eq!(unescape("&lt;test&gt;").unwrap(), "<test>");
     assert_eq!(unescape("&#x30;").unwrap(), "0");
     assert_eq!(unescape("&#48;").unwrap(), "0");
+    assert_eq!(unescape("&#0;").unwrap(), "\0");
+    assert_eq!(unescape("&#x0;").unwrap(), "\0");
     assert!(unescape("&foo;").is_err());
 }
 


dralley:
I'm not sure this is true

 http://lists.xml.org/archives/xml-dev/201211/msg00014.html 
http://lists.xml.org/archives/xml-dev/201211/msg00012.html

sashka:
According to https://www.w3.org/TR/REC-xml/ the value of `&#x0;` has never stated invalid.

dralley:
Do your internal needs require handling a lot of documents with that issue?  Can you speak about the source & purpose of said documents?  The "API concerns" I have to imagine are related to C-strings (with null termination) and not strictly relevant to Rust.  However, there would be nothing preventing Rust from reading such documents, even though we shouldn't allow them to be output.

sashka:
There are lots of huge CPU hardware-related documents (dozens of gigabytes), and there are properly escaped nulls. Minor example of the same  kind is Apple's keyboard layout XML. 

In both cases I have to read those XMLs exclusively by Rust, and I never write XML, so in my practice I would never face error on attempt to write NULL. 

On the other hand, I may put read-write for NULL under the feature flag to allow this use case when a user (like me or me engineers) sure they needs that.

dralley:
@Mingun (and @tafia), what are your thoughts about allowing nominally specification-breaking, but widespread behavior like this?

Is this something that might be acceptable under a feature flag?

Mingun:
Yes, I actually does not understand, why XML forbid character references to some characters in the first place, and I'm glad to allow them. This ability under a feature flag would be welcome.

@sashka, can you add a properly documented feature flag (say, `allow-restricted-chars`) and add tests for all restricted characters (as defined in the rule `RestrictedChar` in https://www.w3.org/TR/xml11/#charsets) + `&#0;`? If we actually already allow some of that ranges (I looks like we allow), then I think we should forbid they usage without a feature flag to be standard compatible. Then rename `EntityWithNull` to `RestrictedChar`, add a doc reference to the standard. We use XML 1.1

dralley:
Let's open an issue and collect a bit more information before we merge anything - for instance, what do other XML libraries do.

It looks like [nokogiri](https://github.com/sparklemotion/nokogiri/issues/1429) (perhaps via libxml2) ignores NUL character references during rather than raising a fatal error.  However, attempting to write XML with those characters will raise a fatal error.

Are the NUL characters in these documents "load-bearing"?  Would it be dangerous to just strip them out like this?

Mingun:
The following behaviors are possible:
1. emit hard error (= stop parsing / writing and return error)
2. silently skip invalid characters (or at least log them, but we currently does not depend on any logger)
3. accept invalid characters

Each behavior could be applied to a raw and an escaped form of characters

I think, that behavior 2 is an anti-pattern and we should not allow it. I cannot imagine a situation where you would need to silently change the data that you operate. If you for whatever reason need that, then you can done that manually in `Writer` API and we can add an API in serde `Serializer` for that.

The behaviors 1 and 3 could be chosen either via feature flag, or via a run-time option. I'll agree with any of the options or even a combination of them.

I didn't have situations where I needed a dynamic choice of this option, so the feature flag should be enough in the first stage and easier to implement.

The choice between support only escaped form, or both raw and escaped form also could be leaved as implementation details in the first stage. Ah, and I suspect, that currently we forbid only escaped form of `&#0;`, but not raw `\0`.

There is also a technical question for a feature flag: it should allow restricted chars, forbidden without them, or it should restrict chars, allowed without them? In both cases the flag is additive is some sense, but I think that the allowing behavior is better. Of course, we could have both flags, but does not allow to enable them simultaneously, but I think that is overkill for now.

Any thoughts?

dralley:
It appears that C#'s dotnet XML library has a flag for this: https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlreadersettings.checkcharacters?view=net-6.0

Same for writing: https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlwritersettings.checkcharacters?view=net-6.0

>Surprisingly, the only restricted character which is actually restricted is the escaped NULL:

I'm not surprised given I filed https://github.com/tafia/quick-xml/issues/368.  It's something we needed to improve anyways.  Do you have any avenue to report the invalid XML?  Obviously that's not an immediate solution, I am just curious.

sashka:
I'll keep all the work for `allow-restricted-chars` right there: https://github.com/sashka/quick-xml/tree/allow-restricted-chars

>  Do you have any avenue to report the invalid XML? Obviously that's not an immediate solution, I am just curious.
No, vendors don't care about general XML validity as soon as their XML files are not for public distribution, and their own tools process their own files perfectly.

dralley:
> The following behaviors are possible:
> 
> * emit hard error (= stop parsing / writing and return error)
> * silently skip invalid characters (or at least log them, but we currently does not depend on any logger)
> * accept invalid characters
> 
> Each behavior could be applied to a raw or an escaped form of characters.
> 
> I think, that behavior 2 is an anti-pattern and we should not allow it. I cannot imagine a situation where you would need to silently change the data that you operate. If you for whatever reason need that, then you can done that manually in Writer API and we can add an API in serde Serializer for that.

Agree that we shouldn't skip over invalid data.  Seems like a terrible approach.

> The behaviors 1 and 3 could be chosen either via feature flag, or via a run-time option. I'll agree with any of the options or even a combination of them.

> I didn't have situations where I needed a dynamic choice of this option, so the feature flag should be enough in the first stage and easier to implement.
> 
> The choice between support only escaped form, or both raw and escaped form also could be leaved as implementation details in the first stage. Ah, and I suspect, that currently we forbid only escaped form of &#0;, but not raw \0.

> There is also a technical question for a feature flag: it should allow restricted chars, forbidden without them, or it should restrict chars, allowed without them? In both cases the flag is additive is some sense, but I think that the allowing behavior is better. Of course, we could have both flags, but does not allow to enable them simultaneously, but I think that is overkill for now.
> 
> Any thoughts?

An initial conservative approach could be to just provide an advanced version of the various escape / unescape functions which would allow you to configure the behavior either way.  I think I would prefer that approach over a feature flag.

Agreed that we're still missing the checks for these characters in the raw form and that we ought to have them.

Reading through the MS docs again, it appears that `CheckCharacters` applies only to character entities.  Names and raw text still need to be completely valid and contain no illegal characters no matter what - and it's checked always.  @sashka I assume all of the illegal characters are at least encoded as character references and not present in raw form?  If that's the case then copying that approach would be sufficient.

sashka:
@dralley, yes, it sounds right for me. I'm working on proper reading mechanics now.

dralley:
@sashka Do you plan to continue with this?

dralley:
Closing as a duplicate of #750

PR #501, opened 2022-10-29
Fix deserialization of top-level sequences of enums

### LINKED ISSUE #500: Sequences of enums at root level are not parsed correctly
The following test is failed on second `assert_eq!` (`cargo test --all-features -- issue500`):
```rust
#[test]
fn issue500() {
    #[derive(Debug, Deserialize, Serialize, PartialEq)]
    struct TagOne {}

    #[derive(Debug, Deserialize, Serialize, PartialEq)]
    struct TagTwo {}

    #[derive(Debug, Deserialize, Serialize, PartialEq)]
    enum Tag {
        TagOne(TagOne),
        TagTwo(TagTwo),
    }

    #[derive(Debug, Deserialize, Serialize, PartialEq)]
    struct Root {
        #[serde(rename = "$value", default)]
        data: Vec<Tag>,
    }

    let data: Root = from_str("\
        <root>\
            <TagOne></TagOne>\
            <TagTwo></TagTwo>\
            <TagOne></TagOne>\
        </root>\
    ").unwrap();

    assert_eq!(
        data,
        Root {
            data: vec![
                Tag::TagOne(TagOne {}),
                Tag::TagTwo(TagTwo {}),
                Tag::TagOne(TagOne {}),
            ],
        }
    );

    let data: Vec<Tag> = from_str("\
        <TagOne></TagOne>\
        <TagTwo></TagTwo>\
        <TagOne></TagOne>\
    ").unwrap();

    assert_eq!(
        data,
        vec![
            Tag::TagOne(TagOne {}),
            Tag::TagTwo(TagTwo {}),
            Tag::TagOne(TagOne {}),
        ]
    );
}
```
Output:
```diff
---- issue500 stdout ----
thread 'issue500' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 [
     TagOne(
         TagOne,
     ),
>    TagTwo(
>        TagTwo,
>    ),
     TagOne(
         TagOne,
     ),
 ]
```

Mingun:
Fixes #500. The behavior of top-level sequence actually very simple -- try to deserialize until `Eof` is received, which is the only stop marker.

After the fix in the first commit it became obvious, that some refactoring could be done to put highly coupled code together, which is done in the next 2 commits.


src/de/seq.rs
```diff
@@ -96,29 +82,7 @@ where
...
-}
-
-#[cfg(feature = "overlapped-lists")]
-impl<'de, 'a, R> Drop for TopLevelSeqAccess<'de, 'a, R>
-where
-    R: XmlRead<'de>,
-{
-    fn drop(&mut self) {
-        self.de.start_replay(self.checkpoint);
+        Ok(Self { de })
```
dralley: Does this need to remain fallible?  ie. can it just return `Self` instead of `Result<Self, DeError>`

src/de/map.rs
```diff
@@ -549,6 +551,72 @@ where
...
+    /// The tag name is stored inside (`b"tag"` for that example)
+    Include(BytesStart<'de>), //TODO: Need to store only name instead of a whole tag
+    /// A `SeqAccess` interested in tags with any name, except explicitly listed.
+    /// Excluded tags are used as struct field names and therefore should not
+    /// fall into a `$value` category
+    Exclude(&'static [&'static str]),
+}
+
+impl<'de> TagFilter<'de> {
+    fn is_suitable(&self, start: &BytesStart, decoder: Decoder) -> Result<bool, DeError> {
```
dralley, line 610: I believe this is now dead code?

src/de/seq.rs
```diff
@@ -74,20 +74,6 @@ where
...
     /// Deserializer used to deserialize sequence items
     de: &'a mut Deserializer<'de, R>,
-    /// Filter that determines whether a tag is a part of this sequence.
-    ///
-    /// When feature `overlapped-lists` is not activated, iteration will stop
-    /// when found a tag that does not pass this filter.
-    ///
-    /// When feature `overlapped-lists` is activated, all tags, that not pass
-    /// this check, will be skipped.
-    filter: TagFilter<'de>,
```
dralley, line 84: In fact, since this was the only use of `TagFilter` and it is now deleted, I believe all of that is dead code 

https://github.com/tafia/quick-xml/search?q=tagfilter

dralley: update: no, github search just sucks and doesn't show all uses.  if I checkout the branch locally, `MapValueSeqAccess` is obviously still using it.

src/de/map.rs
dralley: update: no, github search just sucks

src/de/seq.rs
Mingun: Yeah, missed that. Will fix in few seconds...

Issue #474, opened 2022-09-09
Merge text and CDATA events in serde deserializer


Mingun:
CDATA elements cannot contain sequence `]]>`. When that sequence is appeared in the data, it should be split into two pieces and each piece should be put in their own CDATA container:
```
]]>
```
become
```xml
<![CDATA[]]]>
<![CDATA[]>]]>
```
or
```xml
<![CDATA[]]]]>
<![CDATA[>]]>
```
Currently in serde deserializer only one CDATA event processed at time, that means, that deserialization
```xml
<root>
  <string><![CDATA[]]]]><![CDATA[>]]></string>
</root>
```
into
```rust
struct AnyName {
  string: String,
}
```
would fail or wrongly return `]]` instead of `]]>`.

To fix that we should merge CDATA events, that there are some ambiguities that should be investigated:
1. should we merge CDATA and text events:
   ```xml
   <![CDATA[one]]>two
   ```
   should return `onetwo`?
   > Judging from [that](https://stackoverflow.com/questions/13979582/php-simplexml-decoding-entities-in-cdata) and [that](https://dorward.uk/www/comments-cdata/), we should do that
2. should we ignore comments between CDATA events? Between CDATA and text events?
   ```xml
   <![CDATA[one]]><!--comment--><![CDATA[two]]>
   ```
   should return `onetwo`?
   ```xml
   <![CDATA[one]]><!--comment-->two
   ```
   should return `onetwo`?
   Currently all comments are skips at very early stage and deserializer sees
   ```xml
   <![CDATA[one]]><!--comment--><![CDATA[two]]>
   ```
   as
   ```xml
   <![CDATA[one]]><![CDATA[two]]>
   ```
3. should we ignore processing instructions between CDATA events? Between CDATA and text events?
   ```xml
   <![CDATA[one]]><?pi?><![CDATA[two]]>
   ```
   should return `onetwo`?
   ```xml
   <![CDATA[one]]><?pi?>two
   ```
   should return `onetwo`?
   Currently all processing instructions are skips at very early stage and deserializer sees
   ```xml
   <![CDATA[one]]><?pi?><![CDATA[two]]>
   ```
   as
   ```xml
   <![CDATA[one]]><![CDATA[two]]>
   ```
4. should we ignore whitespaces between CDATAs? Between CDATA and text?
   ```xml
   <![CDATA[one]]>
   <![CDATA[two]]>
   ```
   should return `onetwo`?
   ```xml
   <![CDATA[one]]>
[...]

PR #504, opened 2022-11-06
simple example with sub-tree reading in separate function

ultrasaurus:
This is an attempt to create a simple example that mirrors my use case I think it is a common use case so could be useful to others once I get it to work

Questions:
- I thought there should be a way to read from a buffer in a function that takes the element (where the goal is for the function to process that sub-tree of the XML) -- I think I understand why the code is wrong, just not sure how to make it right!
- I found it unexpected that `<Element/>` syntax didn't work with code that works properly for `<Element></Element>`

Appreciate any and all feedback!  I'm relatively new to Rust, so please feel free to comment on whatever would make this more idiomatic or consistent with library goals of "almost zero-copy" and "easy on memory allocation"

Thank you!

### DIFF
--- a/examples/read_nodes.rs
+++ b/examples/read_nodes.rs
@@ -2,7 +2,9 @@
 
 use quick_xml::events::{BytesStart, Event};
 use quick_xml::reader::Reader;
+use std::borrow::Cow;
 use std::collections::HashMap;
+use std::convert::Infallible;
 use std::str;
 
 // Code doesn't work if DefaultSettings node is constructed like this:
@@ -40,7 +42,6 @@ impl Translation {
         reader: &mut Reader<&[u8]>,
         element: BytesStart,
     ) -> Result<Translation, quick_xml::Error> {
-        use std::borrow::Cow;
         let mut tag = Cow::Borrowed("");
         let mut lang = Cow::Borrowed("");
         // let (tag, lang) =
@@ -85,14 +86,30 @@ fn main() -> Result<(), quick_xml::Error> {
         match reader.read_event_into(&mut buf)? {
             Event::Start(element) => match element.name().as_ref() {
                 b"DefaultSettings" => {
+                    // note: real app would handle errors with good defaults or halt program with nice message
+                    // This example illustrates transforming an attribute's key and value into a String
                     settings = element
                         .attributes()
                         .map(|attr_result| {
-                            let a = attr_result.unwrap();
-                            (
-                                str::from_utf8(a.key.as_ref()).unwrap().to_string(),
-                                a.unescape_value().unwrap().to_string(),
-                            )
+                            match attr_result {
+                                Ok(a) => {
+                                    let key = String::from_utf8(a.key.local_name().as_ref().into())
+                                        .or_else(|err| {
+                                            dbg!("unable to read key in DefaultSettings attribute {:?}, utf8 error {:?}", &a, err);
+                                            Ok::<String, Infallible>(String::new())
+                                        })
+                                        .unwrap();
+                                    let value = a.decode_and_unescape_value(&reader).or_else(|err| {
+                                            dbg!("unable to read key in DefaultSettings attribute {:?}, utf8 error {:?}", &a, err);
+                                            Ok::<Cow<'_, str>, Infallible>(std::borrow::Cow::from(""))
+                                    }).unwrap().to_string();
+                                    (key, value)
+                                },
+                                Err(err) => {
+                                     dbg!("unable to read key in DefaultSettings, err = {:?}", err);
+                                    (String::new(), String::new())
+                                }
+                            }
                         })
                         .collect();
                     println!("settings: {:?}", settings);
--- a/examples/read_nodes.rs
+++ b/examples/read_nodes.rs
@@ -14,17 +14,17 @@ const XML: &str = r#"
   </DefaultSettings>  
   <Localization>
     <Translation Tag="HELLO" Language="ja">
-        <Text>こんにちは</Text>
+      <Text>こんにちは</Text>
     </Translation>
     <Translation Tag="BYE" Language="ja">
-        <Text>さようなら</Text>
+      <Text>さようなら</Text>
     </Translation>
-     <Translation Tag="HELLO" Language="es">
-         <Text>Hola</Text>
-     </Translation>
-     <Translation Tag="BYE" Language="es">
-         <Text>Adiós</Text>
-     </Translation>
+    <Translation Tag="HELLO" Language="es">
+      <Text>Hola</Text>
+    </Translation>
+    <Translation Tag="BYE" Language="es">
+      <Text>Adiós</Text>
+    </Translation>
   </Localization>
 "#;
 
@@ -58,7 +58,7 @@ impl Translation {
         match event {
             Event::Text(ref e) => {
                 text = e.unescape()?;
-                println!("text node content: {}", text);
+                dbg!("text node content: {}", &text);
             }
             _ => (),
         }
--- a/examples/read_nodes.rs
+++ b/examples/read_nodes.rs
@@ -45,7 +45,7 @@ impl Translation {
         let mut lang = Cow::Borrowed("");
         // let (tag, lang) =
         for attr_result in element.attributes() {
-            let a = attr_result.unwrap();
+            let a = attr_result?;
             match a.key.as_ref() {
                 b"Language" => lang = a.unescape_value()?,
                 b"Tag" => tag = a.unescape_value()?,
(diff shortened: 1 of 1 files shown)


examples/read_nodes.rs
```diff
@@ -0,0 +1,115 @@
...
+                b"Tag" => tag = a.unescape_value()?,
+                _ => (),
+            }
+        }
+        let mut text = Cow::Borrowed("");
+        let mut element_buf = Vec::new();
+        let event = reader.read_event_into(&mut element_buf)?;
+        match event {
+            Event::Text(ref e) => {
+                text = str::from_utf8(e.as_ref())?.into();
```
Mingun, line 60: That code has two problems:
- it will not work if you decide to enable `encoding` feature
- it does not decode entities, such as `&lt;`

It is better to use `e.unescape()?`:
```suggestion
                text = e.unescape()?;
```

examples/read_nodes.rs
```diff
@@ -0,0 +1,115 @@
...
+                println!("text node content: {}", text);
+            }
+            _ => (),
```
Mingun, line 63: Probably it would be better to return error here, because if you encounter, for example, a `Start` element here, then all further parsing will be broken.

You also can get an `Empty` or `End` event here, which corresponds to `<Text/>` and `<Text></Text>` pieces accordingly. I suppose you'd like to handle such situations correctly.

Maybe for an example this is not necessary, but at least a comment about that would be worth.

Translation also could be inside of CDATA section, so processing of `CData` event could be valuable.

examples/read_nodes.rs
```diff
@@ -0,0 +1,115 @@
...
+    <Translation Tag="BYE" Language="ja">
+        <Text>さようなら</Text>
+    </Translation>
+     <Translation Tag="HELLO" Language="es">
+         <Text>Hola</Text>
+     </Translation>
+     <Translation Tag="BYE" Language="es">
+         <Text>Adiós</Text>
+     </Translation>
+  </Localization>
```
Mingun, line 28: Need to use consistent indents, for example 2 spaces

examples/read_nodes.rs
```diff
@@ -0,0 +1,115 @@
...
+                _ => (),
+            }
+        }
+        let mut text = Cow::Borrowed("");
+        let mut element_buf = Vec::new();
+        let event = reader.read_event_into(&mut element_buf)?;
+        match event {
+            Event::Text(ref e) => {
+                text = str::from_utf8(e.as_ref())?.into();
+                println!("text node content: {}", text);
```
Mingun, line 61: It could be better to use `dbg!` macro instead, because it shows line number

examples/read_nodes.rs
```diff
@@ -0,0 +1,115 @@
...
+            },
+
+            Event::Eof => break, // exits the loop when reaching end of file
+            _ => (),             // There are `Event` types not considered here
+        }
+    }
+    println!("translations...");
+    for item in translations {
+        println!("{} {} {}", item.lang, item.tag, item.text);
+    }
```
Mingun, line 113: For the sake of clarity it could be useful to use `assert_eq!` to show what you expect to get, so the reader can see the result without running the code

examples/read_nodes.rs
```diff
@@ -0,0 +1,115 @@
...
+                b"DefaultSettings" => {
+                    settings = element
+                        .attributes()
+                        .map(|attr_result| {
+                            let a = attr_result.unwrap();
+                            (
+                                str::from_utf8(a.key.as_ref()).unwrap().to_string(),
+                                a.unescape_value().unwrap().to_string(),
+                            )
+                        })
```
Mingun, line 96: Again, for an example it could be better to show the correct way to get string from attribute names and values. [`reader.decoder()`](https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Reader.html#method.decoder) should be used to obtain decoder, that can be used to decode content of key and value.

For decoding value you also can use a [`decode_and_unescape_value()`](https://docs.rs/quick-xml/latest/quick_xml/events/attributes/struct.Attribute.html#method.decode_and_unescape_value) method.

It is also better to replace `.unwrap()`s with `?` operator, because method signature allows that.

examples/read_nodes.rs
```diff
@@ -0,0 +1,115 @@
+// example that separates logic for reading different top-level nodes of xml tree
+
+use quick_xml::events::{BytesStart, Event};
+use quick_xml::reader::Reader;
+use std::collections::HashMap;
+use std::str;
+
+// Code doesn't work if DefaultSettings node is constructed like this:
+//   <DefaultSettings Language="es" Greeting="HELLO"/>
```
Mingun, line 9: This is because 
```xml
<DefaultSettings Language="es" Greeting="HELLO"/>
```
would produce `Empty` event. You could set [`expand_empty_elements`](https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Reader.html#method.expand_empty_elements) to `true` to always get `Start/End` pair and do not worry about it.

Mingun (changes requested):
Thanks for contribution. Since examples can be seen as a source for copy-paste, then I pointed to places that should be fixed (or at least noted their problems) so that the copy-pastors do not make mistakes.

> I thought there should be a way to read from a buffer in a function that takes the element (where the goal is for the function to process that sub-tree of the XML) -- I think I understand why the code is wrong, just not sure how to make it right!

quick-xml only produces events that reflects XML constructs, as they was encountered in document, and do not produce tree. It have some helpers to deal with tree structure, like [`read_to_end`](https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Reader.html#method.read_to_end), but generally you should view the flow of events as flat and manually track the depth.

So your approach in the example is correct, it only would be better to check in the end of `new_from_element`, or at call site of this function, that you are get corresponding `End` event to defend from broken XML documents, by calling `read_to_end(QName(b"Translation"))` (of course, it is better to supply `QName` from start event, so to not hardcode element name).

Also I could note, that probably your use-case better solved using serde API. The structure of your document is not complicated and it is mapped to serde model well. Probably that note also would be useful in the beginning of that example.

examples/read_nodes.rs
ultrasaurus: changed to use `reader.read_text` -- hoping that handles CDATA?

ultrasaurus:
@Mingun I think I've addressed all of your awesome feedback.  Could you take another look?

examples/read_nodes.rs
Mingun: No, it will return content of the source together with markup -- so for
```xml
<Text><![CDATA[cdata]]></Text>
```
it will return `<![CDATA[cdata]]>` instead of `cdata`

Nowadays there is not good method to do that in one-line, I'll plan to add such in #474. So for now manual processing is required.

Mingun (changes requested):
This is good, only two small changes required and can be merged

examples/read_nodes.rs
```diff
@@ -0,0 +1,149 @@
...
+        reader: &mut Reader<&[u8]>,
+        element: BytesStart,
+    ) -> Result<Translation, quick_xml::Error> {
+        let mut tag = Cow::Borrowed("");
+        let mut lang = Cow::Borrowed("");
+
+        for attr_result in element.attributes() {
+            let a = attr_result?;
+            match a.key.as_ref() {
+                b"Language" => lang = a.unescape_value()?,
```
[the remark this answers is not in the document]
ultrasaurus: I can change, though FYI: the example runs fine for me locally:
```
$ cargo run --example read_nodes
   Compiling quick-xml v0.26.0 (/Users/sallen/src/rust/ref/quick-xml)
    Finished dev [unoptimized + debuginfo] target(s) in 1.27s
     Running `target/debug/examples/read_nodes`
[examples/read_nodes.rs:142] "{:?}" = "{:?}"
[examples/read_nodes.rs:142] &translations = [
    Translation {
        tag: "HELLO",
        lang: "ja",
        text: "こんにちは",
    },
    Translation {
        tag: "BYE",
        lang: "ja",
        text: "さようなら",
    },
    Translation {
        tag: "HELLO",
        lang: "es",
        text: "Hola",
    },
    Translation {
        tag: "BYE",
        lang: "es",
        text: "Adiós",
    },
]
```

ultrasaurus:
@Mingun fixed requested changes.  See comment above... odd that you mention `unescape_value` is unavailable, but it worked for me locally.  Made the change anyhow, though would be nice to understand your comment for my future coding :)

Mingun (approved):

ultrasaurus:
🎉

Mingun (approved):

PR #505, opened 2022-11-08
Example: read using serde

ultrasaurus:
This example fails, but I haven't figured out how to fix.  I would guess it is because of the ` <Localization>` node (which I don't know how to specify using serde.  

I didn't see any docs on this, though tests are quite comprehensive!   I personally find tests easier to understand than docs, though I haven't yet found a test matching this use case.  Is it supported?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
 tokio = { version = "1.21", default-features = false, features = ["macros", "rt"] }
 tokio-test = "0.4"
+anyhow = "1"
+
 
 [lib]
 bench = false
@@ -203,3 +205,7 @@ path = "tests/serde-migrated.rs"
 name = "async-tokio"
 required-features = ["async-tokio"]
 path = "tests/async-tokio.rs"
+
+[[example]]
+name = "read_nodes_serde"
+required-features = ["serialize"]
--- a/examples/read_nodes_serde.rs
+++ b/examples/read_nodes_serde.rs
@@ -1,0 +1,102 @@
+// note: to use serde, the feature needs to be enabled
+// run example with:
+//    cargo run --example read_nodes_serde --features="serialize"
+
+use quick_xml::de::DeError;
+use quick_xml::de::Deserializer;
+use serde::Deserialize;
+
+#[derive(Debug, PartialEq, Default, Deserialize)]
+#[serde(default)]
+struct Translation {
+    #[serde(rename = "@Tag")]
+    tag: String,
+    #[serde(rename = "@Language")]
+    lang: String,
+    #[serde(rename = "Text")]
+    text: String,
+}
+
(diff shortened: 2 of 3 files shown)


ultrasaurus:
note: this is on top of https://github.com/tafia/quick-xml/pull/504  (I just rebased, so history is a bit weird.  I can separate out if you want, just figured the other would likely be ready sooner.)

examples/read_nodes_serde.rs
```diff
@@ -0,0 +1,102 @@
...
+    // If type was deserialized, the whole XML document should be consumed
+    if let Ok(_) = result {
+        match <()>::deserialize(&mut de) {
+            Err(DeError::UnexpectedEof) => (),
+            e => panic!("Expected end `UnexpectedEof`, but got {:?}", e),
+        }
+    }
+
+    result
+}
```
Mingun, line 83: That is a special reimplementation of [`from_str`](https://docs.rs/quick-xml/latest/quick_xml/de/fn.from_str.html) for tests, to ensure that we consume all input, no need to use it in an example. Just use [`from_str`](https://docs.rs/quick-xml/latest/quick_xml/de/fn.from_str.html).

Cargo.toml
```diff
@@ -28,6 +28,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
 tokio = { version = "1.21", default-features = false, features = ["macros", "rt"] }
 tokio-test = "0.4"
+anyhow = "1"
```
Mingun, line 31: I think we can do without that dependency. It does not give any special advantages in our example

Mingun (changes requested):
Only two small changes and can be merged. The CI fail related to #504

examples/read_nodes_serde.rs
ultrasaurus: ah, thank you!

Cargo.toml
ultrasaurus: I'm not sure what error to return from main().  If I change to:
```
fn main() -> Result<(), quick_xml::Error> {
```

I get:
```
  --> examples/read_nodes_serde.rs:70:39
   |
64 | fn main() -> Result<(), quick_xml::Error> {
   |              ---------------------------- expected `quick_xml::Error` because of this
...
70 |     let config: Config = from_str(XML)?;
   |                                       ^ the trait `From<DeError>` is not implemented for `quick_xml::Error`
   |
```

How do you recommend writing this example to avoid this compile error?

ultrasaurus:
I think the CI failure is because of the assertions in the example, see: https://github.com/tafia/quick-xml/pull/507

Mingun:
> I didn't see any docs on this, though tests are quite comprehensive!

Ah, yes, I forgot to mention, that I'm working on docs and plan to include it to the next release. Draft (outdated) is [here](https://github.com/tafia/quick-xml/pull/369).

PR #507, opened 2022-11-09
Read serde (on master)

ultrasaurus:
this fails and I don't know how to fix:
```
Airtemis:quick-xml sallen$ cargo run --example read_nodes_serde --features="serialize"
   Compiling quick-xml v0.26.0 (/Users/sallen/src/rust/ref/quick-xml)
    Finished dev [unoptimized + debuginfo] target(s) in 1.26s
     Running `target/debug/examples/read_nodes_serde`
[examples/read_nodes_serde.rs:71] "{:?}" = "{:?}"
[examples/read_nodes_serde.rs:71] &config = Config {
    settings: DefaultSettings {
        language: "es",
        greeting: "HELLO",
    },
    translation: [
        Translation {
            tag: "",
            lang: "",
            text: "",
        },
    ],
}
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `4`', examples/read_nodes_serde.rs:76:5
```


ultrasaurus:
Note: I added Anyhow to I'm not sure what error to return from main().  If I change to:
```
fn main() -> Result<(), quick_xml::Error> {
```

I get:
```
  --> examples/read_nodes_serde.rs:70:39
   |
64 | fn main() -> Result<(), quick_xml::Error> {
   |              ---------------------------- expected `quick_xml::Error` because of this
...
70 |     let config: Config = from_str(XML)?;
   |                                       ^ the trait `From<DeError>` is not implemented for `quick_xml::Error`
   |
```

Mingun:
Return [`DeError`](https://docs.rs/quick-xml/latest/quick_xml/enum.DeError.html). It [includes](https://docs.rs/quick-xml/latest/quick_xml/enum.DeError.html#variant.InvalidXml) `Error` inside

ultrasaurus:
ok, I've removed `Anyhow` dependency, but I still don't know how to use serde to read nested structure.  In this example, I want to skip `Localization` node and read the `Translation` nodes.  I couldn't figure out the syntax to do that.

dralley:
@Mingun Is this a limitation or is there a trick to make it work?

Cargo.toml
```diff
@@ -28,6 +28,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde-value = "0.7"
 tokio = { version = "1.21", default-features = false, features = ["macros", "rt"] }
 tokio-test = "0.4"
+anyhow = "1"
+
```
Mingun, line 32: Remove this change

Cargo.toml
```diff
@@ -203,3 +205,7 @@ path = "tests/serde-migrated.rs"
 name = "async-tokio"
 required-features = ["async-tokio"]
 path = "tests/async-tokio.rs"
+
+[[example]]
+name = "read_nodes_serde"
+required-features = ["serialize"]
```
Mingun, line 211: Please add newline at the end

examples/read_nodes_serde.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Debug, PartialEq, Default, Deserialize)]
+#[serde(default)]
+struct Translation {
+    #[serde(rename = "@Tag")]
+    tag: String,
+    #[serde(rename = "@Language")]
+    lang: String,
+    #[serde(rename = "Text")]
+    text: String,
```
Mingun, line 16: If this is just an example format (and not some that has a name) I think it would be better to show how to read inner text, using `#[serde(rename = "$text")]`. Then 
```xml
    <Translation Tag="HELLO" Language="ja">
      <Text>こんにちは</Text>
    </Translation>
```
becomes
```xml
    <Translation Tag="HELLO" Language="ja">
      こんにちは
    </Translation>
```

examples/read_nodes_serde.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[derive(Debug, PartialEq, Default, Deserialize)]
+#[serde(default)]
+struct Config {
+    #[serde(rename = "DefaultSettings")]
+    settings: DefaultSettings,
+    #[serde(rename = "Localization")]
+    translation: Vec<Translation>,
+}
```
Mingun, line 35: [This should be](https://docs.rs/quick-xml/latest/quick_xml/de/index.html#element-lists):
```suggestion
#[derive(Debug, PartialEq, Default, Deserialize)]
#[serde(rename_all = "CamelCase")]
struct Config {
    #[serde(rename = "DefaultSettings")]
    settings: DefaultSettings,
    localization: Localization,
}
#[derive(Debug, PartialEq, Default, Deserialize)]
#[serde(rename_all = "CamelCase", default)]
struct Localization {
    translation: Vec<Translation>,
}
```
I think, it is also would better to use `#[serde(rename_all = "...")]` attribute whenever possible and keep individual `#[serde(rename = "...")]` only where necessary. This makes the code cleaner at my glance.

Mingun (changes requested):

examples/read_nodes_serde.rs
ultrasaurus: nice suggestion.  done!

ultrasaurus:
Updated example works now!  Thanks so much for helping me understand how to fix it.  I think I've addressed all the feedback.

Mingun (approved):

PR #509, opened 2022-11-13
Add additional tests for borrowed input

### LINKED ISSUE #328: Support zero-copy reading in serde deserializer

Mingun:
Borrow input in attributes is working now, I suppose, after rewrite in #490. Now that is ensured by tests. Closes #328

### DIFF
--- a/tests/serde-de.rs
+++ b/tests/serde-de.rs
@@ -30,19 +30,6 @@ where
     result
 }
 
-#[test]
-fn string_borrow() {
-    #[derive(Debug, Deserialize, PartialEq)]
-    struct BorrowedText<'a> {
-        #[serde(rename = "$text")]
-        text: &'a str,
-    }
-
-    let borrowed_item: BorrowedText = from_str("<text>Hello world</text>").unwrap();
-
-    assert_eq!(borrowed_item.text, "Hello world");
-}
-
 /// Tests for deserializing into specially named field `$text` which represent
 /// textual content of an XML element
 mod text {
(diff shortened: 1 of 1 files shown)


tests/serde-de.rs
```diff
@@ -6067,3 +6054,126 @@ fn from_str_should_ignore_encoding() {
...
+                    string: "without escape sequences",
+                }
+            );
+        }
+    }
+
+    /// Deserialization of all XML's in that module expected to fail because
+    /// values requires unescaping that will lead to allocating an internal
+    /// buffer by deserializer, but the `Borrowed*` types couldn't take ownership
+    /// on it.
```
dralley, line 6123: Is there any concept of opportunistic zero-copy with `Cow<&str>`?  As someone not terribly familiar with serde.

dralley (approved):

tests/serde-de.rs
Mingun: If I understand you correctly, `Cow<str>` will borrow whenever possible i. e. when deserializer can provide data of underlying source by calling [`Visitor::visit_borrowed_str`](https://docs.rs/serde/latest/serde/de/trait.Visitor.html#method.visit_borrowed_str). The opposite case when the deserializer allocates `String` by itself and calls [`Visitor::visit_string`](https://docs.rs/serde/latest/serde/de/trait.Visitor.html#method.visit_string). The intermediate case when deserializer provides borrow to the data, that it owns, via [`Visitor::visit_str`](https://docs.rs/serde/latest/serde/de/trait.Visitor.html#method.visit_str).

In our case we can either provide access to the XML source by calling [`Visitor::visit_borrowed_str`](https://docs.rs/serde/latest/serde/de/trait.Visitor.html#method.visit_borrowed_str), or allocate a new `String` (because we in any case need to allocate to do unescaping and/or decoding) and give it to the `Visitor`. If XML in UTF-8 and does not contain escaped data or mixed text/CDATA or CDATA/CDATA content (after solving #474), the data will be borrowed

PR #511, opened 2022-11-18
Fix #257 and allow $text to work with tags in the text

### LINKED ISSUE #257: Help deserialize mixed tags and string in body $value (html text formatting)

JOSEPHGILBY:
This is a very early attempt at solving #257 (awful, but functional, code below). Unfortunately, I ran into a design issue so I'd like to open the discussion now and get your feedback on what to do.

Suppose we have the following struct to deserialize (from the test case attached in serde-de.rs below)

```rust
struct Trivial<T> {
    #[serde(rename = "$text")]
    value: T,
}
```
The test case has the following xml
```xml
<root>style tags <em>in this document</em></root>
```
deserialize into
```rust
Trivial {
    value: "style tags <em>in this document</em>".to_string(),
}
```

The test case also assumes the xml
```xml
<outer><root>style tags <em>in this document</em></root></outer>
``` 
**should not** deserialize (missing field `$text`)

However, if we change the design for how $text works to include embedded tags, this would now be deserializable where 'outer' is the root tag and everything between <outer></outer> can now be fed into a text field because root can be part of the string now. 

Outside of the above being newly expected behavior we could make the user describe which tags can be deserialized into text fields instead of read:
```rust
struct Trivial<T> {
    #[serde(rename = "$text$<EM>$<B>")] // only <EM>...</EM>s and <B>...</B>s are allowed to be part of the string
    value: T,
}
```

Let me know if you have any other ideas on what would be best to do here.

### DIFF
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -187,6 +187,8 @@ where
     /// <tag>value for VALUE_KEY field<tag>
     /// ```
     has_value_field: bool,
+    has_text_field: bool,
+    has_read_text_to_end: bool,
 }
 
 impl<'de, 'a, R> MapAccess<'de, 'a, R>
@@ -206,6 +208,8 @@ where
             source: ValueSource::Unknown,
             fields,
             has_value_field: fields.contains(&VALUE_KEY),
+            has_text_field: fields.contains(&TEXT_KEY),
+            has_read_text_to_end: false,
         })
     }
 }
@@ -234,6 +238,13 @@ where
             let de = QNameDeserializer::from_attr(QName(&slice[key]), decoder)?;
             seed.deserialize(de).map(Some)
         } else {
+            if self.has_text_field {
+                if self.has_read_text_to_end {
+                    return Ok(None)
+                }
+                self.de.peek_to_end_as_text(self.start.name())?;
+                self.has_read_text_to_end = true;
+            }
             // try getting from events (<key>value</key>)
             match self.de.peek()? {
                 // We shouldn't have both `$value` and `$text` fields in the same
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -409,6 +409,9 @@ where
 
     #[cfg(not(feature = "overlapped-lists"))]
     peek: Option<DeEvent<'de>>,
+
+    #[cfg(not(feature = "overlapped-lists"))]
+    peek_2: Option<DeEvent<'de>>,
 }
 
 /// Deserialize an instance of type `T` from a string of XML text.
@@ -498,6 +501,9 @@ where
 
             #[cfg(not(feature = "overlapped-lists"))]
             peek: None,
+
+            #[cfg(not(feature = "overlapped-lists"))]
+            peek_2: None,
         }
     }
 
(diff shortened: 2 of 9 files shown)


Mingun:
> The test case also assumes the xml
> 
> ```
> <outer><root>style tags <em>in this document</em></root></outer>
> ```
> 
> **should not** deserialize (missing field `$text`)

The test should be changed. It is expected, that it would return
```rust
"<root>style tags <em>in this document</em></root>"
```

> Outside of the above being newly expected behavior we could make the user describe which tags can be deserialized into text fields instead of read:

I think, it is better just to have a flag is we want to have inner tags inside or not.

JOSEPHGILBY:
@Mingun What do you think would be a good format for indicating that inside tags can be accepted for the text field? Maybe split $text into $text_including_tags and $text_without_tags? Or maybe keep $text and create $text_with_tags. $text should be able to handle embedded tags if there is a <![CDATA[]]> wrapping it (and I think quick-xml already handles this case), but I currently have a document where they didn't wrap the text with CDATA. Instead, there's an accompanying DTD file containing the following lines
```DTD
<!ELEMENT TEST (#PCDATA | EM | B)*>
.
.
.
<!ELEMENT EM (#PCDATA | B)*>
<!ELEMENT B (#PCDATA | EM)*>
```

Since quick-xml doesn't currently parse DTD, but I think you did mention that you would like quick-xml to somewhat follow the spirit behind DTD, I was thinking that's why we might want the user to specify which specific tags could be expected when serializing a $text field.

Mingun:
Probably `$html` (when [`check_end_names`](https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Reader.html#method.check_end_names) should be disabled) and `$xml` (when not) will be the most obvious choice.

The feature should be carefully designed, because there is a lot of edge cases. For example:
```
<text><!-- include?-->content with <em>tags</em><text>
<text>content <!-- ...or this? -->with <em>tags</em><text>
```

dralley:
@JOSEPHGILBY Do you intend to continue this work?

JOSEPHGILBY:
Unfortunately, I don't have the time.

PR #519, opened 2022-12-02
Fix #517: Fix swapped codes for `\r` and `\n` characters when escape them

### LINKED ISSUE #517: Escaping for \r and \n have wrong value
This code here https://github.com/tafia/quick-xml/blob/master/src/escapei.rs#L120-L121 has the escape codes for \r and \n swapped

Mingun:
Some tests was changed for consistency.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -28,6 +28,7 @@
   ```
 - [#514]: Fix wrong reporting `Error::EndEventMismatch` after disabling and enabling
   `.check_end_names`
+- [#517]: Fix swapped codes for `\r` and `\n` characters when escaping them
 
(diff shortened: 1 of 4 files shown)


Changelog.md
```diff
@@ -28,6 +28,7 @@
   ```
 - [#514]: Fix wrong reporting `Error::EndEventMismatch` after disabling and enabling
   `.check_end_names`
+- [#517]: Fix swapped codes for `\r` and `\n` characters when escape them
```
dralley, line 31: s/escape/escaping

dralley (approved):

PR #520, opened 2022-12-04
Merge consequent text and CDATA events into one string 

### LINKED ISSUE #474: Merge text and CDATA events in serde deserializer

Mingun:
This PR fixes #474 and introduces a way to read current parser configuration, which was impossible before that.

I've changed the way how configuration is accessed and changed: instead of having functions to change configuration flags, readers now provides a reference to a `Config` object. Immutable and mutable references are provided. This new feature is used to temporary disable trimming while read text events in serde `Deserializer`.

After fixing #516, all configuration flags are safe to changed at any time, because their does not change the internal state of a reader in a user-visible way (for example, the `expand_empty_elements` changes an internal state of a reader, but that change is rolled back after next call to `read_event`, so user cannot see it consequences. It is safe to disable this setting just after read fake `Start` event and still get a fake `End` event after that).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -27,6 +27,11 @@
   sequence type (for example, `Vec` or tuple)
 - [#540]: Fix a compilation error (probably a rustc bug) in some circumstances.
   `Serializer::new` and `Serializer::with_root` now accepts only references to `Write`r.
+- [#520]: Merge consequent (delimited only by comments and processing instructions)
+  texts and CDATA when deserialize using serde deserializer. `DeEvent::Text` and
+  `DeEvent::CData` events was replaced by `DeEvent::Text` with merged content.
+  The same behavior for the `Reader` does not implemented (yet?) and should be
+  implemented manually
 
 ### Misc Changes
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,8 +16,8 @@
   enums from textual content
 - [#556]: `to_writer` and `to_string` now accept `?Sized` types
 - [#556]: Add new `to_writer_with_root` and `to_string_with_root` helper functions
-- [#520]: Add method `BytesText::inplace_trim_start` to trim leading spaces from
-  text events
+- [#520]: Add methods `BytesText::inplace_trim_start` and `BytesText::inplace_trim_end`
+  to trim leading and trailing spaces from text events
 
 ### Bug Fixes
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,8 @@
   enums from textual content
 - [#556]: `to_writer` and `to_string` now accept `?Sized` types
 - [#556]: Add new `to_writer_with_root` and `to_string_with_root` helper functions
+- [#520]: Add method `BytesText::inplace_trim_start` to trim leading spaces from
+  text events
 
 ### Bug Fixes
 
@@ -31,6 +33,7 @@
 [externally tagged]: https://serde.rs/enum-representations.html#externally-tagged
 [#490]: https://github.com/tafia/quick-xml/pull/490
 [#510]: https://github.com/tafia/quick-xml/issues/510
+[#520]: https://github.com/tafia/quick-xml/pull/520
 [#537]: https://github.com/tafia/quick-xml/issues/537
 [#540]: https://github.com/tafia/quick-xml/issues/540
 [#541]: https://github.com/tafia/quick-xml/pull/541
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -313,7 +313,7 @@ where
             ValueSource::Text => match self.de.next()? {
                 DeEvent::Text(e) => seed.deserialize(SimpleTypeDeserializer::from_text_content(
                     // Comment to prevent auto-formatting
-                    e.decode(true)?,
+                    e.unescape()?,
                 )),
                 DeEvent::CData(e) => seed.deserialize(SimpleTypeDeserializer::from_text_content(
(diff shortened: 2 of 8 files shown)


Mingun:
I've also checked, that XmlBeans is able to parse splitted numbers:
```xml
12<!--comment--><![CDATA[34]]>
```
is a valid representation for integer `1234`, which means that XmlBeans uses the same approach as was selected here: skip all comments and PIs, merge all consequent text and CDATA events and then parse to the required type.

Mingun:
I found, that, because of lookahead, we can cache trimmed event and return incorrect results (= trim spaces inside the string). More work needed

Mingun:
Ok, it's ready for review now. Of course, it would be great, if quick-xml will provide an API to read correctly trimmed text events from `Reader`, but this will require lookahead from one event (to trim start) to undefinitely number of events (to trim end, you need to skip all comments / processing instructions to determine if you need to trim end or not).

So I do not provide a low-level API, but provide a `BytesText::inplace_trim_start()` and `BytesText::inplace_trim_end()` instead. Then users could use that primitives to trim their events.

Changelog.md
```diff
@@ -14,16 +14,22 @@
...
 - [#541]: Deserialize specially named `$text` enum variant in [externally tagged]
   enums from textual content
+- [#520]: Add methods `BytesText::inplace_trim_start` and `BytesText::inplace_trim_end`
+  to trim leading and trailing spaces from text events
 
 ### Bug Fixes
 
 - [#537]: Restore ability to deserialize attributes that represents XML namespace
   mappings (`xmlns:xxx`) that was broken since [#490]
+- [#520]: Merge consequent (delimited only by comments and processing instructions)
```
dralley, line 24: If this only applies to the serde deserializer, that should probably be noted here.

dralley: Also, if this is something that isn't practical to handle in the low-level API for the time being, then it would be good to document the limitation.  And if it may be possible to do in the future (perhaps requiring API changes to do so) then we should clone #474.

I appreciate the detailed writeup of the issue, by the way.

Mingun: I clearly indicated that this applies only to the deserializer. The documentation thing probably solved by the last commit in this branch ([16fca58](https://github.com/tafia/quick-xml/pull/520/commits/16fca5813961c348f36eabfb380fa92d398d0bbb))?

dralley (approved):

Arthur-PI:
Hello, since my question is quite related and issues have been opened for this before, I put my message here, feels more relevant.  
Thanks to this PR, the trimming process is much clearer and with little modifications, it is possible to disable end trimming without breaking anything. But removing start trimming still breaks things.  
But the thing is, even if it is now partially possible to do so, there is no way of doing it through some sort of config struct or something like that when doing `quick_xml::de::from_str`, we are still blocked with default trim which is annoying.  
Is there a solution planed for this or this needs to be discussed/think @Mingun ?

To be sure, I am only talking here about `::Text` and `::CData` trimming and not the rest.  
Thank you for your time

Mingun:
It is possible to add a setting to enable / disable trimming. I'll accept such a PR.

To be clear: `CData` is never trimmed because it's purpose to represent data exactly as it written in XML.

PR #521, opened 2022-12-06
Implement Clone for error enums

zeenix:
This would allow using crates to embed errors from this crate into their error types that implement Clone.

Unfortunately `std::io::Error` [does not implement `Clone`](https://github.com/rust-lang/rust/issues/24135) so we've to loose some of its state during cloning. The most importat bit (error kind) is preserved though.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,10 @@
 
 ### New Features
 
+- [#521]: Implement `Clone` for all error types. This required changing `Error::Io` to contain
+  `Arc<std::io::Error>` instead of `std::io::Error` since `std::io::Error` does not implement
+  `Clone`.
+
 ### Bug Fixes
 
 - [#490]: Ensure that serialization of map keys always produces valid XML names.
--- a/Changelog.md
+++ b/Changelog.md
@@ -51,6 +51,7 @@
     that should be represented either by an `enum`, or by sequence of `enum`s
     (`Vec`, tuple, etc.), or by string. Use it when you want to map field to any
     content of the field, text or markup
+- [#521]: MSRV bumped to 1.52.
 
   Refer to [documentation] for details.
 
@@ -58,6 +59,7 @@
 [#500]: https://github.com/tafia/quick-xml/issues/500
 [#514]: https://github.com/tafia/quick-xml/issues/514
 [#517]: https://github.com/tafia/quick-xml/issues/517
+[#521]: https://github.com/tafia/quick-xml/pull/521
 [XML name]: https://www.w3.org/TR/xml11/#NT-Name
 [documentation]: https://docs.rs/quick-xml/0.27.0/quick_xml/de/index.html#difference-between-text-and-value-special-names
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@1.46.0
+    - uses: dtolnay/rust-toolchain@1.52.0
     - run: cargo check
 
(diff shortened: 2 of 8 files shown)


Mingun:
Cloneable errors seems not in a trend in Rust, it is unknown what the long-term consequences of this. I'm not totally against this PR, but I would like to hear what the situations where you need to clone errors?

zeenix:
> Cloneable errors seems not in a trend in Rust, it is unknown what the long-term consequences of this. I'm not totally against this PR, but I would like to hear what the situations where you need to clone errors?

Thanks for the quick response and that's very fair. My own use case: I've a task that owns a socket and reads and parses messages from the socket and then forwards them to all interested parties through a broadcast channel. Since `Result` only implements `Clone` when `E: Clone`, I can not send a `Result` the channel. Currently I work around this by using a broadcast channel only for success and a separate channel for error but that means that I only communicate the error to one of the streams.

Please note that in my crate, I'm fixing [this a bit differently](https://gitlab.freedesktop.org/zeenix/zbus/-/commit/7829e922798aaa5759ba0813a629e35cf6c8a58c#5f2235c129d7eea459e416cc0555ac5bd76e035d_22_22), which won't work for you unless you're ok with breaking API (my error type is `non_exhaustive`) so I went with the approach in the PR.

Mingun (changes requested):
I'm fine with replacing `Error::Io(IoError)` to `Error::Io(Arc<IoError>)`, which, I think, would be more elegant solution. We also can hide that under a feature flag (say, `cloneable-errors`), but I think that this is not required (besides, that is anti-pattern, because that create a incompatible API depending on the feature flags, which [should be avoided](https://github.com/tafia/quick-xml/issues/180)). @dralley, what to you think?

Whatever path we choose, it is important to describe it in the documentation for `Error`, and add a changelog entry.

I think, that will resolve #360?

zeenix:
> I'm fine with replacing `Error::Io(IoError)` to `Error::Io(Arc<IoError>)`, which, I think, would be more elegant solution.

I completely agree. I will revise the commits after hearing opinion of @dralley.

> Whatever path we choose, it is important to describe it in the documentation for `Error`, 
> and add a changelog entry.

For sure. I'll do that as part of the update.

zeenix:
> I think, that will resolve #360?

Not sure. single consumer APIs should not require cloning.

dralley:
I'm fine with `Arc<IoError>`.  Agreed about not wanting divergent signatures.

zeenix:
@Mingun Updated the commits. PTALA.

Changelog.md
```diff
@@ -12,6 +12,10 @@
 
 ### New Features
 
+- Implement `Clone` for all error types. This required changing `Error::Io` to contain
```
Mingun, line 15:
```suggestion
- [#521]: Implement `Clone` for all error types. This required changing `Error::Io` to contain
```
and add the link into list of links below (keep it sorted by number)

Mingun (approved):
`impl<E: Error> Error for Arc<E>` was implemented in [Rust 1.52](https://github.com/rust-lang/rust/blob/master/RELEASES.md#libraries-13), so need to bump our MSRV. Please add a note about that to the Changelog, before "New Features" section

Changelog.md
```diff
@@ -10,8 +10,14 @@
 
 ## Unreleased
 
+- MSRV bumped to 1.52.
```
dralley, line 13: This should go under a `## Misc Changes` section

Mingun: Yeah, that section is better than in the beginning. Sorry for noise, didn't think about it before

zeenix:
Lol, apparently GH wants me to only request review from one person. :) Sorry about adding and removing dance.

Mingun:
Thanks!

PR #523, opened 2022-12-11
Fix couple of bugs in serde deserializer

Mingun:
This PR fixes two bugs in serde deserializer that I found during work on #521:
- text of CDATA content before the element was silently skipped when attempting to deserialize `struct`s. This should not happen
- unescaping of escaped entities is performed before applying rules to split elements in `xs:list`. That means, that you cannot encode spaces in `xs:list` elements using entity references `&#32;` and `&#x20;` -- that escape sequences should be threated as list delimiters. The Java's mature XmlBeans library confirm that.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -38,6 +38,8 @@
   ```xml
   unwanted text<struct>...</struct>
   ```
+- [#523]: Fix incorrect handling of `xs:list`s with encoded spaces: they still
+  act as delimiters, which is confirmed also by mature XmlBeans Java library
 
 ### Misc Changes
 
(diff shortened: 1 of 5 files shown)


dralley:
IMO we should probably cut a release after this

Mingun:
I hope I finish #521 this year and cut a release after that

Changelog.md
```diff
@@ -38,6 +38,8 @@
   ```xml
   unwanted text<struct>...</struct>
   ```
+- [#523]: Fix incorrect handling of `xs:list`s with encoded spaces: their are still
```
dralley, line 41: Did you mean "they still act as delimiters"?

dralley (approved):

Changelog.md
Mingun: Yes

PR #524, opened 2022-12-12
Fix #473: add explicit serde feature that is independent from dependency

### LINKED ISSUE #473: quick-xml requires implicitly enabled `serde/derive`
It's impossible to depend on `quick-xml/serialize` without depending on `serde/derive`

Details: #472

Mingun:
This should fix #473. @loyd, @xkr47, can you check?

Checked by 
```console
cargo init
cargo add quick-xml --features serialize --path path/to/local/copy/of/quick-xml
cargo check
```
(failed without `--path` with the fix)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,8 +164,21 @@ escape-html = []
 ## [`Deserializer`]: crate::de::Deserializer
 overlapped-lists = []
 
-## Enables support for [`serde`] serialization and deserialization
-serialize = ["serde"]
+## Enables serialization of some types using [`serde`]. Probably your rarely will
+## need this feature enabled.
+##
+## This feature does NOT provide XML serializer or deserializer. You should use
+## the `serialize` feature for that instead.
+# Cannot name "serde" to avoid clash with dependency.
+# "dep:" prefix only avalible from Rust 1.60
+serde-types = ["serde/derive"]
+
+## Enables support for [`serde`] serialization and deserialization. When this
+## feature is enabled, quick-xml provides serializer and deserializer for XML.
+##
+## This feature does NOT enables serializaton of the types inside quick-xml.
+## If you need that, use the `serde-types` feature.
+serialize = ["serde"] # "dep:" prefix only avalible from Rust 1.60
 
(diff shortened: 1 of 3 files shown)


xkr47 (approved):
This works for me, including the comment I posted here: https://github.com/tafia/quick-xml/issues/473#issuecomment-1333986330

Cargo.toml
```diff
@@ -164,8 +164,21 @@ escape-html = []
...
 
-## Enables support for [`serde`] serialization and deserialization
-serialize = ["serde"]
+## Enables serialization of some types using [`serde`]. Probably your rarely will
+## need this feature enabled.
+##
+## This feature does NOT provide XML serializer or deserializer. You should use
+## the `serialize` feature for that instead.
+# Cannot name "serde" to avoid clash with dependency.
+# "dep:" prefix only avalible from Rust 1.60
```
dralley, line 173: Does this mean that with edition 2021 / MSRV 1.60, it is possible to simplify all of this?

Mingun: Yes, in Rust 1.60 it is possible to write
```toml
[dependencies]
serde = { version = "1.0", optional = true }

[features]
# depend on the dependency "serde" instead of feature "serde"
# (which will enable "derive" feature of "serde" which is not required
# for the "serialize" feature)
serialize = ["dep:serde"] 

# depend on "serde" dependency with enabled feature "derive"
serde = ["serde/derive"]
```

dralley: If we don't want to bump the MSRV right now then we should file an issue to simplify this in the future.  I wouldn't be against bumping MSRV though.

dralley (approved):

Cargo.toml
dralley: @Mingun ^

Mingun: You can create an issue if you wish. Anyway, I marked this place with Rust version that is required for more modern solution, so it can be grepped.

I think, this that having more modern solution is not critical, so it is better to avoid bumping MSRV just for it. Otherwise, probably some of our dependents will not happy with that.

Issue #497, opened 2022-10-28
Option<bool> doesn't (de)serialize properly with serde


dralley:
Using the following code:

```Rust
use serde;
use quick_xml;

#[derive(serde::Serialize, serde::Deserialize)]
pub struct Player {
    spawn_forced: Option<bool>
}

fn main() {
    let data = Player {
       spawn_forced: None,
    };

    let mut deserialize_buffer = String::new();
    quick_xml::se::to_writer(&mut deserialize_buffer, &data).unwrap();
    println!("{}", &deserialize_buffer);

    quick_xml::de::from_reader::<&[u8], Player>(&deserialize_buffer.as_bytes())
                    .unwrap();
}
```

The code crashes during the deserialize step

```
[dalley@thinkpad playground]$ cargo run
   Compiling playground v0.1.0 (/home/dalley/Devel/practice/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.28s
     Running `target/debug/playground`
<Player><spawn_forced/></Player>
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidBoolean("")', src/main.rs:98:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If instead `spawn_forced` is set to `Some(true)` or `Some(false)`, it works fine.  Only `None`, which serializes to `<spawn_forced/>`, panics upon deserialization.

```
[dalley@thinkpad playground]$ cargo run
   Compiling playground v0.1.0 (/home/dalley/Devel/practice/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.44s
     Running `target/debug/playground`
<Player><spawn_forced>true</spawn_forced></Player>
[dalley@thinkpad playground]$ cargo run
   Compiling playground v0.1.0 (/home/dalley/Devel/practice/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.28s
     Running `target/debug/playground`
<Player><spawn_forced>false</spawn_forced></Player>
```

dralley:
Test case minimized from https://github.com/djkoloski/rust_serialization_benchmark/pull/34

dralley:
This is using the current master branch ^^

Mingun:
This bug is not specific to `bool` -- any `Option<T>` fields that deserializing from empty element, are affecting. Let's look at such XMLs:
```xml
<root/>                               <!-- (1) None -->
[...]

PR #528, opened 2022-12-21
Add documentation for mapping from XML to Rust used by serde

Mingun:
Reincarnation of #369.

This PR adds a table to the documentation of [`de`](https://docs.rs/quick-xml/latest/quick_xml/de/index.html) module, that shows how XML pieces could be represented by Rust types. Not all of described concepts are work yet, in that case they are marked with `FIXME` marker.

Almost each example is proved by doctest, which should guarantee that they does not break in the future.

I would like to include this in the next release which I plan to make this year.

The best way to review is to generate documentation locally (this is PowerShell command):
```ps
$env:RUSTDOCFLAGS="--cfg docs_rs"; cargo +nightly doc --all-features
```
and go to the documentation of the `de` module.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -25,6 +25,8 @@
 //!     - [Primitives and sequences of primitives](#primitives-and-sequences-of-primitives)
 //!     - [Structs and sequences of structs](#structs-and-sequences-of-structs)
 //!     - [Enums and sequences of enums](#enums-and-sequences-of-enums)
+//! - [Frequently Used Patterns](#frequently-used-patterns)
+//!   - [`<element>` lists](#element-lists)
 //!
 //!
 //!
@@ -895,6 +897,8 @@
 //! Currently not working. The bug is tracked in [#510].
 //! </div>
 //!
+//! See also [Frequently Used Patterns](#element-lists).
+//!
 //! [field]: https://serde.rs/field-attrs.html#default
 //! [struct]: https://serde.rs/container-attrs.html#default
 //! </td>
@@ -1579,7 +1583,109 @@
 //! that is not enforced, so you can theoretically have both, but you should
 //! avoid that.
 //!
+//!
+//!
+//! Frequently Used Patterns
+//! ========================
+//!
+//! Some XML constructs used so frequent, that it is worth to document the recommended
+//! way to represent them in the Rust. The sections below describes them.
+//!
+//! ## `<element>` lists
+//! Many XML formats wrap lists of elements in the additional container,
+//! although this is not required by the XML rules:
+//!
+//! ```xml
+//! <root>
+//!   <field1/>
+//!   <field2/>
+//!   <list><!-- Container -->
+//!     <element/>
+//!     <element/>
+//!     <element/>
+//!   </list>
+//!   <field3/>
+//! </root>
+//! ```
+//! In this case, there is a great desire to describe this XML in this way:
+//! ```
+//! /// Represents <element/>
+//! type Element = ();
+//!
+//! /// Represents <root>...</root>
+//! struct AnyName {
+//!     // Incorrect
+//!     list: Vec<Element>,
+//! }
+//! ```
+//! This will not work, because potentially `<list>` element can have attributes
+//! and other elements inside. You should define the struct for the `<list>`
+//! explicitly, as you do that in the XSD for that XML:
+//! ```
+//! /// Represents <element/>
+//! type Element = ();
+//!
+//! /// Represents <root>...</root>
+//! struct AnyName {
+//!     // Correct
+//!     list: List,
+//! }
+//! /// Represents <list>...</list>
+//! struct List {
+//!     element: Vec<Element>,
+//! }
+//! ```
+//!
+//! If you want to simplify your API, you could write a simple function for unwrapping
+//! inner list and apply it via [`deserialize_with`]:
+//!
+//! ```
+//! # use pretty_assertions::assert_eq;
+//! use quick_xml::de::from_str;
+//! use serde::{Deserialize, Deserializer};
+//!
+//! /// Represents <element/>
+//! type Element = ();
+//!
+//! /// Represents <root>...</root>
+//! #[derive(Deserialize, Debug, PartialEq)]
+//! struct AnyName {
+//!     #[serde(deserialize_with = "unwrap_list")]
+//!     list: Vec<Element>,
+//! }
+//!
+//! fn unwrap_list<'de, D>(deserializer: D) -> Result<Vec<Element>, D::Error>
+//! where
+//!     D: Deserializer<'de>,
+//! {
+//!     /// Represents <list>...</list>
+//!     #[derive(Deserialize)]
+//!     struct List {
+//!         // default allows empty list
+//!         #[serde(default)]
+//!         element: Vec<Element>,
+//!     }
+//!     Ok(List::deserialize(deserializer)?.element)
+//! }
+//!
+//! assert_eq!(
+//!     AnyName { list: vec![(), (), ()] },
+//!     from_str("
+//!         <root>
+//!           <list>
+//!             <element/>
+//!             <element/>
+//!             <element/>
+//!           </list>
+//!         </root>
+//!     ").unwrap(),
+//! );
+//! ```
+//!
+//! Instead of writing such functions manually, you also could try <https://lib.rs/crates/serde-query>.
+//!
 //! [specification]: https://www.w3.org/TR/xmlschema11-1/#Simple_Type_Definition
+//! [`deserialize_with`]: https://serde.rs/field-attrs.html#deserialize_with
 //! [#474]: https://github.com/tafia/quick-xml/issues/474
 //! [#497]: https://github.com/tafia/quick-xml/issues/497
 //! [#510]: https://github.com/tafia/quick-xml/issues/510
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -1,6 +1,1356 @@
-//! Serde `Deserializer` module
+//! Serde `Deserializer` module.
 //!
-//! # Difference between `$text` and `$value` special names
+//! Due to the complexity of the XML standard and the fact that serde was developed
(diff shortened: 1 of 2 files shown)


src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
-//! # Difference between `$text` and `$value` special names
+//! Due to complexity of the XML standard and that fact that serde was developed
+//! with JSON in mind, not all serde concepts smoothly laid down on XML. This
+//! leads to that fact that some XML concepts are inexpressible in terms of serde
+//! derives and you may require implement deserialization manually.
+//!
+//! The most notable restriction is distinguish between _elements_ and _attributes_,
+//! no other serde format haven't such a conception.
+//!
+//! Due to that the mapping is performed in the best effort manner.
```
dralley, line 11:
```
//! Due to the complexity of the XML standard and the fact that serde was developed
//! with JSON in mind, not all serde concepts apply smoothly to XML. This
//! leads to that fact that some XML concepts are inexpressible in terms of serde
//! derives and may require manual deserialization.
//!
//! The most notable restriction is the ablity to distinguish between _elements_ and _attributes_,
//! as no other format used by serde has such a conception.
//!
//! Due to that the mapping is performed in a best effort manner.
```

src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
+//!     - [Enums and sequences of enums](#enums-and-sequences-of-enums)
+//! - [Frequently Used Patterns](#frequently-used-patterns)
+//!   - [`<element>` lists](#element-lists)
+//!
+//!
+//!
+//! Mapping XML to Rust types
+//! =========================
+//!
+//! Type names are never considered when deserializing, so you could name your
```
dralley, line 36:
```suggestion
//! Type names are never considered when deserializing, so you can name your
```

src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
+//!   - attribute and element values;
+//!   - text or CDATA content (including mixed text + CDATA content).
+//!
+//! <div style="background:rgba(120,145,255,0.45);padding:0.75em;">
+//!
+//! NOTE: examples, marked with `FIXME:` does not work yet -- any PRs that fixes
+//! that are welcome! The message after marker is a test failure message.
+//! Also, all that tests are marked with an `ignore` option, although their
+//! compiles. This is by intention, because rustdoc marks such blocks with
+//! an exclamation mark unlike `no_run` blocks.
```
dralley, line 54:
```
//! NOTE: examples marked with `FIXME:` do not work properly yet -- any PRs which fixes
//! them are welcome! The test failure message is noted underneath each of them.
//! Also, some tests are marked with an `ignore` option although they compile fine.
//! This is by intention, because rustdoc marks such blocks with an exclamation
//! mark unlike `no_run` blocks.
```

Could you elaborate on the last sentence also, it is not clear what you mean.

src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
+//!   # assert_eq!(X { x: Language::Rust  }, quick_xml::de::from_str("<x>Rust</x>").unwrap());
+//!   # assert_eq!(X { x: Language::Cpp   }, quick_xml::de::from_str("<x>C<![CDATA[p]]>p</x>").unwrap());
+//!   # assert_eq!(X { x: Language::Other }, quick_xml::de::from_str("<x><![CDATA[other]]></x>").unwrap());
+//!   ```
+//!
+//! <div style="background:rgba(120,145,255,0.45);padding:0.75em;">
+//!
+//! NOTE: deserialization to non-owned types (i.e. borrow from the input),
+//! such as `&str`, is possible only if you parse document in the UTF-8
+//! encoding and content do not contains XML escape sequences, such as `&lt;`
```
dralley, line 111:
```suggestion
//! encoding and content does not contain XML escape sequences such as `&lt;`,
```

`&lt;` is also an entity reference, it is just predefined by the specification. We should also mention character references. 

 You could say "does not contain entity references such as `&amp;`, or character references such as `&#xD;`.

src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
+//! <...><![CDATA[content]]></...>
+//! ```
+//! ```xml
+//! <...>text<![CDATA[cdata]]>text</...>
+//! ```
+//! </td>
+//! <td>
+//!
+//! You can use any type that can be deserialized from an `&str`, for example:
+//! - [`String`] and [`&str`]
```
dralley, line 84: If `Cow` works lets mention that also.

src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
+//! See the next row to learn where in your struct definition you should
+//! use that type.
+//!
+//! According to the XML Schema specification, delimiters for elements is one
+//! or more space (`' '`, `'\r'`, `'\n'`, and `'\t'`) character(s).
+//!
+//! <div style="background:rgba(120,145,255,0.45);padding:0.75em;">
+//!
+//! NOTE: according to the XML Schema restrictions, you cannot escape those
+//! white-space characters, so list elements are _always_ will not contain them.
```
dralley, line 169:
```
//! white-space characters, so list elements will _never_ contain them.
```

src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
+//! ```xml
+//! <any-tag one="..." two="..."/>
+//! ```
+//! </td>
+//! <td>
+//!
+//! A structure where an each XML attribute are mapped to a field with a name
+//! started from `@`. Because an `@` character is not allowed in the Rust identifier,
+//! you should use `#[serde(rename = "@...")]` attribute to rename it.
+//! Name of the struct itself does not matter:
```
dralley, line 202:
```
//! A structure where each XML attribute is mapped to a field with a name
//! starting with `@`. Because Rust identifiers do not permit the `@` character,
//! you should use the `#[serde(rename = "@...")]` attribute to rename it.
//! The name of the struct itself does not matter:
```

src/de/mod.rs
```diff
@@ -1,6 +1,1324 @@
...
+//!   <optional>...</optional>
+//! </any-tag>
+//! ```
+//! ```xml
+//! <any-tag/>
+//! ```
+//! </td>
+//! <td>
+//!
+//! A structure with an optional field:
```
dralley, line 428: This might be a good place to point out some of the known peculiarities, e.g. https://github.com/tafia/quick-xml/issues/497

src/de/mod.rs
Mingun: See the difference:
difference-between-no_run-and-ignore

src/de/mod.rs
Mingun: Done

src/de/mod.rs
Mingun: Done

src/de/mod.rs
Mingun: Replaced "exclamation mark" with "information icon"

src/de/mod.rs
Mingun: Rephrased as your last suggestion.

src/de/mod.rs
Mingun: Added a bullet with `Cow<str>`

src/de/mod.rs
Mingun: Done

src/de/mod.rs
Mingun: Done

src/de/mod.rs
Mingun: Added a known bug mark

src/de/mod.rs
```diff
@@ -1,6 +1,1352 @@
...
+//!     <two>...</two>
+//!   </choice>
+//! </any-tag>
+//! ```
+//! </td>
+//! <td>
+//!
+//! A structure with a field of an intermediate type with one field of `enum` type.
+//! Actually, this example is not necessary, because you can construct it by yourself
+//! using composition rules that was described above. However the XML construction,
```
dralley, line 698:
```suggestion
//! using the composition rules that were described above. However the XML construction
```

src/de/mod.rs
```diff
@@ -1,6 +1,1352 @@
...
+//! described here, are very common, so it is shown explicitly.
```
dralley, line 699:
```suggestion
//! described here is very common, so it is shown explicitly.
```

src/de/mod.rs
```diff
@@ -1,6 +1,1352 @@
...
+//!     <two>...</two>
+//!   </choice>
+//!   <field>...</field>
+//! </any-tag>
+//! ```
+//! </td>
+//! <td>
+//!
+//! A structure with a field of an intermediate type with one field of `enum` type.
+//! Actually, this example is not necessary, because you can construct it by yourself
```
dralley, line 765: Same changes as above

src/de/mod.rs
```diff
@@ -1,6 +1,1352 @@
...
+//!   <item/>
+//!   <item/>
+//! </any-tag>
+//! ```
+//! </td>
+//! <td>
+//!
+//! A structure with a field which have a sequence type, for example, [`Vec`].
+//! Because XML syntax does not distinguish between empty sequences and missed
+//! elements, we should indicate that on a Rust side, because serde will require
```
dralley, line 838:
```suggestion
//! elements, we should indicate that on the Rust side, because serde will require
```

src/de/mod.rs
```diff
@@ -1,6 +1,1352 @@
...
+//! </tbody>
+//! </table>
+//!
+//!
+//!
+//! Composition Rules
+//! =================
+//!
+//! XML format is very different from other formats supported by `serde`.
+//! One of such differences it is how data in the serialized form related to
```
dralley, line 1309:
```suggestion
//! One such difference is how data in the serialized form is related to
```

dralley (approved):
Free to merge after the requested edits

This documentation is very helpful, thanks for adding it!

Mingun:
Thanks for review! In the final push I've also removed unused links that was remained from previous revisions, and copy-paste a one note about multi-root XMLs.

PR #530, opened 2022-12-25
Fix bugs when parsing sequences in `overlapped-lists` mode

Mingun:
This fixes two bugs, that I've found:
- `Deserializer::skip()` could enter to an infinity loop when parse malformed XML (some tags does not closed)
- `Deserializer::read_to_end()` could return success when parse malformed XML (some tags does not closed)

I think both bugs are important enough to release 0.27.1

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,8 @@
 
 - [#530]: Fix an infinite loop leading to unbounded memory consumption that occurs when
   skipping events on malformed XML with the `overlapped-lists` feature active.
+- [#530]: Fix an error in the `Deserializer::read_to_end` when `overlapped-lists`
+  feature is active and malformed XML is parsed
 
 ### Misc Changes
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,7 +14,12 @@
 
 ### Bug Fixes
 
+- [#530]: Fix an infinite loop leading to unbounded memory consumption that occurs when
+  skipping events on malformed XML with the `overlapped-lists` feature active.
+
 ### Misc Changes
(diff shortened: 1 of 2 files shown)


src/de/mod.rs
```diff
@@ -2247,9 +2247,19 @@ where
 
                 // If we do not have skipped events, use effective reading that will
                 // not allocate memory for events
-                None => return self.reader.read_to_end(name),
+                None => {
+                    loop {
+                        self.reader.read_to_end(name)?;
```
dralley, line 2252: Could you add a comment about what the extra logic is for - with context it's not difficult to figure out, but it would help any uninitiated readers of this code.

Changelog.md
```diff
@@ -14,8 +14,14 @@
 
 ### Bug Fixes
 
+- [#530]: Fix an infinity loop that leads to an infinity memory consumption during
```
dralley, line 17:
```
- [#530]: Fix an infinite loop leading to unbounded memory consumption that occurs when
  skipping events on malformed XML with the `overlapped-lists` feature active.
```

dralley (approved):

src/de/mod.rs
Mingun: Added. Could you check that it is clearly enough? [Diff](https://github.com/tafia/quick-xml/compare/9e7dcaad8a19e89a648e8f35795fd49a388e4c15..ca7aa21ba111b28ca54ddc161c8302bba035904f)

dralley: :+1:

Issue #531, opened 2022-12-26
XML parse broken after upgrading to 0.27


Xuanwo:
In https://github.com/datafuselabs/opendal/pull/1101, some XML related test failed after upgrading to 0.27.

## Cases

```toml
quick-xml = { version = "0.27", features = ["serialize", "overlapped-lists"] }
```

```rust
use quick_xml::de;
use serde::Deserialize;

#[derive(Default, Debug, Deserialize)]
#[serde(default, rename_all = "PascalCase")]
struct Output {
    name: String,
    prefix: String,
    contents: Vec<Content>,
    common_prefixes: Option<Vec<CommonPrefix>>,
    marker: String,
    next_marker: Option<String>,
}

#[derive(Default, Debug, Deserialize)]
#[serde(default, rename_all = "PascalCase")]
struct CommonPrefix {
    prefix: String,
}

#[derive(Default, Debug, Deserialize)]
#[serde(default, rename_all = "PascalCase")]
struct Content {
    key: String,
    size: u64,
}

#[cfg(test)]
mod tests {
    use bytes::Buf;

    use super::*;

    #[test]
    fn test_parse_xml() {
        let bs = bytes::Bytes::from(
            r#"<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<ListBucketResult xmlns="http://obs.cn-north-4.myhuaweicloud.com/doc/2015-06-30/">
    <Name>examplebucket</Name>
    <Prefix>obj</Prefix>
    <Marker>obj002</Marker>
    <NextMarker>obj004</NextMarker>
    <MaxKeys>1000</MaxKeys>
    <IsTruncated>false</IsTruncated>
    <Contents>
        <Key>obj002</Key>
        <LastModified>2015-07-01T02:11:19.775Z</LastModified>
        <ETag>"a72e382246ac83e86bd203389849e71d"</ETag>
        <Size>9</Size>
        <Owner>
            <ID>b4bf1b36d9ca43d984fbcb9491b6fce9</ID>
        </Owner>
        <StorageClass>STANDARD</StorageClass>
    </Contents>
    <Contents>
        <Key>obj003</Key>
        <LastModified>2015-07-01T02:11:19.775Z</LastModified>
        <ETag>"a72e382246ac83e86bd203389849e71d"</ETag>
        <Size>10</Size>
        <Owner>
            <ID>b4bf1b36d9ca43d984fbcb9491b6fce9</ID>
        </Owner>
[...]

PR #532, opened 2022-12-26
docs: Add #531 into changelog for breaking changes

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -88,7 +88,7 @@
   by
 
   ```rust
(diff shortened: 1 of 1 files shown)


Mingun:
Heh. I actually thought, that you are willing to _fix_ this error, and not just document it :smile: . I don't think, that we should include descriptions of known bugs in release notes. Anyway, everybody can look at our issue tracker to find them (or report, if they does not exist yet). So I'll close. Sorry for misunderstanding

PR #535, opened 2023-01-01
Fix documentation example for specification of @attribute names

emarsden:
Also include a pointer in the README to the useful documentation on syntax for deserialization.


Mingun:
Thanks!

Issue #464, opened 2022-08-25
Recognize and process some special XML attributes


Mingun:
All names starts with `xml` (case-insensitive) are reserved by the XML standard, and some of them has special meaning. quick-xml could process some of them:
- `xml:lang` -- https://www.w3.org/TR/xml11/#sec-lang-tag (meta information about natural language of texts, stacked like namespace definitions)
- `xml:space` -- https://www.w3.org/TR/xml11/#sec-white-space (related: #285)

funkyfuture:
it's only the concrete prefixes `xml` and `xmlns` that are globally defined: https://www.w3.org/TR/xml-names/#xmlReserved

wt:
The NamespaceResolver doesn't seem to have any entries by default. This seems incorrect by my reading of  https://www.w3.org/TR/xml-names11/#xmlReserved. Shouldn't the `xml` namespace be definitionally mapped to `http://www.w3.org/XML/1998/namespace`?

wt:
It also seems like the xml namespace should not be overridable.

wt:
FWIW, xmlns also has a similar definitional mapping. However, given how xmlns is handled now in quick-xml, that probably doesn't need to be handled the same way. However, I am wondering if handling them the same way would make it more consistent to deal with reserved namespaces like this.

Mingun:
> Shouldn't the `xml` namespace be definitionally mapped to `http://www.w3.org/XML/1998/namespace`?

Yes, it should.

> It also looks like the `xml` namespace should not be overrideable.

It seems that override it is technically possible, but such XML document is incorrect (but seems still [well-formed](https://www.w3.org/TR/xml11/#dt-wellformed) and [valid](https://www.w3.org/TR/xml11/#dt-valid), but not [namespace-well-formed](https://www.w3.org/TR/xml-names11/#dt-nwf) because overriding violates [namespace constrains](https://www.w3.org/TR/xml-names11/#xmlReserved)).

wt:
Well-formed docs have to have a root element conforming to the description at https://www.w3.org/TR/xml11/#NT-element. This section says the following:

> This specification does not constrain the application semantics, use, or (beyond syntax) names of the element types and attributes, except that names beginning with a match to (('X'|'x')('M'|'m')('L'|'l')) are reserved for standardization in this or future versions of this specification.

[...]

PR #545, opened 2023-01-25
Add reserved namespace bindings.

wt:
This adds xml and xmlns namespace bindings. These are defined at https://www.w3.org/TR/xml-names11/#xmlReserved.

This is a a start of the work needed for #464

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -47,6 +47,22 @@ pub enum Error {
     EscapeError(EscapeError),
     /// Specified namespace prefix is unknown, cannot resolve namespace for it
     UnknownPrefix(Vec<u8>),
+    /// Error for when a reserved namespace is set incorrectly.
+    ///
+    /// This error returned in following cases:
+    /// - the XML document attempts to bind `xml` prefix to something other than
+    ///   `http://www.w3.org/XML/1998/namespace`
+    /// - the XML document attempts to bind `xmlns` prefix
+    /// - the XML document attempts to bind some prefix (except `xml`) to
+    ///   `http://www.w3.org/XML/1998/namespace`
+    /// - the XML document attempts to bind some prefix to
+    ///   `http://www.w3.org/2000/xmlns/`
+    InvalidPrefixBind {
+        /// The prefix that is tried to be bound
+        prefix: Vec<u8>,
+        /// Namespace to which prefix tried to be bound
+        namespace: Vec<u8>,
+    },
 }
 
 impl From<IoError> for Error {
@@ -119,6 +135,13 @@ impl fmt::Display for Error {
             Error::UnknownPrefix(prefix) => {
                 f.write_str("Unknown namespace prefix '")?;
                 write_byte_string(f, prefix)?;
+                f.write_str("'")
+            }
+            Error::InvalidPrefixBind { prefix, namespace } => {
+                f.write_str("The namespace prefix '")?;
+                write_byte_string(f, prefix)?;
+                f.write_str("' cannot be bound to '")?;
+                write_byte_string(f, namespace)?;
                 f.write_str("'")
             }
         }
--- a/src/name.rs
+++ b/src/name.rs
@@ -564,16 +564,17 @@ mod namespaces {
             let ns = Namespace(b"default");
 
             let mut resolver = NamespaceResolver::default();
+            let s = resolver.buffer.len();
 
             resolver.push(&BytesStart::from_content(" xmlns='default'", 0));
-            assert_eq!(resolver.buffer, b"default");
+            assert_eq!(&resolver.buffer[s..], b"default");
 
             // Check that tags without namespaces does not change result
             resolver.push(&BytesStart::from_content("", 0));
-            assert_eq!(resolver.buffer, b"default");
+            assert_eq!(&resolver.buffer[s..], b"default");
             resolver.pop();
 
-            assert_eq!(resolver.buffer, b"default");
+            assert_eq!(&resolver.buffer[s..], b"default");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(ns), LocalName(b"simple"))
@@ -593,11 +594,12 @@ mod namespaces {
             let new_ns = Namespace(b"new");
 
             let mut resolver = NamespaceResolver::default();
+            let s = resolver.buffer.len();
 
             resolver.push(&BytesStart::from_content(" xmlns='old'", 0));
             resolver.push(&BytesStart::from_content(" xmlns='new'", 0));
 
-            assert_eq!(resolver.buffer, b"oldnew");
+            assert_eq!(&resolver.buffer[s..], b"oldnew");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(new_ns), LocalName(b"simple"))
@@ -609,7 +611,7 @@ mod namespaces {
             assert_eq!(resolver.find(name), Bound(new_ns));
 
             resolver.pop();
-            assert_eq!(resolver.buffer, b"old");
+            assert_eq!(&resolver.buffer[s..], b"old");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(old_ns), LocalName(b"simple"))
@@ -631,11 +633,12 @@ mod namespaces {
             let old_ns = Namespace(b"old");
 
             let mut resolver = NamespaceResolver::default();
+            let s = resolver.buffer.len();
 
             resolver.push(&BytesStart::from_content(" xmlns='old'", 0));
             resolver.push(&BytesStart::from_content(" xmlns=''", 0));
 
-            assert_eq!(resolver.buffer, b"old");
+            assert_eq!(&resolver.buffer[s..], b"old");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Unbound, LocalName(b"simple"))
@@ -647,7 +650,7 @@ mod namespaces {
             assert_eq!(resolver.find(name), Unbound);
 
             resolver.pop();
-            assert_eq!(resolver.buffer, b"old");
+            assert_eq!(&resolver.buffer[s..], b"old");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(old_ns), LocalName(b"simple"))
@@ -671,16 +674,17 @@ mod namespaces {
             let ns = Namespace(b"default");
 
             let mut resolver = NamespaceResolver::default();
+            let s = resolver.buffer.len();
 
             resolver.push(&BytesStart::from_content(" xmlns:p='default'", 0));
-            assert_eq!(resolver.buffer, b"pdefault");
+            assert_eq!(&resolver.buffer[s..], b"pdefault");
 
             // Check that tags without namespaces does not change result
             resolver.push(&BytesStart::from_content("", 0));
-            assert_eq!(resolver.buffer, b"pdefault");
+            assert_eq!(&resolver.buffer[s..], b"pdefault");
             resolver.pop();
 
-            assert_eq!(resolver.buffer, b"pdefault");
+            assert_eq!(&resolver.buffer[s..], b"pdefault");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(ns), LocalName(b"with-declared-prefix"))
@@ -700,11 +704,12 @@ mod namespaces {
             let new_ns = Namespace(b"new");
 
             let mut resolver = NamespaceResolver::default();
+            let s = resolver.buffer.len();
 
             resolver.push(&BytesStart::from_content(" xmlns:p='old'", 0));
             resolver.push(&BytesStart::from_content(" xmlns:p='new'", 0));
 
-            assert_eq!(resolver.buffer, b"poldpnew");
+            assert_eq!(&resolver.buffer[s..], b"poldpnew");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(new_ns), LocalName(b"with-declared-prefix"))
@@ -716,7 +721,7 @@ mod namespaces {
             assert_eq!(resolver.find(name), Bound(new_ns));
 
             resolver.pop();
-            assert_eq!(resolver.buffer, b"pold");
+            assert_eq!(&resolver.buffer[s..], b"pold");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(old_ns), LocalName(b"with-declared-prefix"))
@@ -738,11 +743,12 @@ mod namespaces {
             let old_ns = Namespace(b"old");
 
             let mut resolver = NamespaceResolver::default();
+            let s = resolver.buffer.len();
 
             resolver.push(&BytesStart::from_content(" xmlns:p='old'", 0));
             resolver.push(&BytesStart::from_content(" xmlns:p=''", 0));
 
-            assert_eq!(resolver.buffer, b"poldp");
+            assert_eq!(&resolver.buffer[s..], b"poldp");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Unknown(b"p".to_vec()), LocalName(b"with-declared-prefix"))
@@ -754,7 +760,7 @@ mod namespaces {
             assert_eq!(resolver.find(name), Unknown(b"p".to_vec()));
 
             resolver.pop();
-            assert_eq!(resolver.buffer, b"pold");
+            assert_eq!(&resolver.buffer[s..], b"pold");
             assert_eq!(
                 resolver.resolve(name, true),
                 (Bound(old_ns), LocalName(b"with-declared-prefix"))
--- a/src/name.rs
+++ b/src/name.rs
@@ -390,7 +390,7 @@ impl NamespaceEntry {
 /// A namespace management buffer.
 ///
 /// Holds all internal logic to push/pop namespaces with their levels.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 pub(crate) struct NamespaceResolver {
     /// Buffer that contains names of namespace prefixes (the part between `xmlns:`
     /// and an `=`) and namespace values.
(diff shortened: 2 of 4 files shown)


src/name.rs
```diff
@@ -539,10 +539,17 @@ impl NamespaceResolver {
             .unwrap_or_else(|| Self::maybe_unknown(prefix))
     }
 
+    const XML_NAMESPACE: &'static [u8] = b"http://www.w3.org/XML/1998/namespace";
+    const XMLNS_NAMESPACE: &'static [u8] = b"http://www.w3.org/2000/xmlns/";
```
Mingun, line 543: I would like to see documentation on one of this consts about special processing of those prefixes with the link to the specification (https://www.w3.org/TR/xml-names11/#xmlReserved). Maybe just an adopted copy fro that specification.

Also, can we change type of these consts to `Namespace<'static>`?

I also prefer to have consts declared at the start of `impl` block, or you can declare them outside of `impl` block.

src/name.rs
```diff
@@ -539,10 +539,17 @@ impl NamespaceResolver {
...
+
     #[inline]
     fn maybe_unknown(prefix: Option<Prefix>) -> ResolveResult<'static> {
         match prefix {
-            Some(p) => ResolveResult::Unknown(p.into_inner().to_vec()),
+            Some(p) => match p.into_inner() {
+                b"xml" => ResolveResult::Bound(Namespace(NamespaceResolver::XML_NAMESPACE)),
+                b"xmlns" => ResolveResult::Bound(Namespace(NamespaceResolver::XMLNS_NAMESPACE)),
+                x => ResolveResult::Unknown(x.to_vec()),
+            },
```
Mingun, line 552: This looks better:
```suggestion
            Some(Prefix(b"xml")) => ResolveResult::Bound(Self::XML_NAMESPACE),
            Some(Prefix(b"xmlns")) => ResolveResult::Bound(Self::XMLNS_NAMESPACE),
            Some(p) => ResolveResult::Unknown(p.into_inner().to_vec()),
```
Also, need to add a TODO comment, that this will not work with UTF-16 and ISO_2022_JP encodings, because namespaces stored in document encoding. As written it will work only because all other [supported encodings](https://docs.rs/encoding_rs/0.8.31/encoding_rs/index.html#no-extensibility-by-design) are ASCII-compatible.

src/name.rs
```diff
@@ -806,6 +813,44 @@ mod namespaces {
         assert_eq!(resolver.find(name, &buffer), Unknown(b"unknown".to_vec()));
     }
 
+    #[test]
+    fn undeclared_xml_prefix() {
```
Mingun, line 817: I would prefer to create a submodule `builtin_prefix` with `xml` and `xmlns` tests, and additional tests for overriding (maybe with TODOs that overriding rules should be changed)

src/name.rs
```diff
@@ -806,6 +813,44 @@ mod namespaces {
...
+        let name = QName(b"xml:lang");
+
+        let resolver = NamespaceResolver::default();
+        let buffer = Vec::new();
+
+        assert_eq!(buffer, b"");
+        assert_eq!(
+            resolver.resolve(name, &buffer, true),
```
Mingun, line 825:
```suggestion
            resolver.resolve(name, b"", true),
```
Here and everywhere.

Mingun (changes requested):
Looks good, some minor notes for maintainability. Also, please add an entry to `Changelog.md` and probably one test to unit tests with a whole non-UTF-8 XML to ensure, that our expectations really works

wt:
Okay, so I have declared the predefined namespaces in what I believe is a better way and added some of the docs you asked for. Is this heading in the right direction? By my calculation, the following still needs to be done:

- [ ] tests for overriding xml namespace (were there other override tests you were looking for?)
- [ ] ???

src/name.rs
wt: I have pulled the definition of the prefixes and the actual mapping to be next to one another in the code. I have used the Prefixes in a more abstract manner that I think obviates the needs to TODO this code since changing the implementation of Prefix is what appears to be needed to fix the non-ascii compat encodings. Please let me know if you agree.

src/name.rs
wt: I documented the reserved namespaces and changed their types. I also moved them to lazily initalized static outside of the impl. I did this as I wanted to have a HashMap that mapped the prefix to the Namespace. I also wanted to define the prefix and appropriate namespace in the same part of the code. I ended up using the once_cell, which is due to be incorporated in core at some point (https://doc.rust-lang.org/core/cell/struct.OnceCell.html#method.eq). The core support for OnceCell is currently experimental, so I hope this (using the crate) is acceptable until it's stabilized as I think it's a much better organization for the code.

wt: https://doc.rust-lang.org/core/cell/struct.LazyCell.html is the location of the Lazy related stuff in core.

Mingun (changes requested):
For override tests: check the behavior of other mature XML libraries, for example, [Java](https://docs.oracle.com/en/java/javase/19/docs/api/java.xml/module-summary.html). Then write the tests in `tests/namespaces.rs`. For example, check this XML document:
```xml
<root xml:attribute="ok">
  <xml:element/>

  <override xmlns:xml="overrided non-overradable prefix" xml:attribute="oh!">
    <xml:element/>
  </override>

  <!-- remove standard prefix? -->
  <reset xmlns:xml="" xml:attribute="eh?">
    <xml:element/>
  </reset>
</root>
```
How other libraries will parse that document? If it will be parsed, what the `QName`s will be returned?

It also would be excellent if you add test to `tests/encoding.rs` with some non-UTF-8 document and check that we correctly process `xml:` and `xmlns:` attributes:
- check that we detect and use non-UTF-8 encoding by querying `.decoder()`
- check that we element and attribute with undeclared prefix resolved correctly (well, we can even made `XML_NAMESPACE` and `XMLNS_NAMESPACE` our public API -- as a `Namespace` constants)
- at the same time check, that other undeclared prefix will resolved as unknown

wt:
Just so we know, the std versions of the oncecell stuff that I used in this PR is being stabilized as we speak. Part of the work landed in 1.70. The lazy initialized types have not yet landed in stable. However, that may happen in one of the next couple releases. My plan is to update this PR when that lands.

dralley:
@wt Even once it's merged, I don't know that we want to immediately bump the minimum supported Rust version up to the latest one.  If that is the only thing holding this PR up, if it's ready and @Mingun likes it, then I would say just merge it sooner rather than later, and we can drop the dependency later on.

wt:
Can we just use the once_cell crate via [this](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html) then? This crate will provide backward compatibility until this is stabilized, and it has been around for while.

dralley:
>Can we just use the once_cell crate via [this](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html) then? This crate will provide backward compatibility until this is stabilized, and it has been around for while.

Yup that's exactly what I'm suggesting.  Use the `once_cell` crate, so that we don't have to wait for it to hit the stdlib and/or immediately bump the MSRV when it does.

src/name.rs
```diff
@@ -787,6 +818,32 @@ mod namespaces {
...
+    mod builtin_prefixes {
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        #[test]
+        fn undeclared_reserved_prefixes() {
+            let resolver = NamespaceResolver::default();
+            let tag = b"random";
+
+            for (prefix, namespace) in WELL_KNOWN_NAMESPACES.iter() {
```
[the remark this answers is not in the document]
dralley: @wt One other remaining review comment ^^

dralley:
Also this should have a changelog entry

dralley:
Looks like `once_cell` 1.18 has an MSRV of 1.60 whereas the current MSRV for this library is 1.52.  

I'm fine with bumping it up to 1.60  (released April 2022), that is still pretty reasonable.  Most linux distros will have something newer which is really the main constraint.  But that should get a changelog entry also.

wt:
Okay, so I hammered out a test xml parser with Xerces. It does in fact map to the well known namespaces to their appropriate URIs the same way this change does.

This doc:
```
<document
    xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
    xsi:noNamespaceSchemaLocation='document.xsd' 
    xml:id='fjdlksfjds'>
</document>
```

produces the following parsed attrs for the document tag:
```
--> [document], Child Element Count = 0
attr: http://www.w3.org/XML/1998/namespace:id
attr: http://www.w3.org/2000/xmlns/:xsi
attr: http://www.w3.org/2001/XMLSchema-instance:noNamespaceSchemaLocation
```

Also, the failing tests appear to not be caused by my change. Do you agree with this?

I am also trying to figure out how to address your comment about checking the encoding. Can you help me figure out what it is that you want there?

wt:
FWIW, Xerces also does not allow overriding the xml namespace to another URI or to no URI. I guess this is the behavior we would want in quick-xml?

And I added a changelog.

Here's a possible task list to finish this change:

- [ ] disallow overriding the xml and xmlns namespace prefixes
- [ ] add integration test that tests a full document
- [ ] properly handle non-ascii compatible encodings

Does that cover everything?

Mingun (changes requested):
> Can we just use the once_cell crate via [this](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html) then? This crate will provide backward compatibility until this is stabilized, and it has been around for while.

There is no need to use hashmap and, as a result, `once_cell`, or `Lazy`, at all. Just two entries can be just hardcoded in place of use via `match`, or you can use `&[(&str, &str)]` constant and use linear search -- this will be faster that any map on such sizes and will not require lazy construction and additional dependencies.

And I still prefer to test each well-known prefix individually (one test for prefix, you can write a small macro if you don't like to duplicate code).

Mingun:
Those failures appeared since serde v1.0.181 due to https://github.com/serde-rs/serde/pull/2505:
```
failures:
    enum_::adjacently_tagged::nested_struct::elements
    enum_::adjacently_tagged::newtype::elements
    enum_::adjacently_tagged::struct_::elements
    enum_::adjacently_tagged::tuple_struct::elements
    enum_::adjacently_tagged::unit::elements
```
I've create https://github.com/tafia/quick-xml/issues/630 for investigation.

wt:
The MSRV test is weird. Why is is failing for 1.52 when the min stated version is 1.60 in this change?

dralley:
>The MSRV test is weird. Why is is failing for 1.52 when the min stated version is 1.60 in this change?

It's hardcoded https://github.com/tafia/quick-xml/blob/master/.github/workflows/rust.yml#L17

But I suppose it's not necessary to change if you go w/ the linear search approach.  With only a small number of strings to check, @Mingun is correct that it's not really going to be worth the Hashmap

wt:
I'll rewrite without the HashMap stuff and change the tests to be now explicit.

Do y'all agree that the semantics around using the xml and xmlns prefixes should be enforced as they are in Java and Xerces?

wt:
I just pushed a new change that uses no HashMap. Please let me know if this addresses the concerns on the design.

Also, I think this makes my task list above accurate. Are we in agreement on that?

src/name.rs
```diff
@@ -399,6 +399,24 @@ pub(crate) struct NamespaceResolver {
...
+/// undeclared or bound to any other namespace name. Other prefixes must not be bound to this
+/// namespace name, and it must not be declared as the default namespace.
+///
+/// The prefix `xmlns` is used only to declare namespace bindings and is by definition bound
+/// to the namespace name http://www.w3.org/2000/xmlns/. It must not be declared or
+/// undeclared. Other prefixes must not be bound to this namespace name, and it must not be
+///  declared as the default namespace. Element names must not have the prefix xmlns.
+///
+/// [reserved namespaces]: https://www.w3.org/TR/xml-names11/#xmlReserved
+static WELL_KNOWN_NAMESPACES: &'static [[&str; 2]] = &[
```
dralley, line 415: I'd just call it `RESERVED_NAMESPACES`

Mingun (commented):
> FWIW, Xerces also does not allow overriding the xml namespace to another URI or to no URI. I guess this is the behavior we would want in quick-xml?

What action Xerces did when it encounters such situation? It returns error while parsing or just does nothing?

> - [ ] disallow overriding the xml and xmlns namespace prefixes

With the current implementation this is already done if our desired action is "do nothing". But if Xerces returns an error then I think we should add a setting to replicate Xerces behavior (a setting "do nothing / produce error").

> - [ ] add integration test that tests a full document

Well, if you do that, fine, but I'm also OK to proceed without it.

> - [ ] properly handle non-ascii compatible encodings

I think that this task will impossible to implement here, so do not bother about it

> I just pushed a new change that uses no HashMap. Please let me know if this addresses the concerns on the design.

Yes, that's much better. Although in my opinion, even the constant is superfluous, readability is better without it (i.e. if you embed everything in `match`):
```rust
    /// Constants in `match` define the [reserved namespaces] for the xml standard.
    ///
    /// The prefix `xml` is by definition bound to the namespace name
    /// `http://www.w3.org/XML/1998/namespace`. It may, but need not, be declared, and must not be
    /// undeclared or bound to any other namespace name. Other prefixes must not be bound to this
    /// namespace name, and it must not be declared as the default namespace.
    ///
    /// The prefix `xmlns` is used only to declare namespace bindings and is by definition bound
    /// to the namespace name `http://www.w3.org/2000/xmlns/`. It must not be declared or
    /// undeclared. Other prefixes must not be bound to this namespace name, and it must not be
    /// declared as the default namespace. Element names must not have the prefix `xmlns`.
    ///
    /// [reserved namespaces]: https://www.w3.org/TR/xml-names11/#xmlReserved
    #[inline]
    fn maybe_unknown(prefix: Option<Prefix>) -> ResolveResult<'static> {
        match prefix {
            Some(Prefix(b"xml")) => {
                ResolveResult::Bound(Namespace(b"http://www.w3.org/XML/1998/namespace"))
            }
            Some(Prefix(b"xmlns")) => {
                ResolveResult::Bound(Namespace(b"http://www.w3.org/2000/xmlns/"))
            }

            Some(p) => ResolveResult::Unknown(p.into_inner().to_vec()),
            None => ResolveResult::Unbound,
        }
    }
```

Mingun:
Oh, and you can rebase on current master to not suffer from failing tests

wt:
rebased

dralley:
Do you happen to know how libxml2 and/or expat handles them?

wt:
I don't know how the other libs work.

wt:
Looks like expat does a similar validation around [here](https://github.com/libexpat/libexpat/blob/20293f2ad29d5a5ebe126174e5bb961962540b81/expat/lib/xmlparse.c#L3834). By my reading errors are returned in the same cases as for Xerces.

src/name.rs
wt: I'm planning on moving the logic depending on our agreement on the error behavior. I'm thinking this change may not use this constant, so I am going to try to get agreement on the error behavior and then rewrite with that in mind.

wt:
Alright, this newest iteration completely changes how the implementation to insert the xml and xmlns namespace entries in the root namespace resolver. I have not yet written tests to cover all the new semantics when pushing a new namespace context, so this change is in progress.

Having said that, can y'all please make sure this implementation is heading in an acceptable direction?

wt:
rebased

dralley:
@wt There's are a few small compilation errors to address.

@Mingun Does this implementation strategy look reasonable to you?

wt:
Fixed the lint issues for the compilation. I ran the individual tests, but no `cargo test` after the previous rebase.

src/errors.rs
```diff
@@ -9,6 +9,16 @@ use std::str::Utf8Error;
 use std::string::FromUtf8Error;
 use std::sync::Arc;
 
+#[derive(Clone, Debug)]
+pub struct ReservedNamespacePrefixError {
+    pub prefix: String,
+}
```
dralley, line 15: Is there a particular reason to make separate structs instead of directly embedding the strings in the enum variants?  I don't think we use this pattern in many, if any, places.

src/name.rs
```diff
@@ -399,12 +399,61 @@ pub(crate) struct NamespaceResolver {
...
+            },
+            buffer,
+        )
+    }
     /// Begins a new scope and add to it all [namespace bindings] that found in
     /// the specified start element.
     ///
     /// [namespace binding]: https://www.w3.org/TR/xml-names11/#dt-NSDecl
-    pub fn push(&mut self, start: &BytesStart, buffer: &mut Vec<u8>) {
+    pub fn push(&mut self, start: &BytesStart, buffer: &mut Vec<u8>) -> Result<()> {
```
dralley, line 456: This is a breaking change (which is fine) but it should probably have its own callout in the changelog.

src/name.rs
```diff
@@ -399,12 +399,61 @@ pub(crate) struct NamespaceResolver {
...
+const RESERVED_NAMESPACE_XMLNS: (Prefix, Namespace) = (
+    Prefix(b"xmlns"),
+    Namespace(b"http://www.w3.org/2000/xmlns/"),
+);
+
+const RESERVED_NAMESPACES: [(Prefix, Namespace); 2] =
+    [RESERVED_NAMESPACE_XML, RESERVED_NAMESPACE_XMLNS];
+
 impl NamespaceResolver {
+    pub fn new_root_resolver_and_buf() -> (Self, Vec<u8>) {
```
dralley, line 428: I have *not* done a thorough search and @Mingun is the expert on this code not me, but it seems that the buffer is always? used alongside the `NamespaceResolver`, in which case, is there a reason it can't be part of the `NamespaceResolver` rather than being passed around?

(I may have missed somewhere where it was being used elsewhere)

wt: I was wondering the same, but I really didn't have an interest in making a change to that design as part of this change.

wt:
Okay, I think the latest version is ready. I've added tests for adding namespace contexts to the resolver. I added these specific tests:

* xml prefix explicit sets work when the uri is correct
* xml prefix sets fail when setting to incorrect url
* setting other prefixes to the xml uri fails
* setting xmlns prefix explicitly fails
* setting other prefixes to the xmlns uri fails
* setting xml to the xmlns prefix fails (I added this as the logic is overlapping in a way that is easy to biff)

I believe that this is ready to a review. @Mingun would you mind taking a look? I would love to get this landed soon.

src/errors.rs
wt: I've collapsed them into a single struct. I think improving the errors should probably be left for another change. I hope this addresses your concern.

src/name.rs
wt: Does this require a major version bump? Also, should I add a completely separate entry for the return type change, or should I add it to the existing entry?

Mingun: No, `NamespaceResolver` is a private struct, so this is not a breaking change

wt: Given that, I'm assuming that this doesn't need to be cashed out in the changelog.

dralley: yes, apologies.  oversight on my part

Mingun:
Don't worry, I don't forgot about this. I was on vacation so tracked issues very lazily. I'll look during this week.

Mingun:
Thanks for working on this PR and investigating all stuff about handling this in other libraries!

I'm not finishing my review yet. Instead of giving feedback in the form of comments, I decided that it would be easier to implement my vision myself. Changes I have broken down into several commits so you can see the reason for them.

I structurized tests and added missing parts (what became apparent after structuring). I also simplified checks and now, I think, it is more easily to understand the code.

I plan to finish review next week.

wt:
I'm glad you find some improvement. I'm not picky as long as the functionality is there. When do you think this will land in the main branch?

Mingun:
I think this week

src/name.rs
Mingun: > I have _not_ done a thorough search and @Mingun is the expert on this code not me, but it seems that the buffer is always? used alongside the `NamespaceResolver`, in which case, is there a reason it can't be part of the `NamespaceResolver` rather than being passed around?

Yes, I also noticed this and will change this in that PR

Mingun (approved):
Well, I think, it is done. I'll merge it next day if there will be no objections

PR #552, opened 2023-02-11
Parser position tracking

### LINKED ISSUE #109: A way to get the row/column of a Reader
xml-rs provides [`TextPosition`](https://docs.rs/xml-rs/0.7.0/xml/common/struct.TextPosition.html) for this, while quick-xml only has [`Reader::buffer_position()`](https://docs.rs/quick-xml/0.11.0/quick_xml/reader/struct.Reader.html#method.buffer_position) to get the byte position. It would be useful for error reporting if there was a way to compute the row and column from this (preferably without overhead if unused).

Themayu:
# Parser position tracking
Store an `Option<quick_xml::reader::Span>` inside each of the `quick_xml::events::Bytes*` structs. This span is computed by the parser during reading, and passed into the relevant `wrap()` function during value construction. It can be accessed by calling relevant `span()` methods, which will clone and return the span (`Range`, sadly, does not implement `Copy` due to its status as an `Iterator` implementation.)

~~Additionally, the derive implementations of `PartialEq` on the structs is replaced with a manual implementation of `PartialEq`. This is to prevent previously `true` equality comparisons such as `bytes == BytesStart::new("...", 4)` suddenly changing their behaviour to return `false`. This does mean that we lose our implementations of `StructuralPartialEq`, meaning that constants of the relevant types can no longer be used as patterns in match arms. However, since `quick_xml` exposes no way to construct values of these types at compile time, and since `StructuralPartialEq` is currently nightly-only and therefore cannot be used in trait bounds in stable Rust, I do not believe this change is sufficiently observable to require a major version bump under semver.~~ Paragraph rendered obsolete by design change.

Other than the above, I do not intend to make any other possibly breaking changes to the public API with this contribution.

## Changes
- Each of the `Bytes*` structs under `quick_xml::events` receives a new field, `span: crate::reader::Span`.
  - If created via parsing, this field will store a value of `n..m`, representing the start and end positions of that event's contents - for tags, this is the positions of the `<` and `>` character; while for text, it is the start and end of the text content, with respect to optional whitespace trimming.
  - If created via the API, this field will store `0..0`.
- Each of the structs receives a new trait, `Spanned`, which defines methods to retrieve and update the stored span.
- Altering data for a given event, such as its element name or adding attributes, does not alter the span it is given. This is because such edits do not alter the source it was initially read from.

## Progress
- [X] Events
  - [X] `quick_xml::events::BytesStart`
  - [X] `quick_xml::events::BytesDecl`
    - As far as I can tell, this one can just reuse the span inside `BytesStart`
  - [X] `quick_xml::events::BytesEnd`
  - [X] `quick_xml::events::BytesText`
  - [X] `quick_xml::events::BytesCData`
  - [X] Method on `quick_xml::events::Event` to return the underlying span if there is one
- [ ] Attributes
  - [ ] `quick_xml::events::attributes::Attributes`
  - [ ] `quick_xml::events::attributes::Attribute`
  - [ ] `quick_xml::events::attributes::Attr`
- [ ] Events
  - [ ] `quick_xml::errors::Error`
  - [ ] `quick_xml::escape::EscapeError`
  - [ ] `quick_xml::events::attributes::AttrError`
- [ ] Namespaces _(unsure whether this one makes sense)_
  - [ ] `quick_xml::name::QName`
  - [ ] `quick_xml::name::LocalName`
  - [ ] `quick_xml::name::Prefix`
  - [ ] `quick_xml::name::Namespace`
- [X] Figure out where to add tests for the new functionality
- [ ] Edit Changelog.md with new functionality

## PR changelog
- **2023/02/16 18:31 GMT**: Rebased PR source on an incomplete implementation that will be continued.
- **2023/02/17 07:16 GMT**: Restructured Progress section, added mini-changelog to PR.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,6 +63,10 @@ jobs:
       env:
         LLVM_PROFILE_FILE: coverage/serialize-escape-html-%p-%m.profraw
       run: cargo test --features serialize,escape-html
+    - name: Run tests (serialize+span)
+      env:
+        LLVM_PROFILE_FILE: coverage/serialize-span-%p-%m.profraw
+      run: cargo test --features serialize,span
     - name: Run tests (all features)
       env:
         LLVM_PROFILE_FILE: coverage/all-features-%p-%m.profraw
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -180,6 +180,10 @@ serde-types = ["serde/derive"]
 ## If you need that, use the `serde-types` feature.
 serialize = ["serde"] # "dep:" prefix only avalible from Rust 1.60
 
+## Enables gathering spans in events, which can slowdown parser a bit and increase
+## memory consumption for events
+span = []
+
 [package.metadata.docs.rs]
 # document all features
 all-features = true
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -1885,6 +1885,11 @@ mod map;
 mod simple_type;
 mod var;
 
+#[cfg(feature = "span")]
+use crate::events::Spanned;
+#[cfg(feature = "span")]
+use crate::reader::Span;
+
 pub use crate::errors::serialize::DeError;
 use crate::{
     encoding::Decoder,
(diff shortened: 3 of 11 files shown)


Themayu:
Rebased PR source on an incomplete implementation by @Mingun. Updated PR details to reflect change in design.

PR #554, opened 2023-02-16
perf(fuzz): Improves code coverage of fuzzers

silvergasp:
Updates fuzz target to include;
- A round trip writer
- A slice reader
- Pretty printing, debug and display code

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,25 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   steps:
+   - name: Build Fuzzers
+     id: build
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'quick-xml'
+       language: rust
+   - name: Run Fuzzers
+     uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'quick-xml'
+       language: rust
+       fuzz-seconds: 600
+   - name: Upload Crash
+     uses: actions/upload-artifact@v3
+     if: failure() && steps.build.outcome == 'success'
+     with:
+       name: artifacts
+       path: ./out/artifacts
+
--- a/fuzz/fuzz_targets/fuzz_target_1.rs
+++ b/fuzz/fuzz_targets/fuzz_target_1.rs
@@ -1,44 +1,98 @@
 #![no_main]
 use libfuzzer_sys::fuzz_target;
+use std::hint::black_box;
 
-use quick_xml::events::Event;
-use quick_xml::reader::Reader;
+use quick_xml::{events::Event, reader::Reader, writer::Writer};
 use std::io::Cursor;
 
-fuzz_target!(|data: &[u8]| {
-    // fuzzed code goes here
-    let cursor = Cursor::new(data);
-    let mut reader = Reader::from_reader(cursor);
+macro_rules! debug_format {
+    ($x:expr) => {
+        let _unused = std::hint::black_box(format!("{:?}", $x));
+    };
+}
+
+fn round_trip<R>(reader: &mut Reader<R>) -> ()
+where
+    R: std::io::BufRead,
+{
+    let mut writer = Writer::new(Cursor::new(Vec::new()));
     let mut buf = vec![];
+    let reader = reader
+        .expand_empty_elements(true)
+        .trim_text(true)
+        .trim_text_end(true);
     loop {
-        match reader.read_event_into(&mut buf) {
+        let event_result = reader.read_event_into(&mut buf);
+        if let Ok(ref event) = event_result {
+            let _event = black_box(event.borrow());
+            let _event = black_box(event.as_ref());
+            debug_format!(event);
+            debug_format!(writer.write_event(event));
+        }
+        match event_result {
             Ok(Event::Start(ref e)) | Ok(Event::Empty(ref e)) => {
+                debug_format!(e);
+                debug_format!(e.name());
                 for a in e.attributes() {
+                    debug_format!(a);
                     if a.ok().map_or(false, |a| a.unescape_value().is_err()) {
                         break;
                     }
                 }
             }
-            Ok(Event::Text(ref e)) | Ok(Event::Comment(ref e))
+            Ok(Event::Text(ref e))
+            | Ok(Event::Comment(ref e))
             | Ok(Event::PI(ref e))
             | Ok(Event::DocType(ref e)) => {
-                if e.unescape().is_err() {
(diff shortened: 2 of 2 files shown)


fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -1,44 +1,98 @@
...
+use quick_xml::{events::Event, reader::Reader, writer::Writer};
 use std::io::Cursor;
 
-fuzz_target!(|data: &[u8]| {
-    // fuzzed code goes here
-    let cursor = Cursor::new(data);
-    let mut reader = Reader::from_reader(cursor);
+macro_rules! debug_and_pretty_print {
+    ($x:expr) => {
+        let _unused = std::hint::black_box(format!("{:?} {:#?}", $x, $x));
```
dralley, line 10: I'm not against it, but I'm not sure if this will actually increase test coverage, as pretty printing is handled withing the `Formatter` object in uses where it is derived automatically, and in the case of our manual implementations, they are formatted as tuple-structs with no alternate pretty formatting.

dralley: I don't think the macro should have `print` in the name if it's not printing anything though.  Maybe replace with `*format`.

Mingun:
I correctly understand, that the purpose of this PR is to call as much methods as we can for fuzzing them? Would it be valuable to test each method in it's own `fuzz_target!` or that would be an overkill?

I also noted, that we do not run fuzzing in our CI. Could you add a job for that to the [workflow](https://github.com/tafia/quick-xml/blob/master/.github/workflows/rust.yml)?

silvergasp:
Yeah, that's right, this fuzz harness is just calling as much functionality as possible to find ways that it could crash. . I think that there could be some value in splitting this fuzz harness up into 1 or more logical units. But I think it'll be impractical to do it as 1 function per `fuzz_target!`. This is mostly because there is some nested logic, e.g. in order to reach some functions you first need to match on a result etc.

There are other ways of fuzzing, for example, you can do property-based fuzz testing. In the context of quickxml this might look like generate_arbitrary_xml->parse_xml then assert that the parsed xml is identical to the generated xml. I'd be happy to open another PR doing this, although this would required some more intrusive changes to quick-xml, namely implementing [arbitrary](https://github.com/rust-fuzz/arbitrary) for the quick-xml types.

I'd be happy add this to the CI. As quick-xml is already integrated with google/oss-fuzz might I suggest adding an [oss-fuzz specific github action instead?](https://google.github.io/oss-fuzz/getting-started/continuous-integration/)

Mingun:
Ok. I just have very limited knowledge about best practices in fuzzing, so I want to hear suggestions from more advanced users. Then the current implementation is fine.

> I'd be happy add this to the CI. As quick-xml is already integrated with google/oss-fuzz might I suggest adding an [oss-fuzz specific github action instead?](https://google.github.io/oss-fuzz/getting-started/continuous-integration/)

Seems reasonable. Do it.

> I'd be happy to open another PR doing this, although this would required some more intrusive changes to quick-xml, namely implementing [arbitrary](https://github.com/rust-fuzz/arbitrary) for the quick-xml types.

I think it is reasonable change, just make new dependency optional and run fuzzing with a special feature enabled.

silvergasp:
> > I'd be happy add this to the CI. As quick-xml is already integrated with google/oss-fuzz might I suggest adding an [oss-fuzz specific github action instead?](https://google.github.io/oss-fuzz/getting-started/continuous-integration/)
> 
> Seems reasonable. Do it.


Done

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   steps:
+   - name: Build Fuzzers
+     id: build
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'quick-xml'
+       language: c++
```
Mingun, line 12: Shouldn't we change that to `rust`?

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,25 @@
...
+   - name: Run Fuzzers
+     uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'quick-xml'
+       language: c++
```
Mingun, line 17: and here?

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,25 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   steps:
+   - name: Build Fuzzers
+     id: build
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
```
Mingun, line 9: Is there a normal practice to depend on `master` for that job? The tagged version would be better, as I think (but I maybe wrong)

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,25 @@
...
+     with:
+       oss-fuzz-project-name: 'quick-xml'
+       language: c++
+       fuzz-seconds: 600
+   - name: Upload Crash
+     uses: actions/upload-artifact@v3
+     if: failure() && steps.build.outcome == 'success'
+     with:
+       name: artifacts
+       path: ./out/artifacts
```
Mingun, line 24: That name / path has some concrete meaning for the job? Maybe better to name them `crashes` if the concrete name do not matter?

.github/workflows/cifuzz.yml
silvergasp: This is normal practice yeah. Although now that you mention it, it does seem a little strange that this is "normal practice".

.github/workflows/cifuzz.yml
silvergasp: oops, yeah good catch!

.github/workflows/cifuzz.yml
silvergasp: good catch!

.github/workflows/cifuzz.yml
silvergasp: Yeah it has a specific meaning in the context of fuzzing. Artifacts encompass more than just crashes. e.g. crashes, out-of-memory errors, timeouts (e.g. triggered by an infinite loop) etc. I think even if we could change the name of this directory (which oss-fuzz won't like), it wouldn't make sense to as some of the outputs here aren't necessarily just crashes. There is some documentation for artifacts [here](https://www.llvm.org/docs/LibFuzzer.html).

.github/workflows/cifuzz.yml
silvergasp: There are some docs here https://google.github.io/oss-fuzz/getting-started/continuous-integration/

Mingun (approved):

PR #556, opened 2023-02-18
Add helper functions `to_*_with_root` that writes XML with the specified root tag name

Mingun:
Also change existing `to_writer` and `to_string` functions to accept `?Sized` types (by default all generic parameters has `Sized` constraint). This is matches the other serde libraries.

### DIFF
--- a/src/se/mod.rs
+++ b/src/se/mod.rs
@@ -87,14 +87,168 @@ use serde::serde_if_integer128;
 use std::fmt::Write;
 use std::str::from_utf8;
 
-/// Serialize struct into a `Write`r
-pub fn to_writer<W: Write, S: Serialize>(writer: W, value: &S) -> Result<W, DeError> {
+/// Serialize struct into a `Write`r.
+///
+/// # Examples
+///
+/// ```
+/// # use quick_xml::se::to_writer;
+/// # use serde::Serialize;
+/// # use pretty_assertions::assert_eq;
+/// #[derive(Serialize)]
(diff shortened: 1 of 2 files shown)


src/se/mod.rs
```diff
@@ -87,16 +87,170 @@ use serde::serde_if_integer128;
...
+///     // This will not work for other types or struct with #[serde(flatten)] fields
+///     "<Root attribute=\"attribute content\">\
+///         <element>element content</element>\
+///         text content\
+///     </Root>"
+/// );
+/// ```
+pub fn to_writer<W, T>(writer: W, value: &T) -> Result<W, DeError>
+where
+    W: Write,
```
dralley, line 125: I'd prefer that in cases where we use `std::fmt::Write`, we do so explicitly by using `fmt::Write`.  Otherwise it is too easy to mix up `io::Write` and `fmt:Write`.

dralley: In the same vein it feels as though `to_writer()` may be ambiguous to some users, however I admit that I cannot think of any better name, so it's OK.

dralley (approved):
Apart from my suggestion, this looks good.

src/se/mod.rs
Mingun: Changing `Write` to `fmt::Write` does not change the generated documentation (which show trait as `Write`), so this change does not give much effect. Because we use only one `Write` in the serialization modules, I think, that using simple name will not confuse developers.

PR #568, opened 2023-03-01
Add accessors to underlying writer: .get_ref() and .get_mut()

### LINKED ISSUE #566: Add `.get_ref()` / `.get_mut()` for `Writer<W>`
In my case, I wrap the inner of `quick_xml::Writer<W>` into another struct that implements `std::io::Write`.
So I end up with `quick_xml::Writer<my::Writer<W>>`.

Currently, it's possible to retrieve the `&mut W` via `.inner()`, but not `&W`. 
An example of such a method in the Rust Standard Library: [std::io::BufWriter](https://doc.rust-lang.org/src/std/io/buffered/bufwriter.rs.html#232-235).

martsokha:
Not sure about `.get_mut()` as the writer already has `.inner()`.
Closes #566.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -81,19 +81,14 @@ impl<W: Write> Writer<W> {
         self.writer
     }
 
-    /// Get inner writer, keeping ownership
-    pub fn inner(&mut self) -> &mut W {
+    /// Get a mutable reference to the underlying writer.
+    pub fn get_mut(&mut self) -> &mut W {
         &mut self.writer
     }
 
     /// Get a reference to the underlying writer.
     pub fn get_ref(&self) -> &W {
         &self.writer
-    }
-
-    /// Get a mutable reference to the underlying writer.
-    pub fn get_mut(&mut self) -> &mut W {
-        &mut self.writer
     }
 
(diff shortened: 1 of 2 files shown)


src/writer.rs
```diff
@@ -86,6 +86,16 @@ impl<W: Write> Writer<W> {
...
 
+    /// Get a reference to the underlying writer.
+    pub fn get_ref(&self) -> &W {
+        &self.writer
+    }
+
+    /// Get a mutable reference to the underlying writer.
+    pub fn get_mut(&mut self) -> &mut W {
+        &mut self.writer
+    }
```
Mingun, line 97: Just rename existing `inner` method and mention both changes in the changelog

Mingun (changes requested):
We already have breaking changes in the upcoming release so one more makes no difference.

Mingun:
Please also update `Changelog.md`

Mingun:
Almost all right, just add a line with the link destination at line 55 in changelog.md:
```markdown
[#568]: https://github.com/tafia/quick-xml/pull/568
```

Mingun (approved):

PR #569, opened 2023-03-02
Make read_event_into_async an async-fn

vilunov:
Unless I'm missing something, this compiles and passes test just fine. Found out it while digging for the reason why my future suddenly becomes non-Send.

### DIFF
--- a/src/reader/async_tokio.rs
+++ b/src/reader/async_tokio.rs
@@ -2,8 +2,6 @@
 //! as underlying byte stream. This reader fully implements async/await so reading
 //! can use non-blocking I/O.
 
-use std::future::Future;
-use std::pin::Pin;
 use tokio::io::{self, AsyncBufRead, AsyncBufReadExt};
 
 use crate::events::Event;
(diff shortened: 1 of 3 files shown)


vilunov:
I realised that I forgot to enable the tokio feature-flag on my local copy. Tried to fix this stuff by rewriting it as a loop instead of recursion, it indeed infers the future as `Send`able now. The alternative would be to duplicate a bunch of methods in `Reader` and `NsReader`, and I didn't really like it.

src/reader/async_tokio.rs
```diff
@@ -81,19 +81,52 @@ impl<R: AsyncBufRead + Unpin> Reader<R> {
...
+                        if self.parser.encoding.can_be_refined() {
+                            self.parser.encoding =
+                                crate::reader::EncodingRef::BomDetected(encoding);
+                        }
+                    }
+                    #[cfg(not(feature = "encoding"))]
+                    (TokioAdapter(&mut self.reader)).remove_utf8_bom().await?;
+                    match self.read_until_open_async(buf).await? {
+                        Ok(ev) => break Ok(ev),
+                        Err(b) => buf = b,
```
vilunov, line 116: I should probably add that this trick is to circumvent [this limitation](https://github.com/rust-lang/rfcs/blob/master/text/2094-nll.md#problem-case-3-conditional-control-flow-across-functions) in pre-non-lexical borrowck

Mingun:
So again, what this PR does except changing the style of the method definition (which of course is great and I like it, but I want to understand if we need to add something valuable to the changelog)? Also, I would like to not duplicate actual realization. Is it possible to keep macro so we have only one place where parsing logic is located?

vilunov:
We are also making the resulting future `Send` if the writer is `Send`, which makes it possible to use the library with multithreaded runtime such as tokio, and this was my primary motivation.

Give me some time, I'll try to rewrite the macro. Also, this probably should be merged into the next 0.28 major release, as it breaks the public API.

Mingun:
Yes, I think I want to see that in the 0.28 release.

src/reader/async_tokio.rs
```diff
@@ -81,19 +79,17 @@ impl<R: AsyncBufRead + Unpin> Reader<R> {
     /// ```
     ///
     /// [`read_event_into()`]: Reader::read_event_into
-    pub fn read_event_into_async<'reader, 'b: 'reader>(
+    pub async fn read_event_into_async<'reader, 'b: 'reader>(
```
Mingun, line 82: Remove lines 31-38 from the docstring, as it no longer relevant.

Remove the `'reader` lifetime as there no reasons to name it explicitly

src/reader/async_tokio.rs
```diff
@@ -155,7 +151,13 @@ impl<R: AsyncBufRead + Unpin> Reader<R> {
     }
 
     /// Read until '<' is found, moves reader to an `OpenedTag` state and returns a `Text` event.
-    async fn read_until_open_async<'b>(&mut self, buf: &'b mut Vec<u8>) -> Result<Event<'b>> {
+    ///
+    /// Returns inner Ok if the loop should be broken and an event returned.
+    /// Returns inner Err with the same [buf] because Rust borrowck stumbles upon this case in particular.
```
Mingun, line 156:
```suggestion
    /// Returns inner `Ok` if the loop should be broken and an event returned.
    /// Returns inner `Err` with the same `buf` because Rust borrowck stumbles upon this case in particular.
```

src/reader/mod.rs
```diff
@@ -593,15 +605,18 @@ impl<R> Reader<R> {
...
         R: XmlSource<'i, B>,
     {
         read_event_impl!(self, buf, self.reader, read_until_open, read_until_close)
     }
 
     /// Read until '<' is found, moves reader to an `OpenedTag` state and returns a `Text` event.
-    fn read_until_open<'i, B>(&mut self, buf: B) -> Result<Event<'i>>
+    ///
+    /// Returns inner Ok if the loop should be broken and an event returned.
+    /// Returns inner Err with the same [buf] because Rust borrowck stumbles upon this case in particular.
```
Mingun, line 618:
```suggestion
    /// Returns inner `Ok` if the loop should be broken and an event returned.
    /// Returns inner `Err` with the same `buf` because Rust borrowck stumbles upon this case in particular.
```

src/reader/mod.rs
```diff
@@ -166,34 +166,46 @@ macro_rules! read_event_impl {
...
+                ParseState::OpenedTag => break $self.$read_until_close($buf) $(.$await)?,
+                ParseState::Empty => break $self.parser.close_expanded_empty(),
+                ParseState::Exit => break Ok(Event::Eof),
+            };
         };
         match event {
             Err(_) | Ok(Event::Eof) => $self.parser.state = ParseState::Exit,
             _ => {}
-        }
+        };
```
Mingun, line 208: Seems not necessary
```suggestion
        }
```

Mingun (changes requested):
Good work! Please add an entry to Changelog.md. Each entry is started with an issue/PR number, the links are resolved in the end of chapter.

Mingun (approved):

PR #571, opened 2023-03-04
Borrow element names when deserialize using serde

Mingun:
Makes it possible to deserialize 
```xml
<root>
  <element>element content<element>
</root>
```
into `HashMap<&str, &str>` (`"element" => "element content"` in this case), for example

### DIFF
--- a/src/de/key.rs
+++ b/src/de/key.rs
@@ -90,10 +90,22 @@ impl<'i, 'd> QNameDeserializer<'i, 'd> {
     }
 
     /// Creates deserializer from name of an element
-    pub fn from_elem(name: QName<'d>, decoder: Decoder) -> Result<Self, DeError> {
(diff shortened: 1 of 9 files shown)


src/de/key.rs
```diff
@@ -60,18 +61,19 @@ fn decode_name<'n>(name: QName<'n>, decoder: Decoder) -> Result<Cow<'n, str>, De
 ///
 /// `deserialize_any()` returns the same result as `deserialize_identifier()`.
 ///
-/// # Lifetime
+/// # Lifetimes
 ///
+/// - `'i`: lifetime of the data that deserializer borrow from the parsed input
```
dralley, line 66:
```suggestion
/// - `'i`: lifetime of the data that the deserializer borrows from the parsed input
```

dralley (approved):

PR #572, opened 2023-03-04
Introduce dedicated `Text` type for `DeEvent::Text` event

Mingun:
The new type would be useful in corrent implementation of #561 and #552, because introduces a place for additional data (a range of text that we should get after trim (assumed that Text would contain untrimmed text) and a span of a text)

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -2066,42 +2066,6 @@ impl<'i, R: XmlRead<'i>> XmlReader<'i, R> {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-/// A structure that deserializes XML into Rust values.
-pub struct Deserializer<'de, R>
-where
-    R: XmlRead<'de>,
-{
-    /// An XML reader that streams events into this deserializer
-    reader: XmlReader<'de, R>,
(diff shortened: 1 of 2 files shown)


src/de/mod.rs
```diff
@@ -2955,12 +2989,12 @@ mod tests {
...
                     // This comment here to keep the same formatting as others
                     // otherwise rustfmt suggest one-line it
-                    Text(Cow::Borrowed("text")),
+                    Text("text".into()),
                     End(BytesEnd::new("inner")),
                 ]
             );
             assert_eq!(de.write, vec![]);
-            assert_eq!(de.next().unwrap(), Text(Cow::Borrowed("text")));
+            assert_eq!(de.next().unwrap(), Text("text".into()));
```
dralley, line 2997: I'm assuming we don't care whether it was borrowed or owned?  (not that this aspect is being properly tested currently)

dralley (approved):

src/de/mod.rs
Mingun: Yes, we don't care. Only the fact that `Text` event is generated is tested here

PR #573, opened 2023-03-05
Implement basic async writer

vilunov:
This adds support for async byte writers in the already existing `Writer` type. I have not added support for indentation here, that's why only the `new` construct has loosened type bounds, but it can be added in a follow-up merge request.

### DIFF
--- a/src/writer/async_tokio.rs
+++ b/src/writer/async_tokio.rs
@@ -1,0 +1,69 @@
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+
+use crate::errors::Result;
+use crate::events::Event;
+use crate::Writer;
+
+impl<W: AsyncWrite + Unpin> Writer<W> {
+    /// Writes the given event to the underlying writer. Async version of [Writer::write_event].
+    pub async fn write_event_async<'a, E: AsRef<Event<'a>>>(&mut self, event: E) -> Result<()> {
+        match *event.as_ref() {
+            Event::Start(ref e) => self.write_wrapped_async(b"<", e, b">").await,
+            Event::End(ref e) => self.write_wrapped_async(b"</", e, b">").await,
+            Event::Empty(ref e) => self.write_wrapped_async(b"<", e, b"/>").await,
+            Event::Text(ref e) => self.write_async(e).await,
+            Event::Comment(ref e) => self.write_wrapped_async(b"<!--", e, b"-->").await,
+            Event::CData(ref e) => {
+                self.write_async(b"<![CDATA[").await?;
+                self.write_async(e).await?;
+                self.write_async(b"]]>").await
+            }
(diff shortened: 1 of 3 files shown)


src/writer/async_tokio.rs
```diff
@@ -0,0 +1,69 @@
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+
+use crate::errors::Result;
+use crate::events::Event;
+use crate::Writer;
+
+impl<W: AsyncWrite + Unpin> Writer<W> {
+    /// Writes the given event to the underlying writer. Async version of [Writer::write_event].
```
Mingun, line 8:
```suggestion
    /// Writes the given event to the underlying writer. Async version of [`Writer::write_event`].
```

src/writer/async_tokio.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::events::*;
+    use pretty_assertions::assert_eq;
+
+    #[tokio::test]
+    async fn xml_header() {
```
Mingun, line 54: Could you add tests for other events as well? Then all paths will be covered

Mingun (changes requested):
Documentation for `Writer` on line 12 became outdated -- now we can write to `AsyncWrite` as well. Could you update it accordingly?

It seems that implementing writing with indent should be not so hard -- it looks that you just need to copy existing code and replace all sync write methods with async counterparts. Did you try that? If there are some difficulties, then it can be postponed to other PR.

vilunov:
> It seems that implementing writing with indent should be not so hard

It seems easy indeed, but here I'm just submitting changes which I've already made and tested in my own project, and since I don't really use indentation support there, I can't claim that what I'd make would work correctly. I'd rather implement it in a separate PR.

Mingun (approved):

dralley:
Just out of personal curiosity, do you have a particular use case in mind where this is needed?

vilunov:
I'm implementing an XML-based duplex stream API. I'm going this the fully async way, that means tokio, its TcpSockets, most of the bufferized but sometimes direct. That means I need to write my data structures into that kind of byte streams. For some time I've been using the sync API and writing it into an intermediate buffer, but that means that most of the time I have two buffers: one in `BufWriter` and one my own `Vec<u8>`. So I worked around that by implementing async APIs and it worked well since then.

A typical (desugared) serialization looks like this:
```rust
pub struct Message {
    body: String,
}
impl Message {
    pub async fn write_xml(&self, writer: &mut Writer<impl AsyncWrite + Unpin>) -> Result<()> {
        writer.write_event_async(Event::Start(BytesStart::new(r#"message"#))).await?;
        writer.write_event_async(Event::Text(BytesText::new(&self.body))).await?;
        writer.write_event_async(Event::End(BytesEnd::new("message"))).await?;
        Ok(())
    }
}
```

PR #574, opened 2023-03-06
Implement correct handling of `$text` fields in enums and add roundtrip tests

Mingun:
This is follow-up PR for #541 in which I forgot to make a corresponding serialization for the deserialization. In order to not make this mistake again, the last 6 commits enables serialization-deserialization roundtrip in serialization tests -- we try to deserialize what was serialized and check that we get the same object.

During that I found that some roundrips does not work, mostly because of serde-rs/serde#1183. That tests does not check the roundtrip.

Some structs failed to derive `Deserialize` due to serde-rs/serde#2371. Although it is possible to implement it manually or use [my fork](https://github.com/Mingun/serde/tree/serde-next) where that error is fixed specially for that, the same tests suffers from the serde-rs/serde#1183 too, so I ended up with the current approach in order to reduce dependencies.

### DIFF
--- a/tests/serde-se.rs
+++ b/tests/serde-se.rs
@@ -477,13 +477,17 @@ mod without_root {
                     </InternallyTagged>");
             // serde serializes internally tagged newtype structs by delegating
             // serialization to the inner type and augmenting it with a tag
-            serialize_as!(newtype:
+            // NOTE: Cannot be deserialized in roundtrip due to
+            // https://github.com/serde-rs/serde/issues/1183
+            serialize_as_only!(newtype:
                 InternallyTagged::Newtype(Nested { float: 4.2 })
                 => "<Nested>\
                         <tag>Newtype</tag>\
                         <float>4.2</float>\
                     </Nested>");
-            serialize_as!(struct_:
+            // NOTE: Cannot be deserialized in roundtrip due to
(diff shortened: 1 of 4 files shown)


dralley:
Looking at the `serde` PRs, it looks a bit stagnant unfortunately.  Lots of them open with no progress

Mingun:
Yeah, it is. Quite surprising for that widely used library.

tests/serde-se.rs
```diff
@@ -75,7 +70,23 @@ enum ExternallyTagged {
     Empty {},
 }
 
+/// Having both `#[serde(flatten)]` and `'static` fields in one struct leads to
+/// incorrect code generation when deriving `Deserialize`.
```
dralley, line 75: Can we have a consistent way we can mark code which depends on fixing external bugs, such that they can be easily searched?  Maybe `TODO(external)` or something.

dralley (approved):

tests/serde-se.rs
Mingun: Yes, I forgot to mark those places with a TODO, will do!

I used an usual `TODO` and `NOTE` markers -- there are not many of them and the todo-tree VSCode extension shows them very well.

PR #579, opened 2023-03-17
Makes ElementWriter.write_inner_content take FnOnce instead of Fn

Tpt:
This closure is only evaluated once


Mingun (approved):
Please add a changelog entry

dralley (approved):

dralley (approved):

Issue #581, opened 2023-03-29
Optional unescaping with serde


pigeonhands:
Currently, trying to use `quick_xml::de::from_str` with xml that uses `entity` tags in the doctype fail to parse.

```XML
<!DOCTYPE dict[
    <!ENTITY unc "unclassified">
]>

<dict>
    <word>&unc;</word>
</dict>
```
```rust
let data = fs::read_to_string("../data/dict.xml").unwrap();
let parsed : Dict = quick_xml::de::from_str(&data).unwrap();
```
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidXml(EscapeError(UnrecognizedSymbol(1..4, "unc")))'

------

Could a feature flag or an alternate method be added that disables unescaping and instead always just returns the raw string?

pigeonhands:
@Mingun I'm not really sure how to go about this. I have created a fork that implements a resolver but  its not going to be able to access the entities defined in the document because that parsing is handled outside the `de` module.

https://github.com/pigeonhands/quick-xml/commit/ddacb310a2d50472e50acbf0b0d3113e28a4b60e

Mingun:
Wow! Your implementation looks promising! I think, it would be better if you create a draft PR so we can discuss on actual implementation in context.

You need to capture `DocType` events in this match:
https://github.com/tafia/quick-xml/blob/2e9123a5435683e1b2cffb0de547f972b383bbd3/src/de/mod.rs#L2745-L2765

For that change the `EntityResolver` trait to that (feel free to fix grammar if I made mistakes):
```rust
/// Used to resolve unknown entities while parsing
///
/// # Example
/// Add an example here -- you can adapt existing custom_entities.rs example
pub trait EntityResolver {
    /// Called on contents of [`Event::DocType`] to capture declared entities.
    /// Can be called multiple times, for each parsed doctype declaration.
    fn capture(&mut self, doctype: BytesText);

    /// Called when an entity needs to be resolved.
    ///
    /// `None` is returned if a suitable value can not be found.
    /// In that case an [`Error::UnrecognizedSymbol`] will be returned.
    fn resolve<'entity>(&'entity self, entity: &str) -> Option<&'entity str>;
}
```
`resolve` method no need to be mutable -- it used only for requesting information. If implementation would need mutation in that case it should use internal mutability, using `Cell` or something.

PR #583, opened 2023-03-29
Add EntityResolver and DocType event

pigeonhands:
fixes https://github.com/tafia/quick-xml/issues/581

+ Added `PayloadEvent::DocType`
+ Added an `EntityResolver`  so that unknown entities can be resolved.
+ Changed `unsescape_` functions to use `FnMut` instead of `Fn` for `resolve_entity` parameters.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -1931,10 +1931,9 @@ pub enum PayloadEvent<'a> {
     Start(BytesStart<'a>),
     /// End tag `</tag>`.
     End(BytesEnd<'a>),
-    /// Escaped character data between `Start` and `End` element.
+    /// Escaped character data between tags.
     Text(BytesText<'a>),
-    /// Unescaped character data between `Start` and `End` element,
-    /// stored in `<![CDATA[...]]>`.
+    /// Unescaped character data stored in `<![CDATA[...]]>`.
     CData(BytesCData<'a>),
     /// End of XML document.
     Eof,
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -1833,10 +1833,13 @@ macro_rules! deserialize_option {
 
 mod key;
 mod map;
+mod resolver;
 mod simple_type;
 mod var;
 
 pub use crate::errors::serialize::DeError;
+pub use resolver::{EntityResolver, NoEntityResolver};
+
 use crate::{
     encoding::Decoder,
     errors::Error,
@@ -1935,6 +1938,8 @@ pub enum PayloadEvent<'a> {
     Text(BytesText<'a>),
     /// Unescaped character data stored in `<![CDATA[...]]>`.
     CData(BytesCData<'a>),
+    /// Document type definition data (DTD) stored in `<!DOCTYPE ...>`.
+    DocType(BytesText<'a>),
     /// End of XML document.
     Eof,
 }
@@ -1948,6 +1953,7 @@ impl<'a> PayloadEvent<'a> {
             PayloadEvent::End(e) => PayloadEvent::End(e.into_owned()),
             PayloadEvent::Text(e) => PayloadEvent::Text(e.into_owned()),
             PayloadEvent::CData(e) => PayloadEvent::CData(e.into_owned()),
+            PayloadEvent::DocType(e) => PayloadEvent::DocType(e.into_owned()),
             PayloadEvent::Eof => PayloadEvent::Eof,
         }
     }
@@ -1956,7 +1962,7 @@ impl<'a> PayloadEvent<'a> {
 /// An intermediate reader that consumes [`PayloadEvent`]s and produces final [`DeEvent`]s.
 /// [`PayloadEvent::Text`] events, that followed by any event except
 /// [`PayloadEvent::Text`] or [`PayloadEvent::CData`], are trimmed from the end.
-struct XmlReader<'i, R: XmlRead<'i>> {
+struct XmlReader<'i, R: XmlRead<'i>, E: EntityResolver = NoEntityResolver> {
     /// A source of low-level XML events
     reader: R,
     /// Intermediate event, that could be returned by the next call to `next()`.
@@ -1964,15 +1970,32 @@ struct XmlReader<'i, R: XmlRead<'i>> {
     /// trailing spaces is not. Before the event will be returned, trimming of
     /// the spaces could be necessary
     lookahead: Result<PayloadEvent<'i>, DeError>,
+
+    /// Used to resolve unknown entities that would otherwise cause the parser
+    /// to return an [`EscapeError::UnrecognizedSymbol`] error.
+    ///
+    /// [`EscapeError::UnrecognizedSymbol`]: crate::escape::EscapeError::UnrecognizedSymbol
+    entity_resolver: E,
 }
 
-impl<'i, R: XmlRead<'i>> XmlReader<'i, R> {
-    fn new(mut reader: R) -> Self {
+impl<'i, R: XmlRead<'i>, E: EntityResolver> XmlReader<'i, R, E> {
+    fn new(reader: R) -> Self
+    where
+        E: Default,
+    {
+        Self::with_resolver(reader, E::default())
+    }
+
+    fn with_resolver(mut reader: R, entity_resolver: E) -> Self {
         // Lookahead by one event immediately, so we do not need to check in the
         // loop if we need lookahead or not
         let lookahead = reader.next();
 
-        Self { reader, lookahead }
+        Self {
+            reader,
+            lookahead,
+            entity_resolver,
+        }
     }
 
     /// Read next event and put it in lookahead, return the current lookahead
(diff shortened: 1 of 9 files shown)


dralley:
Please add an entry to the changelog

pigeonhands:
@dralley Done

src/de/mod.rs
```diff
@@ -2167,12 +2184,12 @@ where
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
 /// A structure that deserializes XML into Rust values.
-pub struct Deserializer<'de, R>
+pub struct Deserializer<'de, R, S: EntityResolver = DefaultEntityResolver>
```
Mingun, line 2187: Let's use consistent naming
```suggestion
pub struct Deserializer<'de, R, E: EntityResolver = DefaultEntityResolver>
```

src/escapei.rs
```diff
@@ -159,11 +159,11 @@ pub fn unescape(raw: &str) -> Result<Cow<str>, EscapeError> {
 /// [HTML5 escapes]: https://dev.w3.org/html5/html-author/charref
 pub fn unescape_with<'input, 'entity, F>(
     raw: &'input str,
-    resolve_entity: F,
+    mut resolve_entity: F,
 ) -> Result<Cow<'input, str>, EscapeError>
 where
     // the lifetime of the output comes from a capture or is `'static`
-    F: Fn(&str) -> Option<&'entity str>,
+    F: FnMut(&str) -> Option<&'entity str>,
```
Mingun, line 166: After consulting with the [standard library](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map) I think, it's a good idea, but that change should be done in it's own commit and have it's own changelog entry.

src/resolver.rs
```diff
@@ -0,0 +1,29 @@
...
+/// An EntityResolver that always returns None.
+#[derive(Default, Copy, Clone)]
+pub struct DefaultEntityResolver;
+
+impl<'entity_out> EntityResolver for DefaultEntityResolver{
+    fn resolve<'entity>(&'entity self, _: &str) -> Option<&'entity str> {
+        None
+    }
+
+    fn capture(&mut self, _: BytesText){
```
Mingun, line 27: Let's be consistent and put this method before the `resolve` method

src/resolver.rs
```diff
@@ -0,0 +1,29 @@
...
+    /// `None` is returned if a suitable value can not be found.
+    /// In that case an [`Error::UnrecognizedSymbol`] will be returned.
+    fn resolve<'entity>(&'entity self, entity: &str) -> Option<&'entity str>;
+}
+
+/// An EntityResolver that always returns None.
+#[derive(Default, Copy, Clone)]
+pub struct DefaultEntityResolver;
+
+impl<'entity_out> EntityResolver for DefaultEntityResolver{
```
Mingun, line 22:
```suggestion
impl EntityResolver for DefaultEntityResolver {
```

src/resolver.rs
```diff
@@ -0,0 +1,29 @@
...
+    /// Called when an entity needs to be resolved.
+    ///
+    /// `None` is returned if a suitable value can not be found.
+    /// In that case an [`Error::UnrecognizedSymbol`] will be returned.
+    fn resolve<'entity>(&'entity self, entity: &str) -> Option<&'entity str>;
+}
+
+/// An EntityResolver that always returns None.
+#[derive(Default, Copy, Clone)]
+pub struct DefaultEntityResolver;
```
Mingun, line 20:
```suggestion
pub struct NoEntityResolver;
```

src/resolver.rs
```diff
@@ -0,0 +1,29 @@
+//! Entity resolver module
+
+use crate::events::BytesText;
+
+/// Used to resolve unknown entities while parsing
+pub trait EntityResolver {
```
Mingun, line 6: Please add an example here as suggested in [my comment](https://github.com/tafia/quick-xml/issues/581#issuecomment-1488823175)

src/de/mod.rs
```diff
@@ -1936,6 +1937,9 @@ pub enum PayloadEvent<'a> {
     /// Unescaped character data between `Start` and `End` element,
     /// stored in `<![CDATA[...]]>`.
     CData(BytesCData<'a>),
+    /// Unescaped character data between `Start` and `End` element,
+    /// stored in `<!DOCTYPE ... [...]]>`.
```
Mingun, line 1941:
```suggestion
    /// Document type definition data (DTD) stored in `<!DOCTYPE ...]>`.
```
Actually, I would like if you would synchronize documentation between `Event` and `PayloadEvent` (in it's own commit, do that first)

src/de/mod.rs
```diff
@@ -1949,31 +1953,40 @@ impl<'a> PayloadEvent<'a> {
...
+struct XmlReader<'i, R: XmlRead<'i>, E: EntityResolver = DefaultEntityResolver> {
     /// A source of low-level XML events
     reader: R,
     /// Intermediate event, that could be returned by the next call to `next()`.
     /// If that is the `Text` event then leading spaces already trimmed, but
     /// trailing spaces is not. Before the event will be returned, trimming of
     /// the spaces could be necessary
     lookahead: Result<PayloadEvent<'i>, DeError>,
+
+    entity_resolver: E
```
Mingun, line 1975: Please add a doc comment explaining for what this field is used

Mingun (changes requested):
Thanks for your work! I think PR will benefit if you make such changes: split it into three commits (rebase your branch and force push it):
- the first one that synchronizes documentation between variants of `Event` and `PayloadEvent`
- the second one which relaxes requirements to `resolve_entity` parameters from `Fn` to `FnMut`. That commit should relax requirements in all places and note this in the changelog
- the third one which adds an `EntityResolver` trait etc. I was too quick in my comment regarding the `resolve` method. Because `resolve_entity` are now would be `FnMut`, the trait could be defined as
  ```rust
  pub trait EntityResolver {
      fn capture(&mut self, doctype: BytesText);
      fn resolve(&mut self, entity: &str) -> Option<&str>;
  }
  ```
  Add tests for the new feature in this commit

src/escapei.rs
dralley: It's not immediately clear why it's necessary, as the docs [say](https://doc.rust-lang.org/std/ops/trait.FnMut.html):

> Use FnMut as a bound when you want to accept a parameter of function-like type and need to call it repeatedly, while allowing it to mutate state. If you don’t want the parameter to mutate state, use [Fn](https://doc.rust-lang.org/std/ops/trait.Fn.html) as a bound; if you don’t need to call it repeatedly, use [FnOnce](https://doc.rust-lang.org/std/ops/trait.FnOnce.html).

What state do you expect the entity resolution process to mutate?  Isn't it a read-only process?

Mingun: Yes, it is just to provide flexibility and be consistent with similar std API, such as `Iterator::map`.

pigeonhands:
> This is almost good, just need to polish some things.
> 
> Actually, the first commit did not what I want. I meant that you to change the doc strings of all matching `Event` and `PayloadEvent` variants to be the same by keeping the best description. And then in the third commit add a new variant to `PayloadEvent` (so it would be just a copy of the corresponding variant from `Event`)

I'm not sure what you mean by this. Should all the enum variants be the same between `PayloadEvent` and `Event` or just the docs?

Mingun:
Just the docs, because `PayloadEvent` is a subset of possible `Event`s, and docs should be the same on same variants.

Mingun (approved):
I've updated the PR, with the following changes:
- Allow `EntityResolver::capture` method to fail
- Change the example on `EntityResolver` more close to possible real world implementation
- Make `resolver` module private and instead re-export trait and type in `de` module

Unfortunately, making `EntityResolver::resolve` taking `&mut self` leads to strange error, that the returned reference could not outlive the captured object in the closure. This is strange, because returning the same reference while resolver is read-only is fine. So if you wish to make lazy parsing in `resolve`, you should use `Cell` or similar.

@dralley, could you look at this once again? It seems good for me now and I plan to make another patch release after merging that.

tests/serde-de.rs
```diff
@@ -6427,3 +6427,71 @@ mod borrow {
...
+
+    impl EntityResolver for TestEntityResolver {
+        fn capture(&mut self, doctype: quick_xml::events::BytesText) {
+            self.capture_called = true;
+            let as_cow = doctype.into_inner();
+            let str_doc = String::from_utf8_lossy(as_cow.as_ref());
+            assert_eq!(str_doc.as_ref(), r#"dict[ <!ENTITY unc "unclassified"> ]"#);
+        }
+
+        fn resolve<'entity>(&'entity self, entity: &str) -> Option<&'entity str> {
```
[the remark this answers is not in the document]
Mingun: Change to `&mut` is not liked by rustc, so keep it as is

dralley (approved):

PR #601, opened 2023-05-11
Add `serde_helper` module and document a frequent pattern of enums usage

### LINKED ISSUE #600: Enum serialization/deserialization problem + #[serde(flatten)] inconsistency
Hi, first of all thank you for this crate!

I am trying to communicate with an API that uses xml for the data exchange, so I have to both serialize and deserialize data from string and I would like to achive this using serde.
Sadly I have the following problem with enums:

Take this enum for example:
```rust
enum Type {
    Local,
    Online
}
```
and the usage in a struct like:
```rust
struct Software {
    type: Type
}
```
I would like this to serialize into:
```xml
<software>
    <type>Local</type>
</software>
```

but sadly serializing this enum (even when non of the variants have inner data) produces the following:
```xml
<software>
    <Local/>
</software>
```

after some trial and error I could come up with the following 'solution' that works for serialization, but is not too elegant/convenient:
```rust
#[serde(tag = "type")]
enum Type {
    Local,
    Online
}

struct Software {
    #[serde(flatten)]
    type: Type
}
```
Unfortunately this only works for serialization, but not for deserialization :/ If I try to deserialize the xml from above it produces the error: `Error: invalid type: map, expected variant identifier`
For this error I was not able to find a solution or combination of serde attributes, that would work.

Is there a way currently to deserialize an Enum with no inner data?
Would it be possible for quick-xml to change the default serialization of enums so that if they hold no inner data, they get serialized into a tag like in the xml above? 
I saw that previously, there was a `#[serde(rename = "$primitive")]` attributum for this, but it got removed later (which I understand)

Mingun:
This PR documents pattern that used in #600 and introduces a module with a helper functions for XML. This module could be grow in the future. @horvbalint if you wish, you can review this. I'm interesting of getting feedback about comprehensibility and my language.

Closes #600

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -27,6 +27,7 @@
 //!     - [Enums and sequences of enums](#enums-and-sequences-of-enums)
 //! - [Frequently Used Patterns](#frequently-used-patterns)
 //!   - [`<element>` lists](#element-lists)
+//!   - [Enum::Unit Variants As a Text](#enumunit-variants-as-a-text)
 //!
 //!
 //!
@@ -1579,7 +1580,8 @@
 //! Some XML constructs used so frequent, that it is worth to document the recommended
 //! way to represent them in the Rust. The sections below describes them.
 //!
-//! ## `<element>` lists
+//! `<element>` lists
+//! -----------------
 //! Many XML formats wrap lists of elements in the additional container,
 //! although this is not required by the XML rules:
 //!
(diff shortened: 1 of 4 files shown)


src/de/mod.rs
```diff
@@ -1672,9 +1674,79 @@
 //!
 //! Instead of writing such functions manually, you also could try <https://lib.rs/crates/serde-query>.
 //!
+//! Enum::Unit Variants As a Text
+//! -----------------------------
+//! A one frequent task and a typical mistake is to creation of mapping a text
```
dralley, line 1679: "A" or "One" but not both

dralley (approved):

PR #603, opened 2023-05-24
Trim next element after DocType

danjpgriffin:
Allow comment to exist between DocType and root element

With the new EntityResolver/Doctype changes, putting an xml comment between a doctype and the root element causes the parse to fail. Trimming the next element fixes this.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -2880,7 +2880,7 @@ impl StartTrimmer {
     #[inline(always)]
     fn trim<'a>(&mut self, event: Event<'a>) -> Option<PayloadEvent<'a>> {
         let (event, trim_next_event) = match event {
-            Event::DocType(e) => (PayloadEvent::DocType(e), false),
+            Event::DocType(e) => (PayloadEvent::DocType(e), true),
             Event::Start(e) => (PayloadEvent::Start(e), true),
             Event::End(e) => (PayloadEvent::End(e), true),
             Event::Eof => (PayloadEvent::Eof, true),
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,9 +18,15 @@
 
 ### Bug Fixes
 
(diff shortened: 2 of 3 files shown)


src/de/mod.rs
```diff
@@ -2880,7 +2880,7 @@ impl StartTrimmer {
     #[inline(always)]
     fn trim<'a>(&mut self, event: Event<'a>) -> Option<PayloadEvent<'a>> {
         let (event, trim_next_event) = match event {
-            Event::DocType(e) => (PayloadEvent::DocType(e), false),
+            Event::DocType(e) => (PayloadEvent::DocType(e), true),
```
Mingun, line 2883: Could you add a comment describing, why we need trim here because this is not obvious (otherwise it would be done initially)

Mingun (changes requested):
Could you add a comment and update changelog. I would like also to have some more formal tests for that situation (can be added in `de/mod.rs`). Preferably to test each possible situation that you can imagine. I have a feeling, that the same problem can be happened not only for `DocType + Comment` events.

src/de/mod.rs
danjpgriffin: The Doctype should trim events here just like the Start/End events. Only the CData/Text events should not trim (otherwise any CData/Text following them will be ignored

Mingun (approved):
Thanks for your work! After investigating situation I realized, that actually comment is not needed, because trimming `Text` events after `DocType` event just follow common rule. Returning `false` was an oversight in #583, which introduced this regression.

After investigating I realized that we should add tests for deserialization that uses `deserialize_struct` or `deserialize_map` methods (`HashMap` uses the last) and an XML contains comments or processing instructions in [XML prolog](https://www.w3.org/TR/xml11/#NT-prolog)

PR #612, opened 2023-06-11
Return an error on empty doctype

### LINKED ISSUE #608: Debug assertion triggered on empty doctype
Thank you for quick-xml and sorry for comming back with an other bug.

When parsing the file `<!DoctyPE>` using rusc "dev" profile the assertion `"DocType must have a name"` is triggered. In the "release" profile the assertion is not triggered and the parser returns an empty `Event::Doctype()`

Test failing in "dev" profile and passing in "release" profile:
```rust
#[test]
fn reproduce() {
    let data = b"<!DoctyPE>";
    let mut reader = Reader::from_reader(data.as_slice());
    let mut buf = Vec::new();
    loop {
            Ok(Event::Eof) => break,
            Ok(Event::DocType(e)) => assert!(e.is_empty()), // assertion only executed in "release" mode
            _ => unreachable!()
        }
    }
}
```

Tpt:
The parser was crashing with debug assertions enabled but passing with debug assertions disabled

Closes #608

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -37,6 +37,10 @@ pub enum Error {
     /// `Event::BytesDecl` must start with *version* attribute. Contains the attribute
     /// that was found or `None` if an xml declaration doesn't contain attributes.
     XmlDeclWithoutVersion(Option<String>),
+    /// Empty `Event::DocType`. `<!doctype foo>` is correct but `<!doctype > is not.
+    ///
+    /// See <https://www.w3.org/TR/xml11/#NT-doctypedecl>
+    EmptyDocType,
     /// Attribute parsing error
     InvalidAttr(AttrError),
     /// Escape error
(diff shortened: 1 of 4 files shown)


src/errors.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
     /// `Event::BytesDecl` must start with *version* attribute. Contains the attribute
     /// that was found or `None` if an xml declaration doesn't contain attributes.
     XmlDeclWithoutVersion(Option<String>),
+    /// Empty `Event::DocType`.
+    EmptyDocType,
```
Mingun, line 41: I suggest to expand description a bit by showing an example of incorrect and correct declarations:
- `<!doctype>` is incorrect
- `<!doctype >` is correct
- `<!doctype sometype>` is correct

Mingun (changes requested):
Need to run `cargo fmt` and please add a changelog entry.

src/errors.rs
Tpt: Thank you! Done.

PR #615, opened 2023-06-21
Add ability to set entity resolver when deserialize using borrowing reader

Mingun:
I noticed, that after #583 there is no way to create a deserializer that will borrow from string and provide ability to resolve custom entities. This PR fixes this gap.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@
 
 - [#609]: Added `Writer::write_serializable` to provide the capability to serialize
   arbitrary types using serde when using the lower-level `Writer` API.
+- [#615]: Added ability to set entity resolver when deserialize using borrowing reader.
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


Changelog.md
```diff
@@ -14,13 +14,15 @@
...
 - [#609]: Added `Writer::write_serializable` to provide the capability to serialize
   arbitrary types using serde when using the lower-level `Writer` API.
+- [#615]: Added ability to set entity resolver when deserialize using borrowing reader.
 
 ### Bug Fixes
 
 ### Misc Changes
 
 
 [#609]: https://github.com/tafia/quick-xml/issues/609
```
dralley, line 24: This should be pull #609, not issue #609

My mistake, looks like I didn't fix it up completely.

dralley (approved):
See changelog comment

Changelog.md
Mingun: Ok, will fix. Just a moment...

Issue #258, opened 2021-02-05
Declared entity is not recognized


pchampin:
When a file defines new entities into its DOCTYPE, quick-xml does not recognize these entities in attribute values or text nodes. This causes an `EscapeError(UnrecognizedSymbol(...))` error.

I built a minimal example demonstrating this problem...
```rust
use quick_xml::Reader;
use quick_xml::events::Event;

const DATA: &str = r#"

    <?xml version="1.0"?>
    <!DOCTYPE test [
    <!ENTITY msg "hello world" >
    ]>
    <test label="&msg;">&msg;</test>

"#;

fn main() {
    
    let mut reader = Reader::from_str(DATA);
    reader.trim_text(true);
    
    let mut buf = Vec::new();
    
    loop {
        match reader.read_event(&mut buf) {
            Ok(Event::Start(ref e)) => {
                match e.name() {
                    b"test" => println!("attributes values: {:?}",
                                        e.attributes().map(|a| a.unwrap().unescape_and_decode_value(&reader).unwrap())
                                        .collect::<Vec<_>>()),
                    _ => (),
                }
            },
            Ok(Event::Text(ref e)) => {
                println!("text value: {}", e.unescape_and_decode(&reader).unwrap());
            },
            Ok(Event::Eof) => break,
            Err(e) => panic!("Error at position {}: {:?}", reader.buffer_position(), e),
            _ => (),
        }
    }
}
```

I wasn't sure if that was a dublicate of #124 because that issue is about *external* entities...

pchampin:
If I understand correctly, quick-xml does not intend to parse the content of the DOCTYPE, which is fair enough.

I would be happy with a (hopefully simpler) feature to add new entities to the parser. That way, I could:
* subscribe to the DOCTYPE event
* make my custom parsing if its content to extract only local entities (that's really just what I need)
* inform the `Reader` about those entities so that it does not choke on their use in the rest of the file.

tafia:
Indeed quick-xml doesn't support parsing DOCTYPE for the moment. I'd be happy to review a PR but I believe this is not a trivial change.

One way you could work around it is to:
- get the `DocType` text event and parse the entities into a hashmap
- when getting the attributes, don't try to unescape values, instead search in your map first then only unescape (using `Attribute::value` field and the unsecape function))
Alternatively if you feel like you could improve quick-xml I'd accept any change, eventually behing a feature flag if needed.

pchampin:
[...]

PR #618, opened 2023-06-25
Fails on empty comments without 4 -"

### LINKED ISSUE #604: Panic when parsing invalid comments `<!-->`
On the following input quick XML 0.28-2 panics with ` 'slice index starts at 3 but ends at 1', quick-xml-0.28.2/src/reader/parser.rs:104:22`

```xml
<?xml version="1.0" encoding="utf-8"?>
<sparql xmlns="http://www.w3.org/2005/sparql-resu:lts#">
<head>
        <variable name="str"/>
</head>
<results>
        <result></result> <!-->
        <result></resulhtmlns>=
```

I believe this error is caused by [this line](https://github.com/tafia/quick-xml/blob/84b07b4e5c9a201b3d3d7fa95c1e8c33ecbe6dd2/src/reader/mod.rs#L828) where the `buf.len() + i > 4` check is wrong.

Tpt:
The parser was crashing because of bad slice bounds

Closes #604

### DIFF
--- a/src/reader/buffered_reader.rs
+++ b/src/reader/buffered_reader.rs
@@ -117,7 +117,9 @@ macro_rules! impl_buffered_source {
                     // somewhere sane rather than at the EOF
                     Ok(n) if n.is_empty() => return Err(bang_type.to_err()),
                     Ok(available) => {
-                        if let Some((consumed, used)) = bang_type.parse(buf, available) {
+                        // We only parse from start because we don't want to consider
+                        // whatever is in the buffer before the bang element
+                        if let Some((consumed, used)) = bang_type.parse(&buf[start..], available) {
                             buf.extend_from_slice(consumed);
 
                             self $(.$reader)? .consume(used);
--- a/tests/issues.rs
+++ b/tests/issues.rs
@@ -4,7 +4,7 @@
 
 use std::sync::mpsc;
 
-use quick_xml::events::{BytesStart, Event};
+use quick_xml::events::{BytesDecl, BytesStart, BytesText, Event};
 use quick_xml::name::QName;
 use quick_xml::reader::Reader;
 use quick_xml::Error;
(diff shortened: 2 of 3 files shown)


tests/fuzzing.rs
```diff
@@ -60,3 +60,29 @@ fn fuzz_empty_doctype() {
     ));
     assert_eq!(reader.read_event_into(&mut buf).unwrap(), Event::Eof);
 }
+
+#[test]
+fn fuzz_empty_comment() {
```
Mingun, line 65: I think, it would be better to place tests to `tests/issues.rs` file. Also, could you check that we actually returns `UnexpectedToken("<!-->")`? That is, explicitly expect all the events that the parser returns:
```rust
assert_eq!(reader.read_event_into(&mut buf), Ok(...));
assert_eq!(reader.read_event_into(&mut buf), Ok(...));
// and so on
```

src/reader/parser.rs
```diff
@@ -100,6 +100,10 @@ impl Parser {
                         return Err(Error::UnexpectedToken("--".to_string()));
                     }
                 }
+                if len < 5 {
+                    // We do not have any text content
+                    return Err(Error::UnexpectedToken("<!-->".to_string()));
```
Mingun, line 105: I think, it would be better to write something more descriptive, because this path will be taken also for the `<!--->` input and having `<!-->` in the error could be misleading. Maybe even introduce a new error variant specially for that case (`NotClosedComment`, for example)

Mingun (changes requested):
I think that approach is good, but we need more bullet-proof tests that will actually test that we take this code path (by checking actual events).

Mingun:
Actually, fix of this issue not so simple. `<!-->-->` is a valid XML (see https://github.com/tafia/quick-xml/issues/258#issuecomment-1586322843)

tests/fuzzing.rs
Tpt: Indeed. Done.

src/reader/parser.rs
Tpt: I have changed approach. I am not creating new error with the new approach. The new error message is not great but I believe it is for an other PR and not this one. Would you be ok with that?

Tpt:
> Actually, fix of this issue not so simple.  is a valid XML (see https://github.com/tafia/quick-xml/issues/258#issuecomment-1586322843)

Indeed. Thank you for spotting this! I have found a cleaner way to fix the bug that behaves properly in this case.

tests/issues.rs
```diff
@@ -105,3 +105,51 @@ mod issue514 {
         assert_eq!(reader.read_event().unwrap(), Event::Eof);
     }
 }
+
+#[test]
+fn issue_604_short() {
```
Mingun, line 110: Please put all tests in a module, like as above. Prefix in names not needed because they will be in a module

tests/issues.rs
```diff
@@ -105,3 +105,51 @@ mod issue514 {
...
+    let data = b"<?xml version=\"1.0\" encoding=\"utf-8\"?><!-->";
+    let mut reader = Reader::from_reader(data.as_slice());
+    let mut buf = Vec::new();
+    assert!(matches!(
+        reader.read_event_into(&mut buf),
+        Ok(Event::Decl(_))
+    ));
```
Mingun, line 117: Use `assert_eq!()` + `unwrap()` for `Ok` results, like as above. `assert_eq!()` is from `pretty_assertions` crate which will print a nice diff in case of error

src/reader/parser.rs
Mingun: Yes, that is OK. We need to rework our errors anyway

tests/issues.rs
```diff
@@ -105,3 +105,51 @@ mod issue514 {
...
+    assert!(matches!(
+        reader.read_event_into(&mut buf),
+        Err(Error::UnexpectedEof(_))
+    ));
+    assert!(matches!(reader.read_event_into(&mut buf), Ok(Event::Eof)));
+}
+
+#[test]
+fn issue_604_valid() {
+    let data = b"<?xml version=\"1.0\" encoding=\"utf-8\"?><!-->-->";
```
Mingun, line 143: Could you also add a test for `<!--->-->`?

Mingun (changes requested):
Almost good, thanks! I left several comments

Tpt:
Thank you! I have updated the tests following your review!

Mingun (approved):
Approved, but I'd like to postpone merging until I'll get idea how it is working

src/reader/buffered_reader.rs
```diff
@@ -117,7 +117,7 @@ macro_rules! impl_buffered_source {
                     // somewhere sane rather than at the EOF
                     Ok(n) if n.is_empty() => return Err(bang_type.to_err()),
                     Ok(available) => {
-                        if let Some((consumed, used)) = bang_type.parse(buf, available) {
+                        if let Some((consumed, used)) = bang_type.parse(&buf[start..], available) {
```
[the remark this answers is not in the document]
Mingun: I think, I understood. We should parse only newly read data from the `buf`. This is correct, because we in the function that parses the whole comment, so any data that exists in the buffer before is not relevant here and should be skipped.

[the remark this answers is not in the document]
Tpt: Yes, exactly. I have added the comment. Not doing so was making some checks like [the one on this line](https://github.com/tafia/quick-xml/blob/60249aeea24bfe00848a70950ae7cd321cacfb3c/src/reader/mod.rs#L828) to be incorrect.

Mingun:
Thanks! I slightly changed tests to more precisely test what an error we returns and changed Changelog description to be less technically specific and more user specific

PR #620, opened 2023-06-27
Add expand_empty_elements fn to Serializer

### LINKED ISSUE #617: quick-xml v0.27+ changed behaviour when writing empty values
I'm using quick-xml for my XML-RPC crate ([dxr](https://github.com/ironthree/dxr)) and I've been a happy user so far, thank you for maintaining the crate!

I'm trying to update my quick-xml dependency to a newer version (dxr is currently stuck at v0.25), but I'm hitting a roadblock. It looks like the new serde serializer that was implemented for v0.27 changed some behaviour when writing empty values.

For example, it now writes self-closing XML tags - previously, I got `<value><string></string></value>` for an empty string value, but with quick-xml v0.27+, I get `<value><string/></value>`. While this is technically valid XML, my library cannot write self-closing tags like this in order to maintain compatibility with other XML-RPC implementations (some older ones don't accept self-closing XML tags).

Is there a way to get back the old behaviour (i.e. deserialize an empty `$value` to `<foo></foo>` instead of `<foo/>`?

bahlo:
Setting this to true will serialize empty elements to `<element></element>` instead of `<element/>`.

Fixes #617

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,7 @@
 - [#609]: Added `Writer::write_serializable` to provide the capability to serialize
   arbitrary types using serde when using the lower-level `Writer` API.
 - [#615]: Added ability to set entity resolver when deserialize using borrowing reader.
+- [#620]: Added ability to enforce the expansion of empty elements.
 
 ### Bug Fixes
 
@@ -23,6 +24,7 @@
 
 [#609]: https://github.com/tafia/quick-xml/pull/609
 [#615]: https://github.com/tafia/quick-xml/pull/615
+[#620]: https://github.com/tafia/quick-xml/pull/620
 
 
 ## 0.29.0 -- 2023-06-13
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,7 +15,7 @@
 - [#609]: Added `Writer::write_serializable` to provide the capability to serialize
   arbitrary types using serde when using the lower-level `Writer` API.
 - [#615]: Added ability to set entity resolver when deserialize using borrowing reader.
-- [#620]: Added ability to enforce the expansion of empty elements.
+- [#617]: Added ability to enforce the expansion of empty elements.
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


src/se/mod.rs
```diff
@@ -522,11 +523,18 @@ impl<'w, 'r, W: Write> Serializer<'w, 'r, W> {
...
             },
             root_tag: root_tag.map(|tag| XmlName::try_from(tag)).transpose()?,
         })
     }
 
+    /// Enable or disable expansion of empty elements.
+    pub fn expand_empty_elements(&mut self, expand: bool) -> &mut Self {
+        self.ser.expand_empty_elements = expand;
+        self
+    }
```
Mingun, line 536: Could you add a doctest example and specify that default value is `false`.

Mingun (changes requested):
Please add changelog entry (do not forgot to add a link target, it in the end of corresponding release section).

Other than that and a minor note about documentation looks good.

src/se/mod.rs
bahlo: @Mingun I added a dockest and noticed changing the expand_empty_elements in the method does not affect the `ContentSerializer` that actually does the serializing. 

I didn't find the culprit yet, is this due to macros or cloning?

Changelog.md
```diff
@@ -15,6 +15,7 @@
 - [#609]: Added `Writer::write_serializable` to provide the capability to serialize
   arbitrary types using serde when using the lower-level `Writer` API.
 - [#615]: Added ability to set entity resolver when deserialize using borrowing reader.
+- [#620]: Added ability to enforce the expansion of empty elements.
```
Mingun, line 18: I prefer to use link to the issue when PR fixes/implements an issue

Mingun (approved):
Thanks! One minor note about link and everything other is good

Changelog.md
bahlo: Fixed in 5250849—did you see my other comment?

bahlo:
@Mingun Ooops, I don't think this should've been merged, tests are failling.

See my comment above:
> I added a dockest and noticed changing the expand_empty_elements in the method does not affect the ContentSerializer that actually does the serializing.
> I didn't find the culprit yet, is this due to macros or cloning?

Mingun:
Oops... I thought I saw the green OK, but forgot, that I did not wait for the tests to be completed last night. Fixed in 3ab7f646c5cc8cc729cae9824b3c0b332a5fdd31

bahlo:
Thanks for fixing it @Mingun 🙏🏻

PR #633, opened 2023-08-16
Reduce duplicate code in namespace addition.

wt:
There is code that is the same between the different types of prefix declarations. I think this is a bit easier to read since the common code is no longer duplicated.


Mingun:
I think, it is more readable in the current state because of only one condition that reader should examine. A small duplication is acceptable.

Issue #629, opened 2023-08-14
Could impl_deserialize_for_internally_tagged_enum handle a default case?


aesteve:
Hello and thanks a lot for your work.

Would it be possible (and would it make sense) for `impl_deserialize_for_internally_tagged_enum` to handle a default case?

Like "for every other value of this tag, just deserialise to this enum variant.

Not sure how common this is but I'm facing this exact pattern where I'd like to handle a few variants that I know are specific, and all the rest can just be fall into a generic struct.

I hope it's clear, sorry if it's already possible and I missed it. Happy to update the docs to highlight it in this case.

Thanks a lot.

Mingun:
It is not clear for me what exactly you need. Perhaps this is already could be implemented using existing serde API. Could you provide an example of your Rust struct and XML that you want to map?

aesteve:
Thanks for the confusion, let me try to explain this a bit better.

```
<Root>
  <SomeElement type="SomeType">
      <MoreSpecific>a</MoreSpecific>
      <Common>123</Common>
  </SomeElement>
  <SomeElement type="AnotherType">
     <Common>MaybeNotAnInt</Common>
  </SomeElement>
  <SomeElement type="YetAnotherType">
     <Common>AnotherString</Common>
  </SomeElement>
</Root>
```

I could map to the following enum:
```rust
enum Elements {
  Specific({Common: u32, MoreSpecific: String}),
  TheRest({Commong: String})
}
```

The idea is that I'd like "everything that's not `type=Specific` to be handled as `TheRest`, no matter `@type`.

I'm sorry I can't explain it better, but basically getting the equivalent of the  `_ => ...` pattern in a `match` expression.

Mingun:
Ah, I didn't immediately realize it was about macro `impl_deserialize_for_internally_tagged_enum!` from quick-xml. Probably this could be implemented. You should replace default arm (`_`) here
https://github.com/tafia/quick-xml/blob/1be35e1645dad7d807347b490b4806339984a38f/src/serde_helpers.rs#L181-L186
with `Ok(deserialize_variant!(...))` with the name of default variant. Feel free to submit a PR!

aesteve:
Thanks for the detailed answer.

Unfortunately I have never written a single macro, so it's probably gonna take a while for me to find the syntax to add an optional argument (so that it's backward compatible), and how to deal with this in the body macro.

I'll try to, at some point, but can't promise I'll manage to 😅


Thanks again.

[...]

PR #634, opened 2023-08-16
Add a default case for `impl_deserialize_for_internally_tagged_enum`

aesteve:
Attempt to implement #629 .

* I did not find any unit tests apart from the doctest, let me know if I missed something
* I tried to make the implementation backwards compatible with the existing one. If I understood correctly, it accepted a `,` at the end of the list of all variants so I replicated this behaviour. But I think in the current version, it'd accept 2 commas, not sure if what I wrote is the cleanest. 
* I'd really encourage anyone to actually check this (backwards-compatibility especially) in very details, since that's the very first macro I'm writing
* let me know if the doc is unclear or if you want me to change anything, no problem

### DIFF
--- a/src/serde_helpers.rs
+++ b/src/serde_helpers.rs
@@ -44,6 +44,54 @@ macro_rules! deserialize_variant {
     });
 }
 
+/// Helper macro that generates different match expressions depending on the presence
+/// of default variant
+#[macro_export]
+#[doc(hidden)]
+macro_rules! deserialize_match {
+    // Only default variant
+    (
+        $tag:ident, $de:ident, $enum:ty,
+        (_ => $($default_variant:tt)+ )
+        $(,)?
+    ) => (
+        Ok($crate::deserialize_variant!( $de, $enum, $($default_variant)+ ))
+    );
+
+    // With default variant
+    (
+        $tag:ident, $de:ident, $enum:ty,
+        $(
+            ($variant_tag:literal => $($variant:tt)+ )
+        ),*
+        , (_ => $($default_variant:tt)+ )
+        $(,)?
+    ) => (
+        match $tag.as_ref() {
+            $(
+                $variant_tag => Ok($crate::deserialize_variant!( $de, $enum, $($variant)+ )),
+            )*
+            _ => Ok($crate::deserialize_variant!( $de, $enum, $($default_variant)+ )),
+        }
+    );
+
+    // Without default variant
+    (
+        $tag:ident, $de:ident, $enum:ty,
+        $(
+            ($variant_tag:literal => $($variant:tt)+ )
+        ),*
+        $(,)?
+    ) => (
+        match $tag.as_ref() {
+            $(
+                $variant_tag => Ok($crate::deserialize_variant!( $de, $enum, $($variant)+ )),
+            )*
+            _ => Err(A::Error::unknown_field(&$tag, &[$($variant_tag),+])),
+        }
+    );
+}
+
 /// A helper to implement [`Deserialize`] for [internally tagged] enums which
 /// does not use [`Deserializer::deserialize_any`] that produces wrong results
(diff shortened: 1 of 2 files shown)


src/serde_helpers.rs
```diff
@@ -118,6 +118,60 @@ macro_rules! deserialize_variant {
 /// );
 /// ```
 ///
+/// You don't have to provide all the enumeration variants and can use `_` to put every undefined tag into an enumeration variant like so:
```
Mingun, line 121: Can you explicitly mention, that `(_ => Other)` should be the last variant?

src/serde_helpers.rs
```diff
@@ -118,6 +118,60 @@ macro_rules! deserialize_variant {
...
+///     #[serde(rename = "@attribute")]
+///     attribute: u64,
+/// }
+///
+/// // The macro needs the type of the enum, the tag name,
+/// // and information about all the variants
+/// impl_deserialize_for_internally_tagged_enum!{
+///     InternallyTaggedEnum, "@tag",
+///     ("NewType"    => NewType(Newtype)),
+///     (_  => Other),
```
Mingun, line 154:
```suggestion
///     ("NewType" => NewType(Newtype)),
///     (_ => Other),
```
In the first example extra spaces used to make table layout, which is redundant here

src/serde_helpers.rs
```diff
@@ -127,7 +181,9 @@ macro_rules! impl_deserialize_for_internally_tagged_enum {
         $tag:literal,
         $(
             ($variant_tag:literal => $($variant:tt)+ )
-        ),* $(,)?
+        ),*
+        $(, (_ => $($default_variant:tt)+ ))?
+        $(,)*
```
Mingun, line 186:
```suggestion
        $(, (_ => $($default_variant:tt)? ))?
        $(,)?
```
to allow only one default variant and do not allow multiple trailing commas

src/serde_helpers.rs
```diff
@@ -182,6 +238,8 @@ macro_rules! impl_deserialize_for_internally_tagged_enum {
                             $(
                                 $variant_tag => Ok(deserialize_variant!( de, $enum, $($variant)+ )),
                             )*
+                            $( _ => Ok(deserialize_variant!( de, $enum, $($default_variant)+ )), )?
+                            #[allow(unreachable)]
```
Mingun, line 242: You can use intermediate macro with two possible variants to select required behavior. General idea:
```rust
macro_rules! def {
  // variant 1
  (
    $($variant_tag:literal),*
    _ => Ok( $tokens:tt )
  ) => {
    _ => Ok($tokens),
  },
  // variant 2
  (
    $($variant_tag:literal),*
  ) => {
    _ => Err(A::Error::unknown_field(&tag, &[$($variant_tag),+])),
  },
}

// use 
def!(
  $($variant_tag),*
  $( _ => Ok(deserialize_variant!( de, $enum, $($default_variant)? )), )?
)
```
If `$(...)?` part exists, then the first alternative would be selected, otherwise the second. That way we would not need to use `#[allow(unreachable)]` attribute, which could be forbidden in the project via `#[forbid(unreachable)]`

Mingun (changes requested):
Good, but small improvements are possible. Also, please update `Changelog.md`: add new entry under **Features** (links to issues are listed at the end of version section). Link it to the issue #629, not to this PR.

src/serde_helpers.rs
aesteve: Thanks for the pointer, I was had the conviction the allow_unreachable wasn't a clean solution.

I get the general idea here, but this code doesn't compile ("no rules expected the token `,`") and I have no clue what I should change. 

Thanks.

Mingun:
Thanks for your work! I implemented the helper macro myself and push it.

PR #635, opened 2023-08-18
Add support for async element writer

tevoinea:
This PR adds support for writing to buffers implementing tokio's `AsyncWrite`. It mirrors the `ElementWriter` implementation, nearly identically. The only difference is in `write_inner_content`, the closure needs to return the Writer, I couldn't make the lifetimes line up any other way.

It also includes a full test suite, again mirrored from `ElementWriter` and docs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ path = "benches/macrobenches.rs"
 [features]
 default = []
 
-## Enables support for asynchronous reading from `tokio`'s IO-Traits by enabling
+## Enables support for asynchronous reading and writing from `tokio`'s IO-Traits by enabling
 ## [reading events] from types implementing [`tokio::io::AsyncBufRead`].
 ##
 ## [reading events]: crate::reader::Reader::read_event_into_async
--- a/src/writer/async_tokio.rs
+++ b/src/writer/async_tokio.rs
@@ -1,8 +1,10 @@
+use std::future::Future;
+
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use crate::errors::Result;
-use crate::events::Event;
-use crate::Writer;
+use crate::events::{BytesCData, BytesText, Event};
+use crate::{ElementWriter, Writer};
 
 impl<W: AsyncWrite + Unpin> Writer<W> {
     /// Writes the given event to the underlying writer. Async version of [`Writer::write_event`].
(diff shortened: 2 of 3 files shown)


src/writer/async_tokio.rs
```diff
@@ -339,4 +385,46 @@ mod indentation_async {
 </paired>"#
         );
     }
+
+    #[tokio::test]
+    async fn element_writer_nested() {
+        let mut buffer = Vec::new();
+        let mut writer = Writer::new_with_indent(&mut buffer, b' ', 4);
```
dralley, line 392: I'm fairly certain that the only reason this works is that you're using a buffer type which also implements `Write` rather than an exclusively-async type. 

`write_text_content` and other methods aren't implemented for `ElementWriter<'a, W: AsyncWrite + Unpin>`

As a sidenote if you can avoid duplicating the definition of `ElementWriter`, please do.

tevoinea: > I'm fairly certain that the only reason this works is that you're using a buffer type which also implements Write rather than an exclusively-async type.

Good catch! I'll update that

> As a sidenote if you can avoid duplicating the definition of ElementWriter, please do.

AFAIK, the only way to do that would be to have an async-only implementation, then have the sync implementation be a wrapper around it. But since async is an optional feature in this crate (and perf is important), I don't think that's the best option. 

If there's a better way, I'd love to know! :)

Cargo.toml
```diff
@@ -49,10 +49,12 @@ path = "benches/macrobenches.rs"
 [features]
 default = []
 
-## Enables support for asynchronous reading from `tokio`'s IO-Traits by enabling
+## Enables support for asynchronous reading and writing from `tokio`'s IO-Traits by enabling
 ## [reading events] from types implementing [`tokio::io::AsyncBufRead`].
 ##
 ## [reading events]: crate::reader::Reader::read_event_into_async
+##
+## [writing events] to buffers implementing [`tokio::io::AsyncWrite`].
```
Mingun, line 57: The `##` doc comments is usual rustdoc markdown comments, which is processed by `document-features` crate. So please rewrite that according to that. If you want to make hyperlink on `writing events`, please add hyperlink target

src/writer/async_tokio.rs
```diff
@@ -339,4 +484,93 @@ mod indentation_async {
 </paired>"#
         );
     }
+
+    #[tokio::test]
+    async fn element_writer_empty() {
```
Mingun, line 489: Instead on using prefix `element_writer_` please put all tests i a module `element_writer`:
```rust
mod element_writer {
    use super::*;

    #[tokio::test]
    async fn empty() { ... }
    ...
}
```

src/writer/async_tokio.rs
Mingun: You could remove `Write` requirement from `ElementWriter` struct and from impl block for attributes and put `Write` or `AsyncWrite + Unpin` to individual impl blocks. Then new functions should be named with `_async` suffix and no need separate method to create async element writer.

Mingun (changes requested):
Good work! Three issues:
- documentation on feature incorrectly formatted
- please use modules to grouping tests instead of common prefix
- I would prefer to use one `ElementWriter` struct with sets of methods for sync and async writing

Cargo.toml
tevoinea: I think I maybe misunderstood the docs there so I removed that change.

src/writer/async_tokio.rs
tevoinea: That's a good idea, I like that structure a lot more!

src/writer/async_tokio.rs
```diff
@@ -57,6 +60,58 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
         Ok(())
     }
 
+    // /// Provides a simple, high-level async API for writing XML elements.
```
Mingun, line 63: Just remove the whole commented block. You always can restore the previous version by checkouting a commit that GitHub shows in a PR

src/writer/async_tokio.rs
```diff
@@ -82,6 +137,95 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
...
+    ///
+    /// The yielded items must be convertible to [`Attribute`] using `Into`.
+    pub fn with_attributes_async<'b, I>(mut self, attributes: I) -> Self
+    where
+        I: IntoIterator,
+        I::Item: Into<Attribute<'b>>,
+    {
+        self.start_tag.extend_attributes(attributes);
+        self
+    }
```
Mingun, line 160: This methods does not depend from writer, so remove them. Move existing `with_attribute` and `with_attributes` methods to the `impl` block that does not restrict `W` type parameter, then they will work for both sync and async writers

src/writer/async_tokio.rs
```diff
@@ -340,3 +484,99 @@ mod indentation_async {
         );
     }
 }
+
+#[cfg(test)]
+mod element_writer {
```
Mingun, line 489: I supposed that this module will be inside the `tests` module, but that's also OK

src/writer/async_tokio.rs
```diff
@@ -340,3 +484,99 @@ mod indentation_async {
...
+    use super::*;
+    use crate::events::*;
+    use pretty_assertions::assert_eq;
+
+    #[tokio::test]
+    async fn empty() {
```
Mingun, line 495: Hm... maybe move all those tests to doc comments of corresponding methods? I think, this will be useful. This one for `write_empty_async`

src/writer/async_tokio.rs
```diff
@@ -340,3 +484,99 @@ mod indentation_async {
...
+        tokio_buffer.flush().await.expect("flush failed");
+
+        assert_eq!(
+            std::str::from_utf8(&buffer).unwrap(),
+            r#"<empty attr1="value1" attr2="value2"/>"#
+        );
+    }
+
+    #[tokio::test]
+    async fn text() {
```
Mingun, line 517: This for `write_text_content_async`

src/writer/async_tokio.rs
```diff
@@ -340,3 +484,99 @@ mod indentation_async {
...
+        tokio_buffer.flush().await.expect("flush failed");
+
+        assert_eq!(
+            std::str::from_utf8(&buffer).unwrap(),
+            r#"<paired attr1="value1" attr2="value2">text</paired>"#
+        );
+    }
+
+    #[tokio::test]
+    async fn nested() {
```
Mingun, line 539: And this for `write_inner_content_async`

Mingun (changes requested):

src/writer/async_tokio.rs
tevoinea: Since `with_attribute` is on the `ElementWriter`, the only way to do this is by removing the `Write` bound on the `ElementWriter` struct (in order to support a general impl. I still made the change.

src/writer/async_tokio.rs
```diff
@@ -82,6 +84,178 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
...
+    ///     .await
+    ///     .expect("failure");
+
+    /// tokio_buffer.flush().await.expect("flush failed");
+
+    /// assert_eq!(
+    ///     std::str::from_utf8(&buffer).unwrap(),
+    ///     r#"<paired attr1="value1" attr2="value2">text</paired>"#
+    /// );
+    /// }
```
Mingun, line 115:
```suggestion
    /// # }
```

src/writer/async_tokio.rs
```diff
@@ -82,6 +84,178 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
...
+    ///        .await
+    ///        .expect("failure");
+    ///
+    ///     tokio_buffer.flush().await.expect("flush failed");
+    ///
+    ///     assert_eq!(
+    ///      std::str::from_utf8(&buffer).unwrap(),
+    ///      r#"<empty attr1="value1" attr2="value2"/>"#
+    ///     );
+    /// }
```
Mingun, line 181:
```suggestion
    /// # }
```
And also please dedent the visible code in example

src/writer/async_tokio.rs
```diff
@@ -82,6 +84,178 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
...
+    ///             r#"<outer attr1="value1" attr2="value2">
+    ///     <fruit quantity="0">apple</fruit>
+    ///     <fruit quantity="1">orange</fruit>
+    ///     <fruit quantity="2">banana</fruit>
+    ///     <inner>
+    ///         <empty/>
+    ///     </inner>
+    /// </outer>"#
+    ///         );
+    /// }
```
Mingun, line 239:
```suggestion
    /// # }
```
And also please dedent the visible code in example and check that indent is correct (the first visible line has different indent that the others)

src/writer/async_tokio.rs
```diff
@@ -82,6 +84,178 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
     }
 }
 
+impl<'a, W: AsyncWrite + Unpin> ElementWriter<'a, W> {
+    /// Write some text inside the current element.
+    /// # Example
+    ///
+    /// ```rust
```
Mingun, line 91:
```suggestion
    /// ```
```
The code blocks already have `rust` type by default

src/writer/async_tokio.rs
```diff
@@ -82,6 +84,178 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
...
+        self.writer
+            .write_event_async(Event::End(self.start_tag.to_end()))
+            .await?;
+        Ok(self.writer)
+    }
+
+    /// Write an empty (self-closing) tag.
+    /// # Example
+    ///
+    /// ```rust
```
Mingun, line 157:
```suggestion
    /// ```
```
The code blocks already have `rust` type by default

src/writer/async_tokio.rs
```diff
@@ -82,6 +84,178 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
...
+        self.writer
+            .write_event_async(Event::Empty(self.start_tag))
+            .await?;
+        Ok(self.writer)
+    }
+
+    /// Create a new scope for writing XML inside the current element.
+    /// # Example
+    ///
+    /// ```rust
```
Mingun, line 192:
```suggestion
    /// ```
```
The code blocks already have `rust` type by default

Mingun:
Also, please add entry to `Changelog.md`:
- Short description here from line 40: https://github.com/tafia/quick-xml/blob/cced485d5331e7122ac5522bcb53cd72de6b8476/Changelog.md?plain=1#L39-L40
- A link to this PR here on line 51: https://github.com/tafia/quick-xml/blob/cced485d5331e7122ac5522bcb53cd72de6b8476/Changelog.md?plain=1#L50-L51
- _(keep two lines between list of links and the previous release section)_

Mingun:
Thanks for your work!

PR #640, opened 2023-08-27
Use thiserror crate to make error definition more ergonomic.

### LINKED ISSUE #638: Would using thiserror crate be acceptable for the error handling?
I would like to refactor the errors crate to use the [thiserror](https://crates.io/crates/thiserror) crate. The patterns implemented in the module are very similar to the crate. The crate is widely used in rust libs. I think it would probably raise the quality of the code a bit to use the crate. Would this be an interesting change?

wt:
The thiserror crate takes care of a bunch of details of implementing errors for a library in rust. This will simplify the code around our errors by making it follow conventions that are widely used in the rust community. It also is much shorter in terms of lines of code.

Closes #638

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ serde = { version = ">=1.0.100,<1.0.181", optional = true }
 tokio = { version = "1.10", optional = true, default-features = false, features = ["io-util"] }
 memchr = "2.1"
 arbitrary = { version = "1.2.3", features = ["derive"], optional = true }
+thiserror = "1.0.47"
 
 [dev-dependencies]
 criterion = "0.4"
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -2,26 +2,28 @@
 
 use crate::escape::EscapeError;
 use crate::events::attributes::AttrError;
-use crate::utils::write_byte_string;
-use std::fmt;
 use std::io::Error as IoError;
 use std::str::Utf8Error;
 use std::string::FromUtf8Error;
 use std::sync::Arc;
 
 /// The error type used by this crate.
-#[derive(Clone, Debug)]
+#[derive(thiserror::Error, Clone, Debug)]
 pub enum Error {
     /// IO error.
     ///
     /// `Arc<IoError>` instead of `IoError` since `IoError` is not `Clone`.
+    #[error("I/O error: {0}")]
     Io(Arc<IoError>),
     /// Input decoding error. If `encoding` feature is disabled, contains `None`,
     /// otherwise contains the UTF-8 decoding error
+    #[error("Malformed input, decoding impossible")]
     NonDecodable(Option<Utf8Error>),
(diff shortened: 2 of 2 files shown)


wt:
It should be noted that this is currently a draft as it only changes the main error type and not the other errors.

FWIW, I really like how it makes the code more concise.

Mingun (commented):
Thanks for a PR!

As I explained in https://github.com/tafia/quick-xml/issues/638#issuecomment-1693360806, I would not merge it right now, because new dependency increases MSRV and does not give any significant improvement, but you a free to continue working on it and add such improvements

dralley:
>As I explained in https://github.com/tafia/quick-xml/issues/638#issuecomment-1693360806, I would not merge it right now, because new dependency increases MSRV and does not give any significant improvement, but you a free to continue working on it and add such improvements

It should be noted that `thiserror` only requires an MSRV of 1.56, which is aligned with several other popular crates such as the latest version of `syn`. 

1.56 was released October 2021 and is the "Rust 2021 edition" release.  I personally don't have any objections with upgrading to that.

wt:
@Mingun what would be the timeline for moving to an MSRV of 1.56?

Mingun:
I think as soon as we would need to use some feature that is available only since 1.56. I think, that adding a new dependency, that increases MSRV and compile time only to save few lines of code is not a big win, but some user could get upset. So I think we should wait until `thiserror` helps us to implement a really useful feature. As I noted before, this feature could be a backtrace capturing if capturing backtrace requires to do something special instead of getting it automatically by free.

Also, investigating MSRV of our primary users also would be a good indicator whether it makes sense to increase it in near future or not.

dralley:
`serde_derive` has an MSRV of 1.56 https://github.com/serde-rs/serde/blob/master/serde_derive/Cargo.toml

wt:
I think it's important to define a time period that MSRV is trying to capture. In my mind, two years of rust releases seems like a good place to be right now. Do you agree?

dralley:
@Mingun As noted in my previous comment I think for most of our users it is possible that our practical MSRV is in fact already 1.56 due to `serde_derive`

wt:
FWIW, I was trying to get consensus on the MSRV's underlying purpose. I agree with your analysis.

dralley:
So the MSRV is bumped.  Are we yes or no on this PR?  (if yes, it needs a rebase)

Mingun:
Not right now. I think, we should not introduce new dependences just because we can. As I already mentioned, if this PR will introduce some useful feature that we doesn't have right now (for example, backtraces, I found that it is difficult to understand where the parser error occurs when I examining some bug reports. Again, I'm not sure that backtraces need some special support from error types, but we don't have them right now when tests fails, so I think, that backtraces should be explicitly requested. That question needs investigation), then I merge it. But if it just changes the way of error declaration, it not a reasonable reason to have it.

PR #642, opened 2023-09-05
Add workflow to run benchmarks

Mingun:
Using https://github.com/boa-dev/criterion-compare-action. Let's see what it can give us

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,13 @@
+on: [pull_request]
+name: benchmark pull requests
+jobs:
+  runBenchmark:
+    name: run benchmark
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


dralley:
There's just too much noise with github actions to get good results.  It looks like there are swings of up to 13% in the test run.

The criteron book recommends using something like IaI instead https://github.com/bheisler/criterion.rs/blob/master/book/src/faq.md#how-should-i-run-criterionrs-benchmarks-in-a-ci-pipeline

Mingun:
Yeah, I read that, but if such actions is exist, it would be interesting to check them. Actually, I want to see a graph somewhere, similar to which https://github.com/benchmark-action/github-action-benchmark is produced, so to have an idea of performance, but that action unfortunately doesn't work with criterion.

I agree, that it is not worth to merge it right now

PR #643, opened 2023-09-05
Bump MSRV to 1.56 - 2021 edition

dralley:


Mingun:
> our effective MSRV is higher than 1.52 if you are using various features.

This is expected. MSRV means absolute minimal version without any optional dependencies. Many our dependents does not use `serde` functionality, so for them effective MSRV still 1.52. But I made investigation and actually almost all of our dependents use either edition 2021 (this is rust 1.56, right?), or MSRV>=1.56. In some rare cases none of those is specified.

I grabbed versions of some crates (those which set `rust-version` explicitly, but many other crates specify rust 2021), listed according to https://lib.rs/crates/quick-xml/rev a week ago:
```ini
2018=1.31
2021=1.56?

inferno=1.59?
plist=1.59
wayland-scanner=1.65
zbuz=1.64
quick-junit=1.66
phonenumber=1.58.0
opendal=1.65
object_store=1.62.1
calamine=1.63
lcov2cobertura=1.59
tauri-winrt-notification=1.59
poem-openapi=1.64
azure_core=1.70
poem=1.64
gtk4-macros=1.70
starship=1.65
nu-command=1.60
ndk-build=1.60
rio_xml=1.60
wayrs-client=1.65
rusty-s3=1.61
cargo-pants=1.69
mavlink=1.60.0
gvdb=1.65
sparesults=1.60
rpki=1.63
dxr=1.65.0
wiktionary-dump-parser=1.58.1
fefix=1.56
office=1.63
threemf=1.56
calyx=1.66
s3-server=1.59
```

So I think we can bump our MSRV, but let's test MSRV without optional features. If someone uses `serde`, it's usually includes it itself and can rely on serde's MSRV directly.

PR #644, opened 2023-09-09
Fix benchmarks in `compare` project

Mingun:
After #490 attributes should be names starting from `@`. The RSS document used in benchmarks have attributes that mapped to a struct, that was not updated for that, which is now fixed.

Also updated all dependencies of `compare` and `fuzz` projects and switch them to Rust 2021.


Mingun:
Interesting, that almost in all benchmarks `maybe_xml` is the first and `quick_xml` is the second. This is very impressful, because `maybe_xml` implements parsing using only the standard library and does not have any dependencies at all.

Issue #590, opened 2023-04-11
Unexpected Bang Since 0.23.1


JojiiOfficial:
The following file can't be parsed since 0.23.1 as it throws an UnexpectedBang(69) when parsing with the given example code. Simply ignoring the error causes the given example code not to parse anything at all. Version 0.22.0 worked fine
[JMdict.zip](https://github.com/tafia/quick-xml/files/11202365/JMdict.zip)
This is just a trimmed file of the JMdict file to demonstrate the issue. You can find the original file [here](http://ftp.edrdg.org/pub/Nihongo/JMdict.gz).

Working v0.22.0 Code:
```rust
use quick_xml::{events::Event, Reader};
use std::{fs::File, io::BufReader};

fn main() {
    let mut reader = Reader::from_reader(BufReader::new(File::open("./JMdict_short").unwrap()));
    reader.trim_text(true);
    let mut buf = vec![];

    loop {
        match reader.read_event(&mut buf) {
            Ok(Event::Start(ref e)) => {
                if let b"entry" = e.name() {
                    println!("{}", std::str::from_utf8(e.name()).unwrap());
                }
            }

            Ok(Event::Eof) => break,

            Err(e) => {
                panic!("{e:?}");
            }

            _ => (),
        }
    }
}
```

Broken v0.28.1 Code (same code, just migrated):
```rust
use quick_xml::{events::Event, Reader};
use std::{fs::File, io::BufReader};

fn main() {
    let mut reader = Reader::from_reader(BufReader::new(File::open("./JMdict_short").unwrap()));
    reader.trim_text(true);
    let mut buf = vec![];

    loop {
        match reader.read_event_into(&mut buf) {
            Ok(Event::Start(ref e)) => {
                if let b"entry" = e.name().0 {
                    println!("{}", std::str::from_utf8(e.name().0).unwrap());
                }
            }

            Ok(Event::Eof) => break,

            Err(e) => {
                panic!("{e:?}");
            }

            _ => (),
        }
    }
}
```

Mingun:
It seems that this bug is happened when `<!DOCTYPE >` content is too long to fit into `BufReader` buffer.

JojiiOfficial:
Interesting though that this is dependent on the buffer which should be transparent to the xml reader. But indeed increasing the buffer size fixes the issue.

Mingun:
Duplicate of #533

Issue #567, opened 2023-03-01
`Unexpected Event::End(...)` or `missing field '...'`


jimy-byerley:
Hello dear maintainers

I'm trying to alter [`ethercat-esi`](https://github.com/ethercat-rs/ethercat-esi) to use `quick_xml` instead of `serde-xml-rs`. This is pretty straight forward as `ethercat-esi` crate is only using the `serde` macros and calling the xml parser at one place.
```rust 
// ethercat-esi//src/lib.rs
let raw_info: parser::EtherCATInfo = quick_xml::de::from_str(xml)
            .map_err(|e| Error::new(ErrorKind::Other, e.to_string()))?;
```
Running this code (in lib or in a test file of that crate) gives me the following output on both xml files
```rust
$ target/debug/test/some ~/erob-custom.xml
Error: Custom { kind: Other, error: "Unexpected `Event::End(\"Groups\")`" }

// then I comment the Groups section in the xml
$ target/debug/test/some ~/erob-custom.xml
Error: Custom { kind: Other, error: "missing field `ProductCode`" }

// then I change quick-xml version from v0.27.1 to v0.26.0
$ target/debug/test/some ~/erob-custom.xml
Error: Custom { kind: Other, error: "missing field `Sm`" }
```

These fields and the Groups markups are good in the xml files. And it doesn't seems to come from the serde declarations of the xml data (in `ethercat-esi/src/mod.rs`), so I suggest there is an issue with the way the `quick_xml` functions are calling or checking results of the serde deserialization functions. (especially since the error is not the same depending on the version of `quick_xml`)
How do you think I could solve this problem ?

jimy-byerley:
I forgot to add the xml files I was using for tests: [xml.zip](https://github.com/tafia/quick-xml/files/10860770/xml.zip)

Mingun:
This is duplicate of #510 and the error is generated by this (and the others `Option<Vec<T>>`) piece of code:
https://github.com/ethercat-rs/ethercat-esi/blob/3d295ca4c1f8c86f1046a85ce93d14b5d85dcc04/src/parser/mod.rs#L43-L46

This is already fixed in master and I plan to release soon (I think in week or two). You can also change the definition to `Vec<T>` with the default value as suggested [here](https://github.com/tafia/quick-xml/issues/510#issuecomment-1315558671).

jimy-byerley:
Thank you very much for your quick answer
I tried both the solutions you proposed, but I still get the `missing field` issue

[...]

PR #660, opened 2023-10-04
Fix `UnexpectedEof` when deserialize `xs:list`s and newtypes

### LINKED ISSUE #580: `UnexpectedEof` error on 0.28.
Greeting!

Start getting error on 0.28 version (0.27 work fine).

Here example to reproduce error:
P.S. `XmlParser` trait needed because I need also json serialization / deserialization.
```rust
#[derive(Debug, Deserialize)]
pub struct Foo {
    #[serde(default, rename = "newIDs")]
    pub new_ids: Vec<NewIds>,
}

#[derive(Debug, Deserialize)]
pub struct NewIds {
    #[serde(rename = "$value", deserialize_with = "XmlParser::parse")]
    pub objects: Vec<Bar>,
}

#[derive(Debug)]
pub struct Bar {
    pub id: u32,
    pub num: u32,
}

pub trait XmlParser {
    fn parse<'de, D>(deserializer: D) -> Result<Self, D::Error>
    where
        Self: Sized,
        D: Deserializer<'de>;
}

impl XmlParser for Bar {
    fn parse<'de, D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        #[allow(clippy::enum_variant_names)]
        #[derive(Copy, Clone, Deserialize)]
        enum _Bar {
            One(NewId),
        }

        #[derive(Copy, Clone, Deserialize)]
        struct NewId {
            #[serde(rename = "@newID")]
            id: u32,
        }

        let (id, num) = match _Bar::deserialize(deserializer)? {
            _Bar::One(NewId { id }) => (id, 1),
        };

        Ok(Bar { id, num })
    }
}

impl<T> XmlParser for Vec<T>
where
    T: XmlParser,
{
    fn parse<'de, D>(deserializer: D) -> Result<Self, D::Error>
    where
        Self: Sized,
        D: Deserializer<'de>,
    {
        #[derive(Deserialize)]
        struct _T<T: XmlParser>(#[serde(deserialize_with = "XmlParser::parse")] T);

        let arr = Vec::deserialize(deserializer)?
            .into_iter()
            .map(|_T(t)| t)
            .collect();

        Ok(arr)
    }
}

fn main() -> Result<(), Box<dyn Error>> {
    const XML: &str = r#"<Foo>
            <newIDs>
                <One newID="1"    />
                <One newID="2"    />
            </newIDs>
        </Foo>"#;

    println!("{:?}", from_str::<Foo>(XML)?);

    Ok(())
}
```

Mingun:
The problem with #590 was in `SeqItemDeserializer::deserialize_newtype_struct`. Previously it was the same as `SeqItemDeserializer::deserialize_seq`, but deserializing sequences in this deserializer assumes, that those sequences are `xs:list`s, because deserializer itself represents a list element.

Also, I noticed, that `<tag></tag>` (or `<tag/>`) could return and error when try to deserialize `xs:list` from it. Funny, that due to incorrect implementation instead of `Unsupported("unsupported event End")` the `UnexpectedEof` was returned.

Fixes #580

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -20,6 +20,9 @@ MSRV bumped to 1.56! Crate now uses Rust 2021 edition.
 
 ### Bug Fixes
 
+- [#660]: Fixed incorrect deserialization of `xs:list`s from empty tags (`<tag/>`
+  or `<tag></tag>`). Previously an `DeError::UnexpectedEof")` was returned in that case
+
 ### Misc Changes
 
 - [#643]: Bumped MSRV to 1.56. In practice the previous MSRV was incorrect in many cases.
@@ -37,6 +40,7 @@ MSRV bumped to 1.56! Crate now uses Rust 2021 edition.
 [#643]: https://github.com/tafia/quick-xml/pull/643
 [#649]: https://github.com/tafia/quick-xml/pull/646
 [#651]: https://github.com/tafia/quick-xml/pull/651
+[#660]: https://github.com/tafia/quick-xml/pull/660
 
 
 ## 0.30.0 -- 2023-07-23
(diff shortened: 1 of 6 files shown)


Mingun:
Deserializers are still does not fully follow updated documentation for them. They will be aligned with it in the following PR for #567.

PR #663, opened 2023-10-09
Fix incorrect deserialization of vectors of enums from sequences of tags

### LINKED ISSUE #567: `Unexpected Event::End(...)` or `missing field '...'`

Mingun:
This PR fixes #567 which is an instance of generic problem, where top-level deserializer used to deserialize elements inside `<tags>`. It changes things in that way that correct deserializer is used to deserialize types.

The diagram below shows how each struct is related to each other in this PR (`SimpleTypeDeserializer` has more usages, not all of them shown in order to keep diagram small):
```mermaid
flowchart TD
    Deserializer --maps and structs--> ElementMapAccess
    ElementMapAccess --for $text field--> SimpleTypeDeserialzier
    ElementMapAccess --for each field--> MapValueDeserializer
    MapValueDeserializer --for sequences--> MapValueSeqAccess
    MapValueSeqAccess --DeEvent::Start--> ElementDeserializer
    MapValueSeqAccess --DeEvent::Text--> TextDeserializer
    ElementDeserializer --maps and structs--> ElementMapAccess
    TextDeserializer --> SimpleTypeDeserialzier
```

Most commits are prepare work, [the actual fix](https://github.com/tafia/quick-xml/commit/74cd2726ae065bea88eed182e5ceea8388563629#diff-529c3affb2cf6fde2450b68cc6d8ddc24ef85fca368a517310a07cb0cd71b820R1021) is just one line change and yet another for consistence (and I believe also fixes some bug).

The last four commits just some cleanup work, but important because it adds documentation how deserializers work and what is their difference from each other.


src/de/map.rs
```diff
@@ -605,6 +582,41 @@ where
...
+        self,
+        _name: &'static str,
+        _variants: &'static [&'static str],
+        visitor: V,
+    ) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        visitor.visit_enum(crate::de::var::EnumAccess::new(self.map.de))
+    }
```
Mingun, line 609: This call also incorrect, because it will use top-level deserializer for deserializing newtype variants, for example, but I leave fixing that for separate PR. We do not have tests for this situation yet, and I want to approve each change by tests that failed before and pass after the change.

PR #672, opened 2023-10-20
Always deserialize simpleTypes (for example, attribute values) as Some(…)

### LINKED ISSUE #671: xml serde read translates optional attributes with value "" to None
If you want to be able to make a difference between attribute missing and attribute empty `""` should serialize to `Some("")` but those attributes deserialize to `None`.
(useful for roundtrips where the output should be equal to input with small change applied)

```rust

#[derive(Debug, Deserialize, Serialize, PartialEq, Eq)]
struct Name {
    #[serde(rename = "@first_name", skip_serializing_if = "Option::is_none")]
    first_name: Option<String>,
    #[serde(rename = "@second_name", skip_serializing_if = "Option::is_none")]
    second_name: Option<String>,
}

#[test]
fn read_and_write() {
    let expected = Name {
        first_name: Some("aaa".to_string()),
        second_name: Some("".to_string()),
    };

    let name =
        quick_xml::de::from_str::<Name>(r#"<Name first_name="aaa" second_name=""/>"#).unwrap();

    assert_eq!(expected, name);

    let s = quick_xml::se::to_string(&name).unwrap();

    assert_eq!(r#"<Name first_name="aaa" second_name=""/>"#, s);
}
```

fails with

```
thread 'read_and_write' panicked at tests/tezt.rs:21:5:
assertion `left == right` failed
  left: Name { first_name: Some("aaa"), second_name: Some("") }
 right: Name { first_name: Some("aaa"), second_name: None }
```

I found similar issues that are marked as fixed: #252 and #527

Mingun:
This also matches behavior for elements (so we will be more consistent, see changed tests), where `<a/>` is always will be Some(...) because tag is present.

Fixes #671. @francisdb, FYI

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -30,6 +30,8 @@ serde >= 1.0.181
   unit structs, unit variants). `<int>123<something-else/></int>` is no longer valid
   content. Previously all data after `123` up to closing tag would be silently skipped.
 - [#567]: Fixed incorrect deserialization of vectors of enums from sequences of tags.
+- [#671]: Fixed deserialization of empty `simpleType`s (for example, attributes) into
+  `Option` fields: now they are always deserialized as `Some("")`.
 
 ### Misc Changes
 
@@ -58,6 +60,7 @@ serde >= 1.0.181
 [#661]: https://github.com/tafia/quick-xml/pull/661
 [#662]: https://github.com/tafia/quick-xml/pull/662
 [#665]: https://github.com/tafia/quick-xml/pull/665
+[#671]: https://github.com/tafia/quick-xml/issues/671
 
 
 ## 0.30.0 -- 2023-07-23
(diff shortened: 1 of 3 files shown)


francisdb:
I guess this keeps deserializing missing optional attributes as None?

Mingun:
Yes:
|Rust|Attribute|Element|
|----|----|-----|
|`None`|missed|missed|
|`Some("")`|`a=""`|`<a/>`|
|`Some("a")`|`a="a"`|`<a>a</a>`|

See those tests:
https://github.com/tafia/quick-xml/blob/ab8d5193a3fda921c6a68530d8dea98e481cd4cd/tests/serde-issues.rs#L13

francisdb:
I see, the tricky part is in the element.

PR #675, opened 2023-10-28
Rework errors

Mingun:
This PR does cleanup of our errors, making them more consistent. In particular, now syntax errors decoupled from other errors and this will allow us to have `no_std` XML parser in the future.

In addition, I also review the way how errors is reported by the serde deserializer. I already wrote my points about correct deserializer behavior [here](https://github.com/serde-rs/serde/pull/2639#issuecomment-1783793554) and in this PR I implement them -- I avoid to return errors if requested data does not match existing data. Instead deserializer will provide the data and the `Visitor` will return an error if it wish.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -2644,7 +2644,7 @@ where
     /// |[`DeEvent::Start`]|`<any-tag>...</any-tag>`   |Emits [`UnexpectedStart("any-tag")`](DeError::UnexpectedStart)
     /// |[`DeEvent::End`]  |`</tag>`                   |Returns an empty slice. The reader guarantee that tag will match the open one
     /// |[`DeEvent::Text`] |`text content` or `<![CDATA[cdata content]]>` (probably mixed)|Returns event content unchanged, expects the `</tag>` after that
-    /// |[`DeEvent::Eof`]  |                           |Emits [`UnexpectedEof`](DeError::UnexpectedEof)
+    /// |[`DeEvent::Eof`]  |                           |Emits [`InvalidXml(IllFormed(MissedEnd))`](DeError::InvalidXml)
     ///
     /// [`Text`]: Event::Text
     /// [`CData`]: Event::CData
@@ -2652,7 +2652,7 @@ where
         match self.next()? {
             DeEvent::Text(e) => Ok(e.text),
             // allow one nested level
-            DeEvent::Start(_) if allow_start => self.read_text(),
+            DeEvent::Start(e) if allow_start => self.read_text(e.name()),
             DeEvent::Start(e) => Err(DeError::UnexpectedStart(e.name().as_ref().to_owned())),
             // SAFETY: The reader is guaranteed that we don't have unmatched tags
             // If we here, then out deserializer has a bug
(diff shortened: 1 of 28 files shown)


src/errors.rs
```diff
@@ -60,6 +62,36 @@ impl std::error::Error for SyntaxError {}
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+/// An error returned if parsed document is not [well-formed], for example,
+/// an opened tag does not closed before end of input.
```
dralley, line 66: is not closed

src/reader/buffered_reader.rs
```diff
@@ -316,7 +316,7 @@ impl<R: BufRead> Reader<R> {
     /// Manages nested cases where parent and child elements have the _literally_
     /// same name.
     ///
-    /// If corresponding [`End`] event will not be found, the [`Error::UnexpectedEof`]
+    /// If corresponding [`End`] event will not be found, the [`Error::IllFormed`]
```
dralley, line 319:
```suggestion
/// If a corresponding [`End`] event is not found, the an error of type [`Error::IllFormed`]
```

src/de/mod.rs
```diff
@@ -1867,6 +1867,7 @@ macro_rules! deserialize_primitives {
         }
 
         /// Character represented as [strings](#method.deserialize_str).
+        #[inline]
```
dralley, line 1870: Have you actually found this to make a difference?

The primary purpose of `#[inline]` is to allow inlining across compilation units.  Within a compilation unit, the compiler is just going to decide on its own and mostly ignore this attribute unless you say `#[inline(always)]`.

If these functions are not going to be called from any external library, I doubt it would make a difference.

src/errors.rs
```diff
@@ -140,12 +140,24 @@ pub enum Error {
...
-    /// `Event::BytesDecl` must start with *version* attribute. Contains the attribute
-    /// that was found or `None` if an xml declaration doesn't contain attributes.
+    /// A `version` attribute was not found in an XML declaration or is not the
+    /// first attribute.
+    ///
+    /// According to the [specification], the XML declaration (`<?xml ?>`) MUST contain
+    /// a `version` attribute and it MUST be the first attribute. This error indicates,
+    /// that the declaration does not contain attributes at all (if contains `None`)
+    /// or either `version` attribute is not present or not the first attribute in
+    /// the declaration. In the last case in contains the name of found attribute.
```
dralley, line 150: last case in  contains the name of found -> last case it contains the name of the found

src/errors.rs
```diff
@@ -84,6 +84,17 @@ pub enum IllFormedError {
         /// Name of actually closed tag
         found: String,
     },
+    /// A comment contains forbidden double-hyphen (`--`) sequence inside.
+    ///
+    /// According to the [specification], for compatibility, comments MUST NOT contain
+    /// double-hyphen (`--`) sequence, in particular, they cannot end by `--->`.
```
dralley, line 90: Is `--->` meant to be `-->`

dralley (approved):
Looks like a nice change

src/errors.rs
Mingun: No, `--->`. This is due a fact that XML grammar too straightforward and disallow any `--` that does not immediately following by the `>`. Which means that when `--` is part of `--->`, it is disallowed

src/de/mod.rs
Mingun: No, added just for consistence. Actually I want to run some profiler one day, just need time to learn how to do that and find a which one that could be run on Windows. Most tools supports only Linux. `cargo flamegraph` looks promising.

src/reader/buffered_reader.rs
dralley: Ooof, that's a typo in my suggestion.  Apologies.

PR #677, opened 2023-11-04
Introduce `Config` struct that holds parser configuration and implement #513

### LINKED ISSUE #513: Allow to continue parsing after `Error::EndEventMismatch`
I'm trying to parse a Netscape bookmark file which has unclosed tags (i.e. `<DL>`), and for this I'm explicitly ignoring the `EndEventMismatch` error during the parsing:

```rust
match reader.read_event() {
    Err(e) => match e {
        QuickXmlError::EndEventMismatch { expected: _, found: _ } => (),
```

The issue is that an `Eof` event immediately follows this, which causes that the reader stops parsing the rest of the document.

I couldn't find in the documentation if this behaviour is expected, hence why I'm opening the issue.

Mingun:
This PR changes a way how reader is configured. Instead of having builder methods in `Reader` / `NsReader` now they provide a `config()` / `config_mut()` methods:
- the `config()` method introduces the ability to read parser configuration, that was not be possible until now;
- the `config_mut()` method allow to set parser options. You can even to set all options at once or store them (the `Config` struct implements `Serialize` / `Deserialize` when `serde-types` feature is active)

To ensure, that the behavior was not changed, this PR also introduces a new integration test-suite which tests all possible parser options. That allowed me to find a bug in `trim_text_end` option, but it is hard to fix it today. Because I plan to rewrite parser (that task is actually mostly done) I just `ignore` that test for now.

As a nice bonus, after implementing new tests it become obvious how to implement #513, which was done.

Closes #513

### DIFF
--- a/fuzz/fuzz_targets/fuzz_target_1.rs
+++ b/fuzz/fuzz_targets/fuzz_target_1.rs
@@ -17,10 +17,10 @@ where
 {
     let mut writer = Writer::new(Cursor::new(Vec::new()));
     let mut buf = vec![];
-    let reader = reader
-        .expand_empty_elements(true)
-        .trim_text(true)
-        .trim_text_end(true);
+    let config = reader.config_mut();
+    config.expand_empty_elements = true;
+    config.trim_text(true);
+    config.trim_text_end = true;
     loop {
         let event_result = reader.read_event_into(&mut buf);
         if let Ok(ref event) = event_result {
--- a/fuzz/fuzz_targets/fuzz_target_1.rs
+++ b/fuzz/fuzz_targets/fuzz_target_1.rs
@@ -20,7 +20,6 @@ where
     let config = reader.config_mut();
     config.expand_empty_elements = true;
     config.trim_text(true);
-    config.trim_text_end = true;
     loop {
         let event_result = reader.read_event_into(&mut buf);
         if let Ok(ref event) = event_result {
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -11,152 +11,188 @@ use crate::reader::state::ReaderState;
 
 use memchr;
 
-macro_rules! configure_methods {
-    ($($holder:ident)?) => {
-        /// Changes whether empty elements should be split into an `Open` and a `Close` event.
-        ///
-        /// When set to `true`, all [`Empty`] events produced by a self-closing tag like `<tag/>` are
-        /// expanded into a [`Start`] event followed by an [`End`] event. When set to `false` (the
-        /// default), those tags are represented by an [`Empty`] event instead.
-        ///
-        /// Note, that setting this to `true` will lead to additional allocates that
-        /// needed to store tag name for an [`End`] event. However if [`check_end_names`]
-        /// is also set, only one additional allocation will be performed that support
-        /// both these options.
-        ///
-        /// (`false` by default)
-        ///
-        /// [`Empty`]: Event::Empty
-        /// [`Start`]: Event::Start
-        /// [`End`]: Event::End
-        /// [`check_end_names`]: Self::check_end_names
-        pub fn expand_empty_elements(&mut self, val: bool) -> &mut Self {
-            self $(.$holder)? .state.expand_empty_elements = val;
-            self
-        }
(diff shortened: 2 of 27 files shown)


tests/reader-config.rs
```diff
@@ -0,0 +1,44 @@
...
+use quick_xml::events::{BytesEnd, BytesStart, Event};
+use quick_xml::reader::Reader;
+
+mod expand_empty_elements {
+    use super::*;
+    use pretty_assertions::assert_eq;
+
+    /// Self-closed elements should be reported as one `Empty` event
+    #[test]
+    fn false_() {
```
dralley, line 356: Perhaps "enabled" and "disabled" would be better names, to avoid keyword clashes?

tests/reader-config.rs
```diff
@@ -42,3 +42,134 @@ mod expand_empty_elements {
...
+            reader.read_event().unwrap(),
+            Event::Empty(BytesStart::from_content("empty \t\r\n", 5))
+        );
+        assert_eq!(
+            reader.read_event().unwrap(),
+            Event::Text(BytesText::new("text"))
+        );
+        assert_eq!(
+            reader.read_event().unwrap(),
+            Event::Comment(BytesText::new(" comment \t\r\n"))
```
dralley, line 686: Do we have any reason to support trimming the text values of "comments"?  I cannot immediately think of a reason to do that, but perhaps one exists.

tests/reader-config.rs
```diff
@@ -43,6 +44,73 @@ mod expand_empty_elements {
...
+    use pretty_assertions::assert_eq;
+
+    mod false_ {
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        #[test]
+        fn check_end_names_false() {
+            let mut reader = Reader::from_str("<root></root \t\r\n>");
+            reader.trim_markup_names_in_closing_tags(false);
```
dralley: What is it the reason that we only have this for closing tags, attributes?

src/reader/mod.rs
```diff
@@ -11,153 +11,189 @@ use crate::reader::state::ReaderState;
 
 use memchr;
 
-macro_rules! configure_methods {
```
dralley, line 14: I appreciate being able to ditch these macros.

fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -17,10 +17,10 @@ where
...
     let mut writer = Writer::new(Cursor::new(Vec::new()));
     let mut buf = vec![];
-    let reader = reader
-        .expand_empty_elements(true)
-        .trim_text(true)
-        .trim_text_end(true);
+    let config = reader.config_mut();
+    config.expand_empty_elements = true;
+    config.trim_text(true);
+    config.trim_text_end = true;
```
dralley: I slightly dislike the aesthetics here and inability to chain (because of converting methods to members), but ultimately it's an incredibly minor detail

dralley (approved):

tests/reader-config.rs
Mingun: I chose those words because if we will have other options that are not booleans, tests would be named as a config value

tests/reader-config.rs
Mingun: I think not, at least nobody request such feature. If such a request appears, we can add a separate option.

Generally speaking, I would delete the current trim options as they simply do not work correctly for text alternating with CDATA / comments / processing instructions, but I suppose that would break many users. I was thinking about renaming current `Event` into `RawEvent` and `DeEvent` to `Event` and give users stream of `Event`s. The `RawEvent` then would be a low-level event which usually not needed by most users. That is very raw thoughts currently, so I decided to not do revolutional changes for now.

tests/reader-config.rs
Mingun: I think, that this is optimization option. Usually end tags does not contain spaces before `>`, so if we will assume that the name ends immediately before the `>`, we could save some time. Such optimisation make sense only for the closing tags -- for opening tags we in any case should check if it has attributes and find the actual end of tag name.

Issue #624, opened 2023-07-22
EndEventMismatch error when asynchronously reading to end in loop


giorgi-o:
I've been getting an unusual `EndEventMismatch` error when reading from a TCP (TLS) socket in a loop.

The error message is `"Expecting </<tag> found </tag>"` `EndEventMismatch { expected: "<tag", found: "tag" }`.

I managed to write a small(ish) self-contained reproducible example:

<details>
<summary>main.rs</summary>

```rust
use std::time::Duration;

use quick_xml::events::Event as XmlEvent;
use tokio::io::{AsyncRead, BufReader};

struct XmlReader {
    next_message_ready: bool,
}

impl XmlReader {
    fn new() -> Self {
        Self {
            next_message_ready: false,
        }
    }
}

impl AsyncRead for XmlReader {
    fn poll_read(
        mut self: std::pin::Pin<&mut Self>,
        _cx: &mut std::task::Context<'_>,
        buf: &mut tokio::io::ReadBuf<'_>,
    ) -> std::task::Poll<std::io::Result<()>> {
        if !self.next_message_ready {
            return std::task::Poll::Pending;
        }

        let response = "<tag></tag>";
        buf.put_slice(response.as_bytes());
        self.next_message_ready = false;
        std::task::Poll::Ready(Ok(()))
    }
}

struct XmlConsumer {
    reader: quick_xml::Reader<BufReader<XmlReader>>,
}

impl XmlConsumer {
    pub fn new() -> XmlConsumer {
        let reader = XmlReader::new();
        let reader = BufReader::new(reader);
        let reader = quick_xml::Reader::from_reader(reader);

        Self { reader }
    }

    pub fn set_next_message_ready(&mut self) {
        self.reader.get_mut().get_mut().next_message_ready = true;
    }

    pub async fn recv_xml(&mut self) {
        // read start event
        let mut buf = Vec::new();
        let start_event = self.reader.read_event_into_async(&mut buf).await.unwrap();
        println!(
            "\n\nstart_event={:?} buffer position={}",
            start_event,
            self.reader.buffer_position()
        );

        let bytes_start = match start_event {
            XmlEvent::Start(e) => e,
            _ => panic!("Expected Start event"),
        };

        // read until corresponding end event
        let mut buf = Vec::new();
        let span = self
            .reader
            .read_to_end_into_async(bytes_start.name(), &mut buf)
            .await
            .unwrap(); // <- panics here
        println!(
[...]

PR #680, opened 2023-11-13
Add test for cancelling read_event_into_async() future

giorgi-o:
Addresses https://github.com/tafia/quick-xml/issues/624#issuecomment-1807077043

### DIFF
--- a/tests/async-tokio.rs
+++ b/tests/async-tokio.rs
@@ -61,8 +61,14 @@ async fn test_cancel_future() {
     let reader = tokio::io::BufReader::new(source);
     let mut reader = Reader::from_reader(reader);
 
-    for _ in 0..3 {
-        for _ in 0..3 {
+    // async function to read the next message from the reader
+    // this function continually reads from the reader, until a read
+    // times out, at which it returns
+    async fn read_message<T>(reader: &mut Reader<T>)
+    where
+        T: tokio::io::AsyncBufRead + Unpin,
+    {
+        loop {
             let fut = async {
                 // read start event
                 let mut buf = Vec::new();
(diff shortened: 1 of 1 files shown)


Mingun:
Thanks for PR! It is useful. I (or someone else who is interested) use it when I'll deal with the problem, but for now let it hang open.

Mingun:
Probably this bug is OS-dependent. I tried to increase for loops to `0..3000` and didn't get errors on Windows 11

Mingun:
In case if GitHub delete logs -- error is
```
failures:

---- test_cancel_future stdout ----
thread 'test_cancel_future' panicked at tests/async-tokio.rs:79:22:
called `Result::unwrap()` on an `Err` value: IllFormed(MismatchedEnd { expected: "<tag", found: "tag" })
```

giorgi-o:
Note that I am also running Windows 11, and the error occurs consistently on my machine.

Mingun:
Well, then I can say, that you probably should try to debug and fix it yourself. I'll plan to make a PR with completely new parser soon, which maybe will free of this error, so you can wait until that.

Because you already did some debugging, maybe you can rewrite the test to be more straightforward, without loops, just with manual calls in sequence. I think that currently the test unstable because it rely on system timer.

giorgi-o:
I refactored it to be more deterministic, removing the arbitrary loop and replacing the 1ms timeout with 100ms.

As for fixing the issue myself, I'll see if I can make time to get familiar with the codebase, although I unfortunately doubt I can in the next few weeks. Hopefully the new parser addresses the problem.

giorgi-o:
And now it crashes on my machine but not on CI... I'll have to take a closer look

PR #681, opened 2023-11-14
RFC: Re-use a long-lived buffer for prefixing attribute names.

adamreichold:
Instead of allocating a temporary string and invoking the formatting machinery for each attribute, re-use a long-lived buffer and manually build up its contents to reduce the overhead of generated the name exposed to Serde.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -123,6 +123,8 @@ pub fn decode_into(bytes: &[u8], encoding: &'static Encoding, buf: &mut String) 
     buf.reserve(
         decoder
             .max_utf8_buffer_length_without_replacement(bytes.len())
+            // SAFETY: None can be returned only if required size will overflow usize,
+            // but in that case String::reserve also panics
             .unwrap(),
     );
     let (result, read) = decoder.decode_to_string_without_replacement(bytes, buf, true);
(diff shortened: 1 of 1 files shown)


src/de/mod.rs
```diff
@@ -2376,6 +2376,8 @@ where
 
     #[cfg(not(feature = "overlapped-lists"))]
     peek: Option<DeEvent<'de>>,
+
+    key_buf: String,
```
Mingun, line 2380:
```suggestion
    /// Buffer to store attribute name as a field name exposed to serde consumers
    key_buf: String,
```

Mingun (approved):
There a still possible unnecessary allocations from decoding. Maybe it is possible to rewrite it to decode into existing buffer, what do you think? It is unnecessary for that PR but if you wish to try, welcome!

adamreichold:
> Maybe it is possible to rewrite it to decode into existing buffer, what do you think? It is unnecessary for that PR but if you wish to try, welcome!

I will have a look into it. I do fear that this would imply to re-implement quite a bit of [`decode_without_bom_handling_and_without_replacement`](https://docs.rs/encoding_rs/latest/encoding_rs/struct.Encoding.html#method.decode_without_bom_handling_and_without_replacement) though.

adamreichold:
I added another commit introduce a `Decoder::decode_into` helper function which as expected does reproduce a bit of code from `encoding_rs` to will avoid allocating an intermediate buffer if decoding it is necessary. Please have another look.

(I have not opted to use this function in `QNameDeserializer::from_elem` because here, sometimes everything will be borrowed whereas `QNameDeserializer::from_attr` will always use the pre-allocated buffer even if only to prepend `@`, i.e. there is an actualy zero-copy case for `from_elem` and it could regress from always copying into `key_buf` whereas `from_attr` copies at least once in any case.)

Issue #650, opened 2023-09-15
Processing instructions should have a dedicated type


Mingun:
Currently we use `BytesText` for them, but this is a text that represents a text with character and entity references, but that references [MUST NOT be recognized within processing instructions](https://www.w3.org/TR/xml11/#sec-pi).

PR #684, opened 2023-11-21
Add new tests for syntax and ill-formed parser errors and fix... emm... errors

Mingun:
This PR is a prepare step for rewriting parser from scratch. First of all I want to guarantee that behaviour will not be changed, especially conditions on which errors is returned and a final state after the error.

Thanks to that tests, I found several bugs related to result of `.buffer_position()` that should report error position.

This PR changes how some ill-formed XML constructs are processed:
- `<?xml?>` now will be threated as syntactically correct XML declaration, although it [is not well-formed](https://www.w3.org/TR/xml11/#NT-XMLDecl). Note, however, that some [online validators](https://codebeautify.org/xmlvalidator) allows such declaration. Previously it was reported as `Event::PI`. According to the specification, a `version` attribute is mandatory, but I follow the quick-xml ideology, that we do not do checks that could slow down parsing, but that checks are performed by demand. In that case when you call `BytesDecl::version()`. Maybe that is not the better way to do things, but this PR not about it.
- `<??>` now will be parsed as syntactically correct processing instruction, although it [is not well-formed](https://www.w3.org/TR/xml11/#NT-PI), previously error is returned. Again, some online validators allows that. Currently there is no way to enforce well-formedless check for PI, but I think, that could be done in #650

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -82,6 +82,14 @@ pub enum IllFormedError {
     ///
     /// [specification]: https://www.w3.org/TR/xml11/#sec-prolog-dtd
     MissedVersion(Option<String>),
+    /// A document type definition (DTD) does not contain a name of a root element.
+    ///
+    /// According to the [specification], document type definition (`<!DOCTYPE foo>`)
+    /// MUST contain a name which defines a document type (`foo`). If that name
+    /// is missed, this error is returned.
+    ///
+    /// [specification]: https://www.w3.org/TR/xml11/#NT-doctypedecl
+    MissedDoctypeName,
     /// The end tag was not found during reading of a sub-tree of elements due to
     /// encountering an EOF from the underlying reader. This error is returned from
     /// [`Reader::read_to_end`].
@@ -120,6 +128,10 @@ impl fmt::Display for IllFormedError {
             Self::MissedVersion(Some(attr)) => {
                 write!(f, "an XML declaration must start with `version` attribute, but in starts with `{}`", attr)
             }
+            Self::MissedDoctypeName => write!(
+                f,
+                "`<!DOCTYPE>` declaration does not contain a name of a document type"
+            ),
             Self::MissedEnd(tag) => write!(
                 f,
                 "start tag not closed: `</{}>` not found before end of input",
@@ -160,14 +172,6 @@ pub enum Error {
     ///
     /// [`encoding`]: index.html#encoding
     NonDecodable(Option<Utf8Error>),
-    /// A document type definition (DTD) does not contain a name of a root element.
-    ///
-    /// According to the [specification], document type definition (`<!doctype foo>`)
-    /// MUST contain a name which defines a document type. If that name is missed,
-    /// this error is returned.
-    ///
-    /// [specification]: https://www.w3.org/TR/xml11/#NT-doctypedecl
-    EmptyDocType,
     /// Attribute parsing error
     InvalidAttr(AttrError),
     /// Escape error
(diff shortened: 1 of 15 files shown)


src/errors.rs
```diff
@@ -103,6 +114,12 @@ pub enum IllFormedError {
 impl fmt::Display for IllFormedError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
+            Self::MissedVersion(None) => {
```
dralley, line 125: IMO, `DeclWithoutVersion` makes more sense as the name.  `MissedVersion` doesn't provide very much context.

`MissedEnd` is a little better, enough to make it OK in my opinion, but on reflection maybe it would be better to be more explicit e.g. `MissingEndTag`.

src/reader/state.rs
```diff
@@ -114,21 +114,19 @@ impl ReaderState {
...
-                    .unwrap_or(len - 8);
-                if start + 8 >= len {
-                    // Because we here, we at least read `<!DOCTYPE>` and offset after `>`.
-                    // We want report error at place where name is expected - this is just
-                    // before `>`
-                    self.offset -= 1;
-                    return Err(Error::IllFormed(IllFormedError::MissedDoctypeName));
+                match buf[8..].iter().position(|&b| !is_whitespace(b)) {
+                    Some(start) => Ok(Event::DocType(BytesText::wrap(
+                        &buf[8 + start..],
```
dralley, line 119: It would be kind of nice if we could document these constants here and above.  With a bit of thought most of them are not so difficult to figure out, but they're not obvious at first clance.

e.g.

```rust
            BangType::DocType if uncased_starts_with(buf, b"!DOCTYPE") => {
                let doctype_tag_len = 8;
                match buf[doctype_tag_len..].iter().position(|&b| !is_whitespace(b)) {
                    Some(start) => Ok(Event::DocType(BytesText::wrap(
                        &buf[doctype_tag_len + start..],

```

Feel free to tweak the details (such as `const` vs `let`, `doctype_tag_len` vs `DOCTYPE_TAG.len()`)

dralley: Lines 87 - 113 could use the same treatment.

tests/reader-errors.rs
```diff
@@ -325,3 +331,316 @@ mod syntax {
...
+        };
+    }
+
+    // IllFormedError::MissedVersion is generated lazily when you call `BytesDecl::version()`
+
+    err!(missed_doctype_name1("<!DOCTYPE>") => 9: IllFormedError::MissedDoctypeName);
+    //                                  ^= 9
+    err!(missed_doctype_name2("<!DOCTYPE \t\r\n>") => 13: IllFormedError::MissedDoctypeName);
+    //                                         ^= 13
+    ok!(missed_doctype_name3("<!DOCTYPE \t\r\nx>") => Event::DocType(BytesText::new("x")));
```
dralley, line 616: Did you intend to leave these names as `missed_doctype_name3` (and below, `mismatched_end1`, and `double_hyphen_in_comment1`) for testing variants that are well-formed?

dralley (approved):
Improvements are suggested.  This looks OK otherwise.

src/reader/state.rs
Mingun: Yes, that code already slightly rewritten (I'll push draft PR soon) and does not contain magic numbers (or they documented)

tests/reader-errors.rs
Mingun: Yes. The main motivation is that these names of the "test about `MissedDoctypeName` error number one... number two... and so on". The final test is a canary test that ensures that after all we able to parse correct representation without mentioned error.

I could wrap that tests into module named `missed_doctype_name` (snake cased error name), but I decide, that indent level would be not very comfort. The similar names of tests for the particular error is useful for running of only related tests via `cargo test ... -- missed_doctype_name`.

In the end, I thought that just a common name with a number is quite convenient and concise.

src/errors.rs
Mingun: I would like to keep errors with a similar meaning named similar. What about
- `MissingDeclVersion`
- `MissingDoctypeName`
- `MissingEndTag`
- `MismatchedEndTag`
- `UnmatchedEndTag`
- `DoubleHyphenInComment` (_not changed_)

?

dralley: `MissingDeclVersion` is fine, I have no issue with those names.

PR #685, opened 2023-11-22
Add ability to set desired level of escaping special characters in XML

### LINKED ISSUE #362: How to ignore double quotation mark when do serialization？
Before serialization:

r#""349DC700140D7F86A0784842780****""

After serialization:
> &quot349DC700140D7F86A0784842780****&quot

However, What expect:
"349DC700140D7F86A0784842780****"

How to ignore serializing double quotation mark?

Mingun:
Fixes #362

### DIFF
--- a/src/escapei.rs
+++ b/src/escapei.rs
@@ -71,6 +71,16 @@ impl std::error::Error for EscapeError {}
 /// | `&`       | `&amp;`
 /// | `'`       | `&apos;`
 /// | `"`       | `&quot;`
+///
+/// This function performs following replacements:
+///
+/// | Character | Replacement
+/// |-----------|------------
+/// | `<`       | `&lt;`
+/// | `>`       | `&gt;`
+/// | `&`       | `&amp;`
+/// | `'`       | `&apos;`
+/// | `"`       | `&quot;`
 pub fn escape(raw: &str) -> Cow<str> {
     _escape(raw, |ch| matches!(ch, b'<' | b'>' | b'&' | b'\'' | b'\"'))
 }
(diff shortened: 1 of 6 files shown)


src/escapei.rs
```diff
@@ -88,10 +98,35 @@ pub fn escape(raw: &str) -> Cow<str> {
...
+/// attribute value. All other characters not necessary to be escaped, although
+/// for compatibility with SGML they also should be escaped. Practically, escaping
+/// only those characters is enough.
+///
+/// This function performs following replacements:
+///
+/// | Character | Replacement
+/// |-----------|------------
+/// | `<`       | `&lt;`
+/// | `&`       | `&amp;`
```
dralley, line 123: So, IIRC, the reason `>` is often escaped despite the fact that the spec doesn't specifically require it is a particular special case:

>The right angle bracket (>) may be represented using the string " &gt; ", and MUST, [for compatibility](https://www.w3.org/TR/2008/REC-xml-20081126/#dt-compat), be escaped using either " &gt; " or a character reference when it appears in the string " ]]> " in content, when that string is not marking the end of a [CDATA section](https://www.w3.org/TR/2008/REC-xml-20081126/#dt-cdsection).
> 
>In the content of elements, character data is any string of characters which does not contain the start-delimiter of any markup and does not include the CDATA-section-close delimiter, " ]]> ". In a CDATA section, character data is any string of characters not including the CDATA-section-close delimiter, " ]]> ".

https://www.w3.org/TR/2008/REC-xml-20081126/#syntax

Escaping `>` even when not strictly required seems preferable to implementing it exactly as described, from both from a complexity and performance perspective.

However, it is possible that it is such a niche situation as to be not all that important in practice, meaning the simple implementation of minimal escaping would be fine 99.999% of the time.

I am not sure I would make it the default though.

dralley: One other thing I just realized, looking at some examples on StackOverflow from this post:

https://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents

We currently re-use `BytesText` for event types such as `Comment` and `PI`

```
    Text(BytesText<'a>),
    /// Unescaped character data stored in `<![CDATA[...]]>`.
    CData(BytesCData<'a>),
    /// Comment `<!-- ... -->`.
    Comment(BytesText<'a>),
    /// XML declaration `<?xml ...?>`.
    Decl(BytesDecl<'a>),
    /// Processing instruction `<?...?>`.
    PI(BytesText<'a>),
    /// Document type definition data (DTD) stored in `<!DOCTYPE ...>`.
    DocType(BytesText<'a>),
```

But the special characters are intended to not be escaped in those contexts.  That may be a UX problem for our users, if they use the default constructors of `BytesText` which do perform escaping.

Mingun: Yes, I plan to introduce other types for events that does not require escaping.

dralley: What do you think re: first comment?

Mingun: I had an impression, that Java XML library that was part of SDK, uses minimal escape rules, but now I checked -- it uses partial escape rules (`&`, `<` and `>` are escaped only). I'll change default to that

dralley: As far as I'm aware, `libxml2` always escapes `>` as well.

dralley (approved):
We probably ought to at least document that if a user users `QuoteLevel::Minimal` the sequence `]]>` cannot be present in the user's provided values or else they will generate invalid XML.

Mingun:
`]]>` is allowed everywhere except CDATA content. We already [have a warning](https://docs.rs/quick-xml/latest/quick_xml/events/struct.BytesCData.html#method.new) about that.

PR #689, opened 2023-11-29
Fix `buffer_position()` after resuming parsing after `IllFormed` errors

Mingun:
Because since #677 we can continue parsing after `Error::IllFormed` we no longer can change `buffer_position()` so that it points to useful position of error. The best solution would be to add position to the `Error` itself, in form of:
```rust
struct Error {
  kind: ErrorKind, // former Error enum
  offset: usize,
}
```
but that would require massive changes, also in the code that I plan to remove soon during rewrite. To keep changes relative small and do not touch that code I decided to add a new `error_position()` getter which would report position of the error. After rewrite I'll try to implement the best solution.

`reader-error.rs` tests was updated to check `buffer_position()` result after resuming parsing after error. Also a couple of new cases was added, which also allowed me to find a bug in one case.

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -449,7 +449,7 @@ pub type Span = Range<usize>;
 ///   Empty     -- End                   --> ClosedTag
 ///   _ -. Eof .-> Exit
 /// ```
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 enum ParseState {
     /// Initial state in which reader stay after creation. Transition from that
     /// state could produce a `Text`, `Decl`, `Comment` or `Start` event. The next
@@ -490,7 +490,7 @@ enum ParseState {
 ///   BomDetected -- "encoding=..." --> XmlDetected
 /// ```
 #[cfg(feature = "encoding")]
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
 enum EncodingRef {
     /// Encoding was implicitly assumed to have a specified value. It can be refined
     /// using BOM or by the XML declaration event (`<?xml encoding=... ?>`)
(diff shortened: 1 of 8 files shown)


dralley (approved):

dralley:
Ignore that, I hit the button early by mistake.

tests/reader-errors.rs
```diff
@@ -14,23 +14,27 @@ macro_rules! ok {
             fn borrowed() {
                 let mut reader = Reader::from_str($xml);
                 reader.config_mut().enable_all_checks(true);
-                assert_eq!(reader.read_event().unwrap(), $event);
+                assert_eq!(reader.read_event().unwrap(), $event, "Reader");
```
dralley, line 17: Given this is only a few dozen lines of tests, why not just split it up into individual tests under a `mod reader` and `mod ns_reader`?

src/reader/mod.rs
```diff
@@ -692,6 +697,21 @@ impl<R> Reader<R> {
         }
     }
 
+    /// Gets the last error byte position in the input data. If there is no errors
+    /// yet, returns `0`.
+    ///
+    /// Unlike `buffer_position` it will point to the place where it is rationale
```
dralley, line 703: s/rationale/rational, related but different words

src/reader/mod.rs
```diff
@@ -692,6 +697,21 @@ impl<R> Reader<R> {
...
+    /// to report error to the end user. For example, all [`SyntaxError`]s are
+    /// reported when the parser seen EOF inside of some kind of markup. The
```
dralley, line 705: s/seen/sees

dralley (approved):

PR #691, opened 2023-12-01
Update maybe_xml in `compare` project to the latest version and add quick-xml borrowed to compare table

Mingun:
I found, that at least part of our slowdown against maybe_xml is because we benchmarked in buffered mode, while maybe_xml completely allocation-free. Add the same variant for quick-xml

### DIFF
--- a/compare/Cargo.toml
+++ b/compare/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports"] }
-maybe_xml = "0.6"
+maybe_xml = "0.10.1"
 quick-xml = { path = "..", features = ["serialize"] }
 rapid-xml = "0.2"
 rusty_xml = { version = "0.3", package = "RustyXML" }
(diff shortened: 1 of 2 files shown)


dralley (approved):

dralley:
>I found, that at least part of our slowdown against maybe_xml is because we benchmarked in buffered mode, while maybe_xml completely allocation-free. Add the same variant for quick-xml

By what proportion, though?  Have you done any profiling to discover exactly which functionality is making up the rest of the difference?

Mingun:
Not much actually. After the change we win in couple tests, but in other we still 1.5x-2x slower that maybe_xml. I did not ran the profiler yet (I need to find the right tool and how to correctly interpret the results, my experience in profiling is almost zero).

I think about two possible reasons:
- maybe_xml parser consist from `const` functions so maybe it is greatly optimized by the compiler;
- maybe_xml performs lazy detection of end of the tag name in start/empty tags (`<tag attributes...>`). Actually, in the benchmark it never runs the corresponding piece of code. Because tags are significant part of any XML, I think, this could be major cause of slow down.

Mingun:
Right now I'm not know from where I should start. I tried to benchmark the [new parser](https://github.com/tafia/quick-xml/pull/690) and unfortunately it is always 2x slower that current implementation. I expected an improvement or at least the same results. Need to figure out why it so slow, because the new implementation is definitely simpler to understand. It will be great to benchmark and profile the `quick_xml::parser::Parser` separately to see where the bottleneck. I will try to do that.

PR #704, opened 2024-01-12
Fixed expand empty not expanding tags with attributes

M-Valts:
In PR #620 the expand_empty_elements function was added to the serializer. With the current implementation only unit variants are expanded, but if a tag has an attribute it would not expand the tag which seems inconsistent.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -67,6 +67,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 [#677]: https://github.com/tafia/quick-xml/pull/677
 [#684]: https://github.com/tafia/quick-xml/pull/684
 [#689]: https://github.com/tafia/quick-xml/pull/689
+[#704]: https://github.com/tafia/quick-xml/pull/704
 
 
 ## 0.31.0 -- 2023-10-22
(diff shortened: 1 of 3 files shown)


src/se/element.rs
```diff
@@ -467,7 +467,14 @@ impl<'w, 'k, W: Write> SerializeStruct for Struct<'w, 'k, W> {
         self.ser.ser.indent.decrease();
 
         if self.children.is_empty() {
-            self.ser.ser.writer.write_str("/>")?;
+            if self.ser.ser.expand_empty_elements {
+                self.ser.ser.writer.write_char('>')?;
+                self.ser.ser.writer.write_str("</")?;
```
Mingun, line 472:
```suggestion
                self.ser.ser.writer.write_str("></")?;
```

Changelog.md
```diff
@@ -35,6 +35,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 - [#622]: Fix wrong disregarding of not closed markup, such as lone `<`.
 - [#684]: Fix incorrect position reported for `Error::IllFormed(DoubleHyphenInComment)`.
 - [#684]: Fix incorrect position reported for `Error::IllFormed(MissingDoctypeName)`.
+- [#704]: Fix empty tags with attributes not being expanded when `expand_empty_elements` is set to true.
```
Mingun, line 38: You forgot to add actual link in line 70

Mingun (changes requested):
Could you add tests for that? There are no writer-only tests so I suggest to create `tests/writer-config.rs` and add test for expand option there (it then will be filled with tests for other options as well, similarly to `reader-config.rs`).

dralley (approved):

Mingun:
Thanks @M-Valts. I still wait tests to ensure that now it works in all situations.

Mingun:
Thanks, @M-Valts !

PR #705, opened 2024-01-14
Adds NsReader::prefixes to iterate on all the prefixes currently declared

Tpt:
This is useful for some use cases like RDF/XML where prefixes might be interesting to extract

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -29,6 +29,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 - [#362]: Added `escape::minimal_escape()` which escapes only `&` and `<`.
 - [#362]: Added `BytesCData::minimal_escape()` which escapes only `&` and `<`.
 - [#362]: Added `Serializer::set_quote_level()` which allow to set desired level of escaping.
+- [#705]: Added `NsReader::prefixes()` to list all the prefixes currently declared.
 
 ### Bug Fixes
 
@@ -68,6 +69,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 [#684]: https://github.com/tafia/quick-xml/pull/684
 [#689]: https://github.com/tafia/quick-xml/pull/689
 [#704]: https://github.com/tafia/quick-xml/pull/704
+[#705]: https://github.com/tafia/quick-xml/pull/705
 
 
 ## 0.31.0 -- 2023-10-22
--- a/src/name.rs
+++ b/src/name.rs
@@ -617,6 +617,65 @@ impl NamespaceResolver {
             None => ResolveResult::Unbound,
         }
     }
+
+    #[inline]
+    pub fn iter(&self) -> PrefixIter {
+        PrefixIter {
+            resolver: self,
+            // We initialize the cursor to 2 to skip the two default namespaces xml: and xmlns:
+            bindings_cursor: 2,
+        }
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/// Iterator on the current declared prefixes.
+///
+/// See [`NsReader::prefixes`](crate::NsReader::prefixes) for documentation.
+#[derive(Debug, Clone)]
+pub struct PrefixIter<'a> {
+    resolver: &'a NamespaceResolver,
+    bindings_cursor: usize,
+}
+
+impl<'a> Iterator for PrefixIter<'a> {
+    type Item = (PrefixDeclaration<'a>, Namespace<'a>);
+
+    fn next(&mut self) -> Option<(PrefixDeclaration<'a>, Namespace<'a>)> {
+        while let Some(namespace_entry) = self.resolver.bindings.get(self.bindings_cursor) {
+            self.bindings_cursor += 1; // We increment for next read
+
+            // We check if the key has not been overridden by having a look
+            // at the namespaces declared after in the array
+            let prefix = namespace_entry.prefix(&self.resolver.buffer);
+            if self.resolver.bindings[self.bindings_cursor..]
+                .iter()
+                .any(|ne| prefix == ne.prefix(&self.resolver.buffer))
+            {
+                continue; // Overridden
+            }
+            let namespace = if let ResolveResult::Bound(namespace) =
+                namespace_entry.namespace(&self.resolver.buffer)
+            {
+                namespace
+            } else {
+                continue; // We don't return unbound namespaces
+            };
+            let prefix = if let Some(Prefix(prefix)) = prefix {
+                PrefixDeclaration::Named(prefix)
+            } else {
+                PrefixDeclaration::Default
+            };
+            return Some((prefix, namespace));
+        }
+        None // We have exhausted the array
+    }
+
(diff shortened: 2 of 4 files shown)


Tpt:
Sorry for the out-of-the-blue PR. Don't feel bad for rejecting/requesting big changes. Thank you so much for quick-xml

src/reader/ns_reader.rs
```diff
@@ -48,6 +50,12 @@ impl<R> NsReader<R> {
     pub fn config_mut(&mut self) -> &mut Config {
         self.reader.config_mut()
     }
+
+    /// Returns all the prefixes currently declared except the default xml: and xmlns: namespaces.
+    #[inline]
+    pub fn prefixes(&self) -> impl Iterator<Item = (PrefixDeclaration, Namespace)> {
```
Mingun, line 56: What if you wish to store returned iterator in the struct? Is it possible? If yes, you should name returned iterator explicitly

tests/namespaces.rs
```diff
@@ -28,6 +28,11 @@ fn namespace() {
             e
         ),
     }
+    assert_eq!(
+        r.prefixes().collect::<Vec<_>>(),
+        vec![(PrefixDeclaration::Named(b"myns"), Namespace(b"www1"))]
+    );
+
```
Mingun, line 35: I think, you should add those checks after each `read_resolved_event()` in all places where you've add that check

src/name.rs
```diff
@@ -617,6 +617,30 @@ impl NamespaceResolver {
             None => ResolveResult::Unbound,
         }
     }
+
+    #[inline]
+    pub fn iter(&self) -> impl Iterator<Item = (PrefixDeclaration, Namespace)> {
+        self.bindings[2..].iter().enumerate().filter_map(|(i, n)| {
+            // We check if the key has not been overriden
+            if self.bindings[i + 3..]
```
Mingun, line 625: Could you add comments here that explains what the magic numbers 2 and 3 means

src/reader/ns_reader.rs
```diff
@@ -48,6 +50,12 @@ impl<R> NsReader<R> {
     pub fn config_mut(&mut self) -> &mut Config {
         self.reader.config_mut()
     }
+
+    /// Returns all the prefixes currently declared except the default xml: and xmlns: namespaces.
```
Mingun, line 54: I think, we need also adopt tests from `tests/namespace.rs` to include them as doctests here to show the behavior. I suppose to show results for this XML:
```xml
<root>
  <a xmlns="a1" xmlns:a="a2">
    <b xmlns="b1" xmlns:b="b2">
      <c/>
    </b>
    <d/>
  </a>
</root>
```

Mingun (changes requested):
This is nice addition!

Could you add changelog entry and a doctest to better explain the results of the new method. There is also a couple of questions that I left in the code.

src/name.rs
Tpt: Done!

src/reader/ns_reader.rs
Tpt: Done!

src/reader/ns_reader.rs
Tpt: Yes! Done!

tests/namespaces.rs
Tpt: Done!

Tpt (commented):
Thank you for your so quick code review. I have made the changes you suggested

Changelog.md
```diff
@@ -29,6 +29,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 - [#362]: Added `escape::minimal_escape()` which escapes only `&` and `<`.
 - [#362]: Added `BytesCData::minimal_escape()` which escapes only `&` and `<`.
 - [#362]: Added `Serializer::set_quote_level()` which allow to set desired level of escaping.
+- `NsReader::prefixes` to list all the prefixes currently declared.
```
Mingun, line 32:
```suggestion
- [#705]: Added `NsReader::prefixes()` to list all the prefixes currently declared.
```
and add link to this PR at line 70

src/name.rs
```diff
@@ -617,6 +617,62 @@ impl NamespaceResolver {
...
+            // We check if the key has not been overriden by having a look at the namespaces declared after in the array
+            if self.resolver.bindings[self.bindings_cursor..]
+                .iter()
+                .any(|ne| {
+                    namespace_entry.prefix(&self.resolver.buffer)
+                        == ne.prefix(&self.resolver.buffer)
+                })
+            {
+                continue; // Overriden
+            }
```
Mingun, line 654: It seems we can return `None` here because on all next iterations we also enter in this `if`, don't we?

Also, `namespace_entry.prefix(&self.resolver.buffer)` could be calculated outside of closure and I have a feeling that the whole this loop can be embedded.

src/name.rs
```diff
@@ -617,6 +617,62 @@ impl NamespaceResolver {
...
+            } else {
+                continue; // We don't return unbound namespaces
+            };
+            return Some((prefix, namespace));
+        }
+        None // We have exhausted the array
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (0, Some(self.resolver.bindings.len() - 2))
```
Mingun, line 674:
```suggestion
        // -2 for xml and xmlns prefixes that are always stored in bindings
        (0, Some(self.resolver.bindings.len() - 2))
```

Mingun (changes requested):
Thanks! A minor fix in changelog is required and it seems to me a small optimization is possible

src/name.rs
Tpt: > Also, namespace_entry.prefix(&self.resolver.buffer) could be calculated outside of closure and I have a feeling that the whole this loop can be embedded.

Indeed. Done. Thank you!

> It seems we can return None here because on all next iterations we also enter in this if, don't we?

I don't think so because we check only if a namespace of the same name is declared, not if an other namespace in general is declared.

Let's say we have
```xml
<foo:outer xmlns:foo="a" xmlns:bar="b">
  <foo:inner xmlns:foo="c"></foo:inner>
</foo:outer>
```
The `bindings` list when parsing `foo:inner` is going to be something like `[("foo", "a"), ("bar", "b"), ("foo", "c")]` so, when we iterate on it we don't want to stop iterating on the first `("foo", "a")` because it has been overriden by `("foo", "c")`.

src/name.rs
Tpt: Done

Changelog.md
Tpt: Done.

Tpt:
> Thanks! A minor fix in changelog is required and it seems to me a small optimization is possible

Thank you! Done.

Mingun:
I've made returned iterator `Clone`able and `Debug`able and slightly tune the `size_hint` method to return more close upper bound and test it. Implementing `ExactSizeIterator` is unreasonably although, because this will require to go through `self.resolver.bindings` to count exact length.

Thanks for your work, @Tpt!

PR #721, opened 2024-02-28
enhancement: return bytes written from `write_event`

phdavis1027:
`Writer` now reports the number of bytes written by the underlying `Write`. My use case is that I need to send a prelude over TCP reporting how long serialized XML messages will be before I send the actual message.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -20,7 +20,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 
 ### New Features
 
-- [#721]: Report bytes from `Writer::write_event`
+- [#721]: Report the number of bytes from `Writer::write_event`
 - [#513]: Allow to continue parsing after getting new `Error::IllFormed`.
 - [#677]: Added methods `config()` and `config_mut()` to inspect and change the parser
   configuration. Previous builder methods on `Reader` / `NsReader` was replaced by
@@ -62,7 +62,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 - [#689]: `buffer_position()` now always report the position the parser last seen.
   To get an error position use `error_position()`.
 
-[#721]: https://github.com/tafia/quick-xml/pull/721
+[#721]: 
 [#362]: https://github.com/tafia/quick-xml/issues/362
 [#513]: https://github.com/tafia/quick-xml/issues/513
 [#622]: https://github.com/tafia/quick-xml/issues/622
--- a/Changelog.md
+++ b/Changelog.md
@@ -20,6 +20,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 
 ### New Features
 
+- [#721]: Report bytes from `Writer::write_event`
 - [#513]: Allow to continue parsing after getting new `Error::IllFormed`.
 - [#677]: Added methods `config()` and `config_mut()` to inspect and change the parser
   configuration. Previous builder methods on `Reader` / `NsReader` was replaced by
--- a/Changelog.md
+++ b/Changelog.md
@@ -62,6 +62,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 - [#689]: `buffer_position()` now always report the position the parser last seen.
   To get an error position use `error_position()`.
 
+[#721]: https://github.com/tafia/quick-xml/pull/721
 [#362]: https://github.com/tafia/quick-xml/issues/362
 [#513]: https://github.com/tafia/quick-xml/issues/513
 [#622]: https://github.com/tafia/quick-xml/issues/622
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -246,10 +246,10 @@ impl<W: Write> Writer<W> {
                 n += i.current().len();
             }
         }
-        self.write(before)?;
-        self.write(value)?;
-        self.write(after)?;
-        Ok(before.len() + value.len() + after.len() + n)
+        n += self.write(before)?;
+        n += self.write(value)?;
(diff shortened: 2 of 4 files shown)


src/writer.rs
```diff
@@ -231,22 +231,25 @@ impl<W: Write> Writer<W> {
...
                 self.writer.write_all(b"\n")?;
                 self.writer.write_all(i.current())?;
+                n += i.current().len();
             }
         }
         self.write(before)?;
         self.write(value)?;
         self.write(after)?;
-        Ok(())
+        Ok(before.len() + value.len() + after.len() + n)
```
Mingun, line 252: Use the result of `write` instead

Mingun (changes requested):
I think this is valuable addition. Please add a changelog entry. Do not forget to add actual link in section with links, contributors frequently forget about that

Changelog.md
```diff
@@ -20,6 +20,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 
 ### New Features
 
+- [#721]: Report bytes from `Writer::write_event`
```
dralley, line 23: Could you make this 

> Report the number of bytes written from ...

Just so that it's more obvious what it's referring to

dralley:
And reformat, and fix the new fuzzing failure

phdavis1027:
Actually the fuzzer has caught a harder case -- `write_text_content` already has a non-empty type for its result. First thought that jumps to mind is to just return a `Result<(usize, &'a mut Writer<W>)>` but obviously that breaks the kind of chaining used in the tests.

src/writer.rs
```diff
@@ -131,13 +131,14 @@ impl<W> Writer<W> {
...
+    ///         let mut n = 0;
     ///         for (quant, item) in fruits.iter().enumerate() {
-    ///             writer
+    ///             n += writer
     ///                 .create_element("fruit")
     ///                 .with_attribute(("quantity", quant.to_string().as_str()))
     ///                 .write_text_content(BytesText::new(item))?;
     ///         }
-    ///         Ok(())
+    ///         Ok(n)
```
dralley, line 141: Honestly, I think I would prefer not following this pattern for the `ElementWriter` APIs.  You're right, it's messy.

The `ElementWriter` APIs are just for convenience, you can't do anything that you can't also do with the lower level ones.  So I think as long as it's possible to get the behavior you want, we can keep these APIs as convenient and simple as possible.

phdavis1027: Alright, I'll undo those changes and try to finalize this weekend.

dralley:
About the original purpose - are you serializing the XML out to a buffer first?  If so, wouldn't it be easier to just get the length of the data in the buffer?

phdavis1027:
> About the original purpose - are you serializing the XML out to a buffer first? If so, wouldn't it be easier to just get the length of the data in the buffer?

It's very possible I'm missing an obvious alternative solution. Ideally, I'd pack multiple things into one buffer in reverse order (i.e., the last thing serialized is the first element in the buffer.) For example the Header lives somewhere in `buf[4..MAX_HEADER_LEN]`, but I don't know how to determine which slice to send without knowing how many bytes were actually written there by the serializer.

Also, sorry it slipped by this weekend. I'm running the fuzzer on a commit with just the original changes right now but it really beats up my crummy HP. Unless it would blow up y'all's notifications too much I might have to finish this out tomorrow with some test runs on Github.

phdavis1027:
I'm going to second-guess myself here: I just discovered you can use `std::io::Cursor` for basically the same purpose. Doesn't that make this PR redundant?

dralley:
I didn't think of `io::Cursor` specifically but that's kind of what I was wondering with my previous comment.  But you're right, `io::Cursor` does do [something very similar](https://doc.rust-lang.org/src/std/io/cursor.rs.html#322).  I was thinking that perhaps you might not even need that - if you're writing to `Vec<u8>` for instance you could just check the length of the `Vec<u8>`.  `io::Cursor` is a more general solution, though.

I probably could have been more specific about that, sorry.

Anyway, my suggestion would be to try out `io::Cursor` and see if you can find any use case that it doesn't handle.  And if it turns out that you can accomplish everything you need to without this PR then we can ice it for now.

phdavis1027:
> I didn't think of `io::Cursor` specifically but that's kind of what I was wondering with my previous comment. But you're right, `io::Cursor` does do [something very similar](https://doc.rust-lang.org/src/std/io/cursor.rs.html#322). I was thinking that perhaps you might not even need that - if you're writing to `Vec<u8>` for instance you could just check the length of the `Vec<u8>`. `io::Cursor` is a more general solution, though.
> 
> I probably could have been more specific about that, sorry.
> 
> Anyway, my suggestion would be to try out `io::Cursor` and see if you can find any use case that it doesn't handle. And if it turns out that you can accomplish everything you need to without this PR then we can ice it for now.

Thanks for the suggestion. In my particular case I'm writing to a `&mut [u8]`, but it does turn out that Cursor can do everything I need it to. The overhead of creating the Cursor is so minimal that I doubt anyone would need other solutions. Sorry for the frivolous PR!

phdavis1027:
`std::io::Cursor::position()` can report how many bytes any `Writer` wrote. See the following:

```rs
let mut cursor = Cursor::new(buf);
let mut writer = Writer::new(&mut cursor);
... // Write a bunch of stuff
// Once you're done writing...
Ok(cursor.position() as usize)
```

dralley:
No worries at all!  Thanks for the time & discussion.

PR #722, opened 2024-02-29
Don't unnecessarily restrict create_element to borrowed names

2xsaiko:
This allows passing owned strings to create_element by making the bounds for the name the same as in the parameter for BytesStart::new.


Mingun:
Please add a changelog entry. Do not forget to add actual link in section with links, contributors frequently forget about that. Please keep the links list sorted by issue number.

Issue #275, opened 2021-03-09
Writer: Add ability to separate attributes by newlines


phillord:
I can control indentation now on tags, but individual tags and their attributes all appear on one line. As I need to make pretty XML, I need this level of control. As tags are rendered in one, I don't think I can achieve this by writing to the inner writer.

hoijui:
I want to help here/implement this.

Please give any tips @Mingun, especially for the interface/API you'd like for this:

1. Yet an other function for `Writer`, like `Writer::new_with_indent_MORE`
2. Using a configuration struct like xml-rs has it: `Writer::new_with_config(config: &WriterConfig)`
3. ... something else?

## Why I want this

To store XML files in git, it would be nice to have a fast, reliable, one-way git-filter that gets applied to all XML based files (especially interesting for SVGs, to me).
This is one key requirement I see for that purpose, apart form some sorting of elements and attributes, which can be done already with quick-xml. An other - less pressing - thing would be, to try to preserve comments; I have no clue how well quick-xml fares in that.

hoijui:
... thinking about it ....
would it ever be a viable scenario, to auto-convert all XML based files in a git repo, **before** checking them in?
If one did not have 100% trust in my code an in the used XML library and the human or software that wrote the original/input XML to be correct, this could potentially result in data-loss.
From some testing I have done in the past with a few other XML prettifier tools (none rust-based though), I often got out faulty/missing/altered data/content.

How confident would you be about quick-xml in such a scenario?

Mingun:
I can't imagine which API will be convenient without using it. Probably we should start from adding methods to [`ElementWriter`](https://docs.rs/quick-xml/latest/quick_xml/writer/struct.ElementWriter.html). Probably this API can cover basic needs:
```rust
writer
  .create_element("element")
    //.new_line() (1)
    .with_attribute("first", "1")
    .with_attribute("second", "2")
    .new_line()
    .with_attribute("third", "3")
    //.new_line() (2)
    .write_empty();
```
```xml
<!-- result of the code above -->
<element first="1" second="2"
         third="3"/>

<!-- if uncomment (1) - indent depends on indentation settings - 2 spaces here -->
<element
  first="1" second="2"
  third="3"/>

<!-- if uncomment (2)  -->
<element first="1" second="2"
         third="3"
/>
```

> How confident would you be about quick-xml in such a scenario?

[...]

PR #731, opened 2024-03-28
First attempt to add newlines between attributes. (#275)

areleu:
Closes https://github.com/tafia/quick-xml/issues/275 
This is an implementation of the suggestion here: https://github.com/tafia/quick-xml/issues/275#issuecomment-1322157730

It may be a little hacky, I am new to Rust so if I am doing some obvious mistake please tell me.

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -242,13 +242,8 @@ impl<'a> BytesStart<'a> {
     where
         A: Into<Attribute<'b>>,
     {
-        let a = attr.into();
-        let bytes = self.buf.to_mut();
-        bytes.push(b' ');
-        bytes.extend_from_slice(a.key.as_ref());
-        bytes.extend_from_slice(b"=\"");
-        bytes.extend_from_slice(a.value.as_ref());
-        bytes.push(b'"');
+        self.buf.to_mut().push(b' ');
+        self.push_attr(attr.into());
     }
 
     /// Remove all attributes from the ByteStart
@@ -286,6 +281,26 @@ impl<'a> BytesStart<'a> {
             }
         }
         Ok(None)
+    }
+
+    /// Adds an attribute to this element.
+    pub(crate) fn push_attr<'b>(&mut self, attr: Attribute<'b>) {
+        let bytes = self.buf.to_mut();
+        bytes.extend_from_slice(attr.key.as_ref());
+        bytes.extend_from_slice(b"=\"");
+        // FIXME: need to escape attribute content
+        bytes.extend_from_slice(attr.value.as_ref());
+        bytes.push(b'"');
+    }
+
+    /// Adds new line in existing element
+    pub(crate) fn push_newline(&mut self) {
+        self.buf.to_mut().push(b'\n');
+    }
+
+    /// Adds indentation bytes in existing element
+    pub(crate) fn push_indent(&mut self, indent: &[u8]) {
+        self.buf.to_mut().extend_from_slice(indent);
     }
 }
 
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -151,6 +151,8 @@ impl<W> Writer<W> {
         ElementWriter {
             writer: self,
             start_tag: BytesStart::new(name),
+            state: AttributeIndent::NoneAttributesWritten,
+            spaces: Vec::new(),
         }
     }
 }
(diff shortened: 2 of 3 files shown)


src/writer.rs
```diff
@@ -363,6 +363,20 @@ impl<'a, W> ElementWriter<'a, W> {
...
+    pub fn element_new_line(mut self, indentation: usize) -> Self  {
+        if let Some(ref i) = self.writer.indent {
+            self.start_tag.push_newline();
+            let current_indent = i.current();
+            let mut indent = vec![i.indent_char; indentation];
+            indent.extend(current_indent);
+            self.start_tag.push_indent(&indent);
+        } else {
+            self.start_tag.push_newline();
+        }
```
Mingun, line 377: I think, that when indents are not enabled, this should be no-op
```suggestion
        }
```

Mingun (commented):
Thanks for working on it. I think, that my initial proposal is simpler to use. To implement it we need a 3-state enum:
```rust
enum AttributeIndent {
  /// Initial state
  NoneAttributesWritten,
  /// Keep indent that should be used if `new_line()` would be called
  SomeAttributesWritten(usize),
  /// Write specified indent before writing attribute in `with_attribute()`
  Indent(usize),
}
// in new_line()
match state {
  NoneAttributesWritten => state = Indent(self.writer.indent.len()),
  SomeAttributesWritten(i) => state = Indent(i),
  Indent(_) => // write \n
}

// in with_attribute()
let i = match state {
  // neither .new_line() or .with_attribute() not yet called
  NoneAttributesWritten => self.writer.indent.len(),
  // .new_line() was not called, but .with_attribute() was
  SomeAttributesWritten(i) => i,
  Indent(i) => {
    // write \n + self.writer.indent + i
    i
  }
};
// write attribute
state = SomeAttributesWritten(i);
```
The resulting state diagram:
```mermaid
stateDiagram-v2
    [*] --> NoneAttributesWritten 
    NoneAttributesWritten --> SomeAttributesWritten1 : .with_attribute()
    NoneAttributesWritten --> Indent2 : .new_line()

    SomeAttributesWritten1 --> SomeAttributesWritten1 : .with_attribute()
    SomeAttributesWritten1 --> Indent1 : .new_line()

    Indent1 --> SomeAttributesWritten1 : .with_attribute()
    Indent1 --> Indent1 : .new_line()

    SomeAttributesWritten2 --> SomeAttributesWritten2 : .with_attribute()
    SomeAttributesWritten2 --> Indent2 : .new_line()

    Indent2 --> SomeAttributesWritten2 : .with_attribute()
    Indent2 --> Indent2 : .new_line()
```

areleu:
@Mingun Thanks a lot for the review, I implemented your suggestion. I hope I interpreted it properly. Let me know if you think it needs changes, otherwise I will write a couple of tests to  look for edge cases.

src/events/mod.rs
```diff
@@ -71,15 +71,29 @@ pub struct BytesStart<'a> {
     pub(crate) buf: Cow<'a, [u8]>,
     /// end of the element name, the name starts at that the start of `buf`
     pub(crate) name_len: usize,
+    /// element indentation state
+    pub(crate) state: AttributeIndent
```
Mingun, line 75: That field should be in `ElementWriter`

src/writer.rs
```diff
@@ -363,6 +376,19 @@ impl<'a, W> ElementWriter<'a, W> {
...
     }
+
+    /// Push a new line inside an element.
+    pub fn new_line(mut self) -> Self  {
+        if let Some(i) = &self.writer.indent {
+            let state = self.start_tag.state;
+            match state {
+                AttributeIndent::NoneAttributesWritten => self.start_tag.state = AttributeIndent::Indent(i.current_indent_len),
+                AttributeIndent::SomeAttributesWritten(isize) => self.start_tag.state = AttributeIndent::Indent(isize),
+                AttributeIndent::Indent(_) => {self.start_tag.push_newline(i.current())},
```
Mingun, line 387: You need to indent on `Indent(_) + i.current()`, pseudocode:
```suggestion
                AttributeIndent::Indent(ind) => self.start_tag.push_newline(ind + i.current()),
```

src/writer.rs
```diff
@@ -782,3 +808,24 @@ mod indentation {
...
+    .with_attribute(("second", "2"))
+    .new_line()
+    .with_attribute(("third", "3"))
+    .with_attribute(("fourth", "4"))
+    .write_empty().expect("write tag failed");
+
+    assert_eq!(
+        std::str::from_utf8(&buffer).unwrap(),
+        r#"<element first="1" second="2"
+ third="3" fourth="4"/>"#
```
Mingun, line 829: The correct output of this test is:
```suggestion
        r#"<element first="1" second="2"
         third="3" fourth="4"/>"#
```
I think, this is because of line 387 above.

Mingun (changes requested):
I didn't run code myself, but it seems to be correct. Please add other tests, update changelog.md and I think, it would be almost complete.

src/writer.rs
```diff
@@ -340,6 +341,18 @@ impl<W: Write> Writer<W> {
...
+
+/// Track indent inside elements state
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub enum AttributeIndent {
+    /// Initial state
+    NoneAttributesWritten(usize),
+    /// Keep indent that should be used if `new_line()` would be called
+    SomeAttributesWritten(usize),
+    /// Write specified indent before writing attribute in `with_attribute()`
+    Indent(usize),
```
Mingun, line 355: Please put it before `ElementWriter` (so that `struct` and `impl` blocks was not breaked by other type) and no need to make it public, it is used only internally

areleu: Done!

Mingun:
Thanks for you work!

I squashed all your commits, write tests in my standards (nested modules), add documentation.

I realized, that currently, because `BytesStart::push_attribute` is used, an extra space is inserted in front of the attribute. This is probably the reason why you originally put the `state` in `BytesStart`. In general, I have long planned to split event struct for events for reading and writing. I think, this is time to do that, or at least, introduce an internal `BytesStart` for writer where the code for writing attributes could be placed.

In any case, current writer code is not good, because:
- it is not optimal, because it uses intermediate vector to store serialized attributes;
- it does not escape attribute content, so you can put string `" otherattr="vulnerable` in it;
- it does not allow you to choose between `"` and `'` for attribute writing

For this reason, I want to hold onto a little bit of this PR to adapt my old code and then use its results here.

Mingun (approved):
Ok, I found a way to do that without much refactoring.

@dralley, would you like to look? Otherwise it is good to me.

Mingun:
Wait on merge it this, I found that tests are incomplete. In case of 
```xml
<element first="1"
         second="2"/>
```
we always should use spaces, but now indent symbols are used. The fix is trivial, just need to add a buffer to `ElementWriter` with spaces (in order to not allocate it each time) and borrow from it when we want to write such indent. Also introduce a new variant to `AttributeIndent` to distinguish this indent from this situation:
```xml
<element
  first="1"
  second="2"/>
```
To test this situation it is enough adjust tests to use `\t` as indent symbol.

areleu:
> Wait on merge it this, I found that tests are incomplete. In case of
> 
> ```
> <element first="1"
>          second="2"/>
> ```
> 
> we always should use spaces, but now indent symbols are used. The fix is trivial, just need to add a buffer to `ElementWriter` with spaces (in order to not allocate it each time) and borrow from it when we want to write such indent. Also introduce a new variant to `AttributeIndent` to distinguish this indent from this situation:
> 
> ```
> <element
>   first="1"
>   second="2"/>
> ```
> 
> To test this situation it is enough adjust tests to use `\t` as indent symbol.

What about implementing it in such a way that (1) only happens if the indent symbol is set to b` ` otherwise (2) should happen. One could also calculate some sort of `\t` approximation for the latter, something that produces:

 ```
 <element
		first="1"
		second="2"/>
 ```

I'm coming from the opinion that combining indentation symbols is not a good idea.

Mingun:
Yes, I also think, that mixing indentation symbols is a bad idea, because some editors can easily break things by replacing spaces with tabs, but auto-selecting behaviour based on intendation settings seems to me even worse. The user can implement such behaviour yourself using our stupid [plumbing interface](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain).

Issue #347, opened 2021-12-28
Inconsistent namespace stripping with serde


rijenkii:
Example:
```xml
<ns:list ns:value="hello">
    <ns:item ns:value="world!"/>
</ns:list>
```
To correctly serialize/deserialize this document, I have to create following structs:
```rust
#[derive(Debug, Serialize, Deserialize)]
#[serde(rename = "ns:list")]
struct List {
    #[serde(rename = "ns:value")]
    value: String,
    #[serde(rename(deserialize = "value", serialize = "ns:value"))]
    items: Vec<Item>,
}

#[derive(Debug, Serialize, Deserialize)]
struct Item {
    #[serde(rename = "ns:value")]
    value: String,
}
```
Only if I put all these seemingly random attributes, result of serialization is exact same as original document.
And while I understand why I need to put `rename` on the root element and why on the child elements it does nothing, I do not understand why I need this explicit separation of deserialize and serialize renames for the `items` field. `quick-xml` seems to strip namespaces from the tag names when deserializing, which it does not do for attributes, and it makes for a very annoying "dance" with these renames.

rijenkii:
Also, because of namespace stripping, it is currently impossible to deserialize following xml:
```xml
<list>
    <ns1:item value="hello" />
    <ns2:item value=123 />
</list>
```
Naive approach does not work, as expected:
```rust
#[derive(Debug, serde::Deserialize)]
#[serde(rename = "list")]
struct List {
    #[serde(rename = "ns1:item")]
    item1s: Vec<Item1>,
    #[serde(rename = "ns2:item")]
    item2s: Vec<Item2>,
}

#[derive(Debug, serde::Deserialize)]
struct Item1 {
    value: String,
}

#[derive(Debug, serde::Deserialize)]
struct Item2 {
    value: u32,
}

fn main() {
    let xml = r#"<list><ns1:item value="hello"/><ns2:item value=123/></list>"#;
    println!("{:#?}", quick_xml::de::from_str::<List>(xml).unwrap());
}
```
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("missing field `ns1:item`")', src/main.rs:22:60
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Issue #212, opened 2020-05-15
Inconsistency with the use of name vs local_name in Serde deserialization implementation.


blankname:
```rust
use quick_xml::de::from_str;
use serde::Deserialize;

#[derive(Debug, Deserialize)]
struct Child {
}

#[derive(Debug, Deserialize)]
struct Parent {
  #[serde(rename = "child")]
  children: Vec<Child>
}

fn main() {
    let xml = r#"<parent xmlns:ns="http://example.org/namespace">
                  <child />
                  <ns:child />
                 </parent>"#;
    let p: Parent = from_str(xml).unwrap();
}
```

Gives the following error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("duplicate field `child`")', src/main.rs:19:21
```

If I understand correctly, this is because structs are being deserialized as maps using the `MapAccess` struct, which uses the element's `local_name` as the key.

So, for the purpose of mapping the child elements to the fields on the Parent struct they both have the same key.

However the `SeqAccess` struct, which is used to create the actual `Vec<Child>` doesn't use `local_name`.

This means that when `SeqAccess` gets to `<ns:child/>` it treats the element as having a different name than `<child/>` so it thinks the sequence has ended.

Control then returns back to `MapAccess` but `MapAccess` immediately tries mapping `<ns:child />` to the same same field (`children`), which it's already filled so it treats it as a duplicate field.

Do you think it would be desirable to replace the use of `name` with `local_name` in the deserialization implementation?

Mingun:
Duplicate of #347

PR #736, opened 2024-04-12
de: consider local name only for namespaced tags in structs with `$value`

### LINKED ISSUE #683: Repeated choice plus other field

Xiphoseer:
This updates the "not_in" check that decides whether to pass a new start tag within a struct to a $value field, to only consider the local part of a QName. It now uses the same decode_name function as the QNameDeserializer that is used for keys/fields already to ensure they stay in sync.

Using the namespaced name in serde (i.e. `#[serde(rename = "ns1:tag")]`) fails with ``Custom("missing field `ns1:tag`")`` today, so this will not break existing code.

Might help with #347

### DIFF
--- a/src/de/key.rs
+++ b/src/de/key.rs
@@ -23,7 +23,7 @@ macro_rules! deserialize_num {
 /// The method will borrow if encoding is UTF-8 compatible and `name` contains
 /// only UTF-8 compatible characters (usually only ASCII characters).
 #[inline]
-fn decode_name<'n>(name: QName<'n>, decoder: Decoder) -> Result<Cow<'n, str>, DeError> {
+pub(super) fn decode_name<'n>(name: QName<'n>, decoder: Decoder) -> Result<Cow<'n, str>, DeError> {
     let local = name.local_name();
     Ok(decoder.decode(local.into_inner())?)
 }
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -1192,4 +1192,18 @@ fn test_not_in() {
         not_in(&["some", "tag", "included"], &tag, Decoder::utf8()).unwrap(),
         false
     );
+
+    let tag_ns = BytesStart::new("ns1:tag");
+    assert_eq!(
+        not_in(&["no", "such", "tags"], &tag_ns, Decoder::utf8()).unwrap(),
+        true
+    );
+    assert_eq!(
+        not_in(&["some", "tag", "included"], &tag_ns, Decoder::utf8()).unwrap(),
+        false
+    );
+    assert_eq!(
+        not_in(&["some", "namespace", "ns1:tag"], &tag_ns, Decoder::utf8()).unwrap(),
+        true
+    );
 }
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -789,7 +789,7 @@ fn not_in(
     start: &BytesStart,
     decoder: Decoder,
 ) -> Result<bool, DeError> {
-    let tag = decoder.decode(start.name().into_inner())?;
+    let tag = super::key::decode_name(start.name(), decoder)?;
 
     Ok(fields.iter().all(|&field| field != tag.as_ref()))
 }


Xiphoseer:
Looks like I did not enable the "serialize" feature when running tests locally and got confused by the double negative

Mingun (changes requested):
Could you test if this actually fixes #347? Add regression test into `./tests/serde-issues.rs`. Try to keep original code as much as possible to be sure that's user problem is actually solved, but instead of printing results use assertions. **That test is needed only if #347 is fixed.**

If your patch does not fix the error, then I would like to see a test which will demonstrate working of this fix, so it will not break in the future (you are also free to add it anyway even if regression test fit our needs).

I also interested to check how it behaves with serialization. Ideally, we should be able to deserialize serialized data. I'm not sure that this is always true now, but we strive for this. So if your patch fix that aspect, it also would be welcome to be presented in a test.

When test will demonstrate changes that was made, please add an entry to `Changelog.md` with a short description of what is fixed.

Xiphoseer:
Hey, thank you for the review. My case is probably closer to #212, but that was closed as a duplicate of #347.

I agree that a regression test would be useful so I'll add one when I find time for it. This change doesn't do anything to fix serialization, but it tries to improve upon the current behavior of _only considering the local part of a QName for deserialization_.

My actual case is as follows:

```xml
<ns1:parent>
    <ns1:field />
    <ns1:one />
</ns1:parent>
```
with the following data model
```
#[derive(Deserialize)]
#[serde[renameAll = "camelCase")]
enum Choice {
  One,
  Two
}

#[derive(Deserialize)]
struct Field;

#[derive(Deserialize)]
#[serde[renameAll = "camelCase")]
struct Parent {
  field: Field;
  #[serde(rename = "$value")]
  choice: Vec<Choice>,
}
```

Today, this will try to find "field" as a variant in "Choice" because the `not_in && has_value_field` case triggers when deserializing the struct.

With the change in this PR, it will correctly identify "field" as a key on the struct even for "ns1:field".

Renaming the field to "ns1:field" via serde does not work, which I guess is what #437 is about: The `not_in` would trigger but the `QNameDeserializer` will pick just the local part and fail to find the field so you get "missing field" errors.

I imagine that adding a toggle in `Deserializer` (use local name vs full name for `decode_name`) could be a backwards compatible change to resolve #437.

Mingun:
I checked, and this PR does not solve #347, but solve #683, because it actually does the same thing as #702.

Unfortunately, #683 and #347 seems to contradict each other. #347 requires to consider prefix part of a tag. Well, it should be possible to fix both issues, but this simple fix just solve the #683 case, but makes us a little more far from solving #347 (in particular, this change makes it impossible to deal with [that example](https://github.com/tafia/quick-xml/issues/347#issuecomment-1002362006)).

I would prefer to investigate possibilities to solve both cases (#683 and #347).

cc @leftmostcat

leftmostcat:
> I would prefer to investigate possibilities to solve both cases (#683 and #347).

Would it make sense to, unless the field is named `$text` or `$value`, parse the field name as an XML name and when deserializing, only compare local name to local name?

Xiphoseer:
FWIW this question sent me down a rabbit hole of "canonical" encodings of a namespaced name and/or QName into a single string in XML and whether it's appropriate to use that as a field identifier. [Turns out](https://www.w3.org/2001/tag/issues.html#rdfmsQnameUriMapping-6) that URI isn't the answer to that, instead the `{http://example.com/namespace}element` syntax seems to be established for it. If we assume the `@attribute` syntax is indeed [inspired by XPath](https://www.w3.org/TR/xpath-10/#path-abbrev), it might be reasonable to explore `Q{http://example.com/namespace}element` [from XPath 3.0](https://www.w3.org/TR/xpath-31/#doc-xpath31-BracedURILiteral) for namespace support.

Furthermore, I think it would be _wrong_ to use a serde field name of `ns1:element` unless going for a pure, pre <https://www.w3.org/TR/xml-names/> tag/attribute name handling mode. QNames are defined to be equivalent if their namespace and local name match, so attaching meaning to a specific prefix string value via making it part of the serde property would prevent its use for certain well-formed documents.

In my mind, #347 is more about making sure the _roundtrip_ works without different configuration for serialization and deserialization, regardless of any particular style of `#[serde(...)]` attributes. Suggesting that it's valid to _serialize_ with an arbitrary namespace prefix by annotating the field with a literal prefix `ns1:` not tied to some `xmlns:ns1` machinery is, in my understanding of XML Namespaces, a flawed assumption.

All that to say: After all that reading, I still think using the local part only is the correct first step toward full namespace support.

Mingun:
> Furthermore, I think it would be _wrong_ to use a serde field name of `ns1:element` unless going for a pure, pre https://www.w3.org/TR/xml-names/ tag/attribute name handling mode.

Yes, I agree with that.

> All that to say: After all that reading, I still think using the local part only is the correct first step toward full namespace support.

Yes, that's what I say in my previous comment.

Xiphoseer:
> > Furthermore, I think it would be _wrong_ to use a serde field name of `ns1:element` unless going for a pure, pre https://www.w3.org/TR/xml-names/ tag/attribute name handling mode.
> 
> Yes, I agree with that.
> 
> > All that to say: After all that reading, I still think using the local part only is the correct first step toward full namespace support.
> 
> Yes, that's what I say in my previous comment.

In that case, would it be possible to merge this PR as-is and do improvements to namespace handling as follow ups? Right now this bug (QNameDeserializer and not_in are inconsistent) severely reduces the usability of `$value`.

Mingun:
Yes, I'll combine your PR and #702, add tests and changelog. I'll plan to release new version soon, maybe this weekend. Unfortunately too many job at work and Kreia in KotOR2 is so intriguing. Anyway, thanks for your work!

PR #738, opened 2024-04-19
examples: add flattened enum deserialization

Frankkkkk:
Add a small example about flattened enum deserialization. This is a small workaround around serde's [serde's
issue](https://github.com/serde-rs/serde/issues/1905).

Big thanks to @Mingun for (most) of the code + help!

### DIFF
--- a/examples/flattened_enum.rs
+++ b/examples/flattened_enum.rs
@@ -1,0 +1,109 @@
+//! This example demonstrates how to deserialize enum nodes using an intermediate
+//! custom deserializer.
+//! The `elem` node can either be a `Foo` or a `Bar` node, depending on the `type`.
+//! The `type` attribute is used to determine which variant to deserialize.
+//! This is a workaround for [serde's issue](https://github.com/serde-rs/serde/issues/1905)
+//!
+//! note: to use serde, the feature needs to be enabled
+//! run example with:
+//!    cargo run --example flattened_enum --features="serialize"
+
+use std::fmt;
+
+use quick_xml::de::from_str;
+use serde::de::value::MapAccessDeserializer;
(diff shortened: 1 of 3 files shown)


Mingun:
Thanks for the example. Need to fix lint issues and find a way to enable "serialize" feature for this example or just run it when that feature is active.

Frankkkkk:
It should be fixed, sorry for the oversight. Cheers

examples/flattened_enum.rs
```diff
@@ -0,0 +1,115 @@
...
+//!    cargo run --example flattened_enum --features="serialize"
+
+use std::fmt;
+
+use quick_xml::de::from_str;
+use serde::Deserialize;
+use serde::{
+    de::{self, value::MapAccessDeserializer, MapAccess, Visitor},
+    Deserialize,
+};
```
Mingun, line 18: This line 14 is redundant because you add this import in line 17
Although I prefer to not use nested imports.

```suggestion
use serde::{Deserialize, Deserializer};
use serde::de::{Error, MapAccess, Visitor};
use serde::de::value::MapAccessDeserializer;
```

and replace `serde::Deserializer` with `Deserializer` and `de::Error` with `Error`

Frankkkkk: Fixed; thanks!

Mingun:
Forget to mention: could you add an entry to the Changelog.md about new example, so people will aware that it is exist? Remember, that changelog contains section with changes and list of link targets at the end of each chapter (== version). This is standard markdown way to split link usage and link target.

All everything is good.

Frankkkkk:
done. thanks

Mingun:
Thanks! I slightly improved text in changelog

PR #739, opened 2024-04-20
Allow override of default escape/unescape behavior in more situations

### LINKED ISSUE #734: Question about changing un/escape behavior
Hi all,

Me again! I've circled back to iterating on serialization and deserialization again, and I have a combo question/proposition for a PR.  My project talks to a server which speaks an idiosyncratic dialect of XML in which, for example, proper unescape behavior is to map \&quot; to  \` (backtick). Furthermore, this dialect of XML does not encode \' (single quote/apostrophe) , but it does escape \" (double quote). As it is, the unescape method evaluates standard named entities before it evaluates custom entities, so I can't implement my unescape behavior. Furthermore,  escape only offers `partial_escape` and `escape`, without offering a resolver. I am willing to throw together a PR, but since there are security issues involved I first wanted to poll for thoughts on the changes I'm proposing, which are:

- In[ `unescape_with`](https://docs.rs/quick-xml/latest/src/quick_xml/escapei.rs.html#186), reverse the order of the checks for resolver and named entities, allowing users to override standard XML unescapes.

- Create an `escape_with` family of functions which allow this kind of customization going the other way. 

EDIT: Just to make clear, since the previous issue I raised here I have abandoned the serde API and am hand-parsing my structs use `Reader`

phdavis1027:
Addresses #734, adds tests to verify that no functionality was lost.


Mingun:
I squashed all commits, refactor our code a bit and move tests to doctests.

I didn't include `escape_with` method, because it is not used in quick-xml and it is easy to implement it outside of quick-xml. I'm afraid these changes could negatively impact performance. That requires investigation, and if the impact would be small, we could include such method. For now I leave my implementation here:

```rust
/// Escapes an `&str` with custom replacements. This method does not have any predefined escapes.
///
/// # Example
///
/// ```
/// # use quick_xml::escape::escape_with;
/// # use pretty_assertions::assert_eq;
/// let custom_resolver = |ch: u8| match ch {
///     b'\"' => Some("&amp;"),
///     b'}' => Some("&lcurl;"),
///     b'{' => Some("&rcurl;"),
///     _ => None,
/// };
///
/// let raw = "<&\"'>";
/// let unchanged = escape_with(raw, |_| None);
///
/// let custom = r##"f'A {weird} f-string that says "Hi!"'"##;
/// let changed = r##"f'A &rcurl;weird&lcurl; f-string that says &amp;Hi!&amp;'"##;
///
/// assert_eq!(escape_with(custom, custom_resolver), changed);
/// assert_eq!(raw, unchanged);
/// ```
pub fn escape_with<'input, 'entity, F>(raw: &'input str, mut escape_chars: F) -> Cow<'input, str>
where
    // the lifetime of the output comes from a capture or is `'static`
    F: FnMut(u8) -> Option<&'entity str>,
{
    let bytes = raw.as_bytes();
    let mut escaped = None;
    let mut pos = 0;
    for (i, b) in bytes.iter().enumerate() {
        if let Some(replacement) = escape_chars(*b) {
            if escaped.is_none() {
                escaped = Some(Vec::with_capacity(raw.len()));
            }
            let escaped = escaped.as_mut().expect("initialized");
            escaped.extend_from_slice(&bytes[pos..i]);
            escaped.extend_from_slice(replacement.as_bytes());

            pos = i + 1;
        }
    }

    if let Some(mut escaped) = escaped {
        if let Some(raw) = bytes.get(pos..) {
            escaped.extend_from_slice(raw);
        }
        // SAFETY: we operate on UTF-8 input and search for an one byte chars only,
        // so all slices that was put to the `escaped` is a valid UTF-8 encoded strings
        // TODO: Can be replaced with `unsafe { String::from_utf8_unchecked() }`
        // if unsafe code will be allowed
        Cow::Owned(String::from_utf8(escaped).unwrap())
    } else {
        Cow::Borrowed(raw)
    }
}
```

src/escape.rs
```diff
@@ -221,10 +252,10 @@ where
                 if let Some(entity) = pat.strip_prefix('#') {
                     let codepoint = parse_number(entity, start..end)?;
                     unescaped.push_str(codepoint.encode_utf8(&mut [0u8; 4]));
-                } else if let Some(value) = named_entity(pat) {
-                    unescaped.push_str(value);
                 } else if let Some(value) = resolve_entity(pat) {
                     unescaped.push_str(value);
+                } else if let Some(value) = named_entity(pat) {
+                    unescaped.push_str(value);
```
Mingun, line 258: I also considering to completely remove this `if` then the `unescape_with` will never resolve predefined entities and instead caller will need to call functions for default processing which we make public. Is that something that we want to do?

dralley: I think it would be poor UX to force the user to declare all predefined entities.  These are built into every other XML parser so I don't see a reason to diverge.

Mingun: Yes, I mean making `named_entity` public (as `resolve_predefined_entity` + `resolve_xml_entity` and `resolve_html5_entity`) so the user can call it after processing it's own entities.

PR #743, opened 2024-05-08
Add Deserializer.into_inner destructor

dishmaker:
It's useful to extract position from Deserializer when error occurs.
Line and column can be extracted from inner reader without the need for RefCell inside a custom Cursor.

Other serde crate `rmp_serde` supports `.into_inner()`
https://docs.rs/rmp-serde/latest/rmp_serde/decode/struct.Deserializer.html#method.into_inner

```rust
pub fn from_str_xml<'de, T>(source: &'de str) -> eyre::Result<T>
where
    T: Deserialize<'de>,
{
    use quick_xml::de::Deserializer;
    let cur = Cursor::new(source);

    let mut de = Deserializer::from_reader(cur);
    let result = T::deserialize(&mut de);

    let cur: Cursor<&str> = de.into_inner().into_inner().into_inner();
    let object: T = result.wrap_err_with(|| format!("position: {}", cur.position()))?;

    Ok(object)
}
```

or even better:
```rust
    .wrap_err_with(|| {
        if let Some(before) = source.as_bytes().get(..cur.position() as usize) {
            let line_num = before.iter().filter(|&c| *c == b'\n').count() + 1;
            let line: &[u8] = before.rsplit(|e| *e == b'\n').next().unwrap_or(before);
            let col_num = line.len();
            format!("Ln {}, Col {}", line_num, col_num)
        }else{
            format!("position: {}", cur.position())
        }
        
    })?;
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -34,6 +34,7 @@ to get an offset of the error position. For `SyntaxError`s the range
   it can handle every attribute that does not match existing cases within an enum variant.
 - [#722]: Allow to pass owned strings to `Writer::create_element`. This is breaking change!
 - [#275]: Added `ElementWriter::new_line()` which enables pretty printing elements with multiple attributes.
+- [#743]: Add `Deserializer::get_ref()` to get XML Reader from serde Deserializer
 
 ### Bug Fixes
 
@@ -83,6 +84,7 @@ to get an offset of the error position. For `SyntaxError`s the range
 [#705]: https://github.com/tafia/quick-xml/pull/705
 [#722]: https://github.com/tafia/quick-xml/pull/722
 [#738]: https://github.com/tafia/quick-xml/pull/738
+[#743]: https://github.com/tafia/quick-xml/pull/743
 [#748]: https://github.com/tafia/quick-xml/pull/748
 [`DeEvent`]: https://docs.rs/quick-xml/latest/quick_xml/de/enum.DeEvent.html
 [`PayloadEvent`]: https://docs.rs/quick-xml/latest/quick_xml/de/enum.PayloadEvent.html
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -2436,6 +2436,38 @@ where
         false
     }
 
+    /// Returns the underlying XML reader.
+    ///
+    /// ```
+    /// # use pretty_assertions::assert_eq;
+    /// use serde::Deserialize;
+    /// use quick_xml::de::Deserializer;
+    /// use quick_xml::Reader;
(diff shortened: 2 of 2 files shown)


Mingun:
Yes, nice addition. Could you also add a changelog entry? Some tests would be also good -- maybe add a doc example?

dishmaker:
I've added doc example and an example file

dishmaker:
`cargo test` works now.

dishmaker:
I hope addition of `Reader.inner()` function wasn't too big of a change.

Mingun:
I modified your PR slightly, I hope I didn't lost something important:
- squashed all commits
- rename `inner` to `get_ref` to match the default Rust agreement (see [standard library](https://doc.rust-lang.org/std/io/struct.Cursor.html?search=get_ref#method.get_ref))
- remove explicit `inner_position` tests, doctests is enough
- remove example as it a slightly overburdened in my opinion (too complex types that is unnecessary to show the idea). The existing documentation should be enough

The two things remains about which I'm not sure yet:
- It is really necessary to have `.into_inner()`? For the case of getting position the `.get_ref()` is enough. The danger with `into_inner()` is that by calling it, you will lose all events already read, but not yet used. If you use it to convert `Deserializer` to `Reader` after successful deserialization, you may get unexpected results. It would be better to work out this moment more carefully and not add this method yet.
- I don't like that in order to get `Reader` from the `Deserializer`, you must call `into_inner().into_inner()`, or `get_ref().get_ref()`, or `into_inner().get_ref()`. It seems you will never need to get `IoReader` / `SliceReader` (result of the first `into_inner()` / `get_ref()`). We can rework this by moving `[Slice/Io]Reader::into_inner()` and `get_ref()` into `XmlRead` trait. See the first additional commit in my [`convert`](https://github.com/dishmaker/quick-xml/compare/dishmaker/into_inner...Mingun:quick-xml:convert) branch (it is based on that PR). From the other hand getting the intermediate `[Slice/Io]Reader` can help with the first problem, because `IoReader` can have destruction method which also would return inner buffer. Another consideration is that currently `XmlRead` is object-safe, but changes in `convert` will make it non-object-safe. I think this is fine, because anyway this trait is the implementation detail on which you should not depend

So, what do you think?

Mingun:
Ok, then I tend to keep only `.get_ref()` in their current implementation for now. `.into_inner()` can be added later if we find a good reason to have it. I'll make necessary changes at the end of week and now I want to wait if somebody else wants to add their thoughts.

dishmaker:
❤️

PR #745, opened 2024-05-20
Add .rustfmt.toml, specify edition = "2021"

alisianoi:
There are two common ways to run rustfmt:

  $ cargo fmt
  $ rustfmt

In the first case, Cargo.toml will be used to configure rustfmt. In the second case, Cargo.toml won't be used and rustfmt will assume certain defaults. Specifically, it will default to Rust 2015 edition which for the quick-xml project results in errors like:

[E0670]: async fn is not permitted in Rust 2015

One of the scenarios when rustfmt is called instead of cargo fmt is "rustic" mode in emacs. The developers of "rustic" mode ask to configure .rustfmt.toml:

See: https://github.com/brotzeit/rustic?tab=readme-ov-file#edition-2018


dralley:
@Mingun Rust 2021 maps to v1.56, we need to bump the MSRV sticker in the README

Issue #749, opened 2024-05-27
Add conformance tests


Mingun:
The official test suite is located at https://www.w3.org/XML/Test/. The all links from that page or found by references from that page:
- Links under **Releases** header. Only last test-suite we should use (this is different versions of the same data)
- direct link to the latest test-suite: https://www.w3.org/XML/Test/xmlts20130923.zip
- CVS repository: https://dev.w3.org/cvsweb/2001/XML-Test-Suite/
- Java generator: https://github.com/belingueres/xml-conformance
- NPM package: https://github.com/lddubeau/xml-conformance-suite

There also exist W3C XML Schema 1.1 (XSD 1.1) test suite, maybe would be useful: https://github.com/w3c/xsdtests

The preferred implementation should automatically download test suite if it does not exist locally and run tests. We also could have a copy as a submodule in our repository. I prefer to use default Rust tools, i.e. `cargo test`. Test could be implemented in a separated crate to not slow down everyday workflow.

Mingun:
Some intermediate results of investigation:
1. The official conformance testsuite has very low quality: many repeated tests, weak structure, has errors (file with "pass" in name marked as "invalid", but comment say that it is valid)
2. Found bug in quick-xml: we end processing instruction after the first `>` even if there is no `?` before it. Actually, we produce an error in that case. We should continue search of `?>` until EOF is reached.
3. Processing instructions:
    - continued to the first `?>` which cannot be escaped
    - any parameter entities (`%name;`) or general entities (`&name;`) are not recognized
    - character references (`&#...;`) are not recognized
    - no characters has special meaning. In particular, `'` and `"` does nothing and no need to be paired
    - literal characters restricted in XML still forbidden. So you can write `&#0;` (which is just string `"&#0;"`), but cannot use character U+0000

Issue #623, opened 2023-07-16
Support for embedded raw binary


elrnv:
There are use cases where embedding binary data into an XML document is convenient. Although not officially supported by the XML spec there are valid and popular uses that exist in the wild (e.g. [VTK](https://vtk.org)). Binary VTK files are quite common, in fact the <AppendedData> tag (which can contain binary data) is the default when writing vtk files, and there are plenty of files like this in the wild.

In the past `quick-xml` was able to support binary data, but this has changed and in the current state requires some non-trivial changes for adding support for being able to write and read uninterpreted binary data.

My first [attempt](https://github.com/tafia/quick-xml/pull/555) didn't really address the issue correctly since binary data can have valid '<' and '>' tags, which can trip up the parser, meaning that at least reading must be augmented from the lowest level (probably as part of `Reader`).

As discussed previously the implementation for this should work as an additional feature.

# Reading
I think it is not difficult to add a customization point for the reader in `quick-xml` without being too invasive by adding a "context" field (supplied externally) to Reader that can call ask how many bytes should be read by feeding it context about what has already been read.

# Writing
All types in the serialize part of the code is bound by `fmt::Write`. I haven't found a way to maintain that bound everywhere and also support for writing binary files to an optional or custom `io::Write` type. To enable this I think we have to switch back to `io::Write`. I understand that there are some performance concerns about having to write to a buffer if `io::Write` is used, but I don't see another way of adding support for binary writing without `io::Write`. I can really use some help from the maintainers to address this point.

Mingun:
It would be valuable, if you could link or attach some examples of files and provide links to the explanation, how VTK parser should understand that some byte is not a markup, but a part of a binary data.

I found an archive with some files in https://github.com/elrnv/vtkio/issues/27 and fight now I suspect the following algorithm for resolving ambiguity:
- when read opening tag with a special mark (`<AppendedData encoding="raw">` in an example) the parser turned into special binary reading mode
- in that mode any sequence of bytes are allowed until it contains the closing tag sequence (`</AppendedData>` in an example)
- closing tag sequence should be interpreted in the document encoding (`utf-8` in an example)

So I think we could work on this issue in a few steps:
[...]

PR #750, opened 2024-06-01
Allow `&#0;` character references

Mingun:
We should either restrict all invalid characters both in literal form and as character references, or none of them. Disallowing only the one character is inconsistently. Because checking literal forms means that we should decode and check all the input, this will influence performance. We are not ready to get that performance lost for now. Users of the Reader API could do their own checks themselves.

Ironically, this is what firstly was proposed in #496. After trying to finish that PR I found that we have some unanswered questions (above) which we should work out before we will be create a consistent solution. I think, it will be tight linked with #749.

I think, the best what we can do now is to not check validity of any characters and allow users to that themselves.

cc @sashka

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -80,6 +80,7 @@ resolve predefined entities.
 - [#734]: No longer resolve predefined entities (`lt`, `gt`, `apos`, `quot`, `amp`)
   in `unescape_with` family of methods. You should do that by yourself using the methods
   listed above.
+- [#750]: Remove `EscapeError::EntityWithNull` and allow `&#0;` character reference.
 
 [#275]: https://github.com/tafia/quick-xml/issues/275
 [#362]: https://github.com/tafia/quick-xml/issues/362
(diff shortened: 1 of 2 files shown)


Mingun:
> Some characters (like [control characters](https://github.com/tafia/quick-xml/issues/368), at least in XML 1.1), are legal when escaped but illegal otherwise, while others (like [NUL](https://stackoverflow.com/questions/730133/what-are-invalid-characters-in-xml)) are illegal in both cases.

Yes, it is. My point is that:
- we agreed in #496 that this checks should be optional although disabling them will create non-conforming parser;
- to create good solution some investigation required;
- checking NUL only when resolving character references is anyway does only half of work, we still need to check literal NULs, but this will impact performance; so this should be done carefully and maybe even as an explicit preprocess step from the user side;
- the restriction is anyway seems artificial with roots in С null-terminated strings;
- for some of our users this check is undesirable.

So I think that removing this check should not make things worse and I propose to accept a short-term solution (yes, I know what you thought now) of removing this check and add it later with understanding and formal tests (which we should get after resolving #749).

dralley:
>the restriction is anyway seems artificial with roots in С null-terminated strings;  for some of our users this check is undesirable.

I'm still not certain that disabling them will help this user.  I have a vague recollection (though I cannot find where it was written) that whatever weird nonstandard format they were using was dumping raw binary data into the space between tags (i.e. text event in our parlance) and using an attribute to denote the length so that the parser could entirely skip over that region.  

I could be mixing it up with another issue though.

Mingun:
You are talking about #623, I meant [this](https://github.com/tafia/quick-xml/pull/496#issuecomment-1292669844) and [this](https://github.com/tafia/quick-xml/pull/496#issuecomment-1294142441) comments. With this change we can make life for some people easier untilw we implement proper solution that will check all according to the standard.

dralley:
>You are talking about https://github.com/tafia/quick-xml/issues/623

Ah, correct.

dralley (approved):
I still have mixed feelings but I'm OK with going forwards with this if @sashka confirms that it would be helpful to them

PR #753, opened 2024-06-05
Fix processing instruction parsing

Mingun:
Handle the situation with parsing `<?pi >?>`. Previously we erroneously consume only `<?pi >` and return error about not closed processing instruction.


dralley (approved):
A release soon would be a good idea, the changelog is getting sizable

PR #755, opened 2024-06-12
Fix incorrect missing of trimming all-space text events when `trim_text_start = false` and `trim_text_end = true`

Mingun:
I was refactoring for another task and accidentally found 2 errors:
- in `ReaderState::emit_text` there was a misprint: if string does not contain non-space symbols it remains unchanged instead of making empty
- This error was actually masked by the previous. After fixing it `ReaderState::emit_text` could make empty slice which then translated to `Event::Eof`. However, this function is called in two cases:
  - when we found `<` which switches parser from `InsideText` to `InsideMarkup` state
  - when we found EOF while parser in the `InsideText` state

  `Event::Eof` should be generated only in the latter case.

The changes also shows small performance improvements:
```
> critcmp master trim -t 5
group                                                          master                                 fix-trim-end
-----                                                          ------                                 ------------
One event/Comment                                              1.08    157.2±2.92ns        ? ?/sec    1.00    146.0±2.42ns        ? ?/sec
decode_and_parse_document/rpm_filelists.xml                    1.06     68.4±1.24µs   160.6 MB/sec    1.00     64.5±1.93µs   170.3 MB/sec
decode_and_parse_document_with_namespaces/rpm_filelists.xml    1.07     95.7±1.62µs   114.7 MB/sec    1.00     89.3±1.63µs   122.9 MB/sec
decode_and_parse_document_with_namespaces/rpm_primary.xml      1.06    205.8±4.17µs    98.5 MB/sec    1.00    193.6±3.81µs   104.7 MB/sec
decode_and_parse_document_with_namespaces/rpm_primary2.xml     1.06     66.9±1.38µs   107.1 MB/sec    1.00     63.4±1.34µs   113.1 MB/sec
parse_document_nocopy/rpm_filelists.xml                        1.09     63.1±1.27µs   174.0 MB/sec    1.00     58.1±1.06µs   189.1 MB/sec
parse_document_nocopy/rpm_primary2.xml                         1.05     45.1±1.34µs   159.0 MB/sec    1.00     42.8±0.88µs   167.6 MB/sec
read_event/trim_text = false                                   1.27    206.1±3.85µs        ? ?/sec    1.00    162.9±2.66µs        ? ?/sec
read_event/trim_text = true                                    1.15    199.5±3.72µs        ? ?/sec    1.00    172.8±3.04µs        ? ?/sec
unescape_text/mixed                                            1.05    384.1±6.97ns        ? ?/sec    1.00    364.9±6.66ns        ? ?/sec
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,11 +14,15 @@
 
 ### Bug Fixes
 
+- [#755]: Fix incorrect missing of trimming all-space text events when
+  `trim_text_start = false` and `trim_text_end = true`.
+
 ### Misc Changes
 
 - [#650]: Change the type of `Event::PI` to a new dedicated `BytesPI` type.
 
 [#650]: https://github.com/tafia/quick-xml/issues/650
+[#755]: https://github.com/tafia/quick-xml/pull/755
 
 
 ## 0.32.0 -- 2024-06-10
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -244,13 +244,21 @@ macro_rules! read_event_impl {
                         }
                         ReadTextResult::UpToMarkup(bytes) => {
                             $self.state.state = ParseState::InsideMarkup;
-                            // Return Text event with `bytes` content or Eof if bytes is empty
-                            Ok($self.state.emit_text(bytes))
+                            // FIXME: Can produce an empty event if:
+                            // - event contains only spaces
+                            // - trim_text_start = false
+                            // - trim_text_end = true
+                            Ok(Event::Text($self.state.emit_text(bytes)))
                         }
                         ReadTextResult::UpToEof(bytes) => {
                             $self.state.state = ParseState::Done;
-                            // Return Text event with `bytes` content or Eof if bytes is empty
-                            Ok($self.state.emit_text(bytes))
+                            // Trim bytes from end if required
+                            let event = $self.state.emit_text(bytes);
+                            if event.is_empty() {
+                                Ok(Event::Eof)
+                            } else {
+                                Ok(Event::Text(event))
+                            }
                         }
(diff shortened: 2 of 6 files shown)


src/reader/mod.rs
```diff
@@ -211,7 +211,7 @@ macro_rules! read_event_impl {
     ) => {{
         let event = loop {
             break match $self.state.state {
-                ParseState::Init => { // Go to OpenedTag state
```
dralley, line 214: The commit message says "renamed for brevity".  I'm not sure that brevity is the correct word since these names are longer than the original ones.

I'm fine with the new names, though.

Changelog.md
```diff
@@ -14,11 +14,15 @@
 
 ### Bug Fixes
 
+- [#755]: Fix incorrect missing of trimming all-space text events when
+  `trim_text_start = false` and `trim_text_end = true`.
+
```
dralley, line 19: Re: " events should not be trimmed at boundary of text / CDATA, or text / PI, or text / comment in some cases", is there an issue filed for that with some examples of such boundary cases?

dralley (approved):

src/reader/mod.rs
Mingun: Ah, yes, I confused it with "clarity"

Changelog.md
Mingun: No, such issue does not exist. I made investigation in #520 for serde deserializer and implement reading text properly. I cross-checked that using Java XmlBeans library.

src/reader/mod.rs
Mingun: I fixed commit text in force push

PR #760, opened 2024-06-16
Remove requirement to have Reader when decoding attributes and write anything converted to Event

### LINKED ISSUE #332: Lifetime on Attributes is wrong

Mingun:
This PR includes two changes, which both related to my work of splitting event into two event classes -- for reading and for writing. Events for reading then would only borrow data and this will help to fix #332 and also open doors to `no_std` reader (not sure why it would be needed although).

The first commit just removes unused import that I noticed while working on this.

The second commit relaxes requirements for decoding values from `Attribute`. If you created attribute for writing and then for whatever reason wants to read its value back, you need a `Reader` which you does not have. This commit fixes that.

The third commit makes writing methods take `Into<Event>` instead of `AsRef<Event>`. The former trait give a lesser abilities then `Into`. It allows you to pass your own struct to write, but you need to have event in your struct to be able to take a ref to it. `Into` would allow you to construct event on demand and also would allow to directly pass reader event to the writer method that would require writer event if (when?) event splitting will be implemented. In any case, I think, that taking `Into` is better.

If there are any objections, then unnecessary commits can be removed from PR.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,6 +21,7 @@
 
 - [#760]: `Attribute::decode_and_unescape_value` and `Attribute::decode_and_unescape_value_with` now
   accepts `Decoder` instead of `Reader`. Use `Reader::decoder()` to get it.
+- [#760]: `Writer::write_event` now consumes event. Use `Event::borrow()` if you want to keep ownership.
 
 [#760]: https://github.com/tafia/quick-xml/pull/760
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,11 @@
 
 ### Misc Changes
 
+- [#760]: `Attribute::decode_and_unescape_value` and `Attribute::decode_and_unescape_value_with` now
+  accepts `Decoder` instead of `Reader`. Use `Reader::decoder()` to get it.
+
+[#760]: https://github.com/tafia/quick-xml/pull/760
+
 
 ## 0.33.0 -- 2024-06-21
 
--- a/benches/macrobenches.rs
+++ b/benches/macrobenches.rs
@@ -50,7 +50,7 @@ fn parse_document_from_str(doc: &str) -> XmlResult<()> {
         match criterion::black_box(r.read_event()?) {
             Event::Start(e) | Event::Empty(e) => {
                 for attr in e.attributes() {
-                    criterion::black_box(attr?.decode_and_unescape_value(&r)?);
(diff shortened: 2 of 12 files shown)


dralley (approved):

Mingun:
Usability of the last commit can be demonstrated in #766. There I plan to introduce conception of low-level `RawReader` which produces `RawEvent`s. Those are current `Reader` and `Event` types. Also I plan introduce conception of high-level `Reader` which would produce new `Event`s.

I plan to make `RawEvent` borrow-only, i.e. they will contain `&[u8]`, and `Event` will contain `Cow<[u8]>`, because new `Reader` would produce events from different files. However, I'd want to be able to write both types of events using the same `Writer` (there are no reasons to have `RawWriter`). Because `Event` will not contain `RawEvent` inside, it will not be possible to implement `AsRef<RawEvent>` for it.

dralley:
> However, I'd want to be able to write both types of events using the same Writer (there are no reasons to have RawWriter). Because Event will not contain RawEvent inside, it will not be possible to implement AsRef<RawEvent> for it.

@Mingun is it possible to go the opposite direction, i.e. have `RawEvent` implement `Into<Event>`?

Mingun:
Yes, it will be done that way.

PR #765, opened 2024-06-21
Remove `escape-html` feature and the `resolve_html5_entity` function

Mingun:
Because since #739 custom entity resolution function have precedence over standard one, user can implement resolution of HTML entities by yourself. So I think, we can completely remove HTML 5 support, which in any case was not ideal. Instead of use big `match` statement it is probably better to use `HashMap` with predefined entities and user can do that.


Mingun:
Ok, I'll investigate this later.

Issue #719, opened 2024-02-27
XML deserializate failed when text contains `&()_+-=;`


Xuanwo:
Hi, I'm from [Apache OpenDAL](https://github.com/apache/opendal) Community. I found an interesting cases that `quick-xml` failed to deserializate content like `&()_+-=;`.

## Context

We are developing WebDAV support with [JFrog Artifactory](https://jfrog.com/artifactory/). I'm not sure how JFrog implements WebDAV support, but it seems they do not correctly escape the XML content.

We found that jfrog will return invalid content like 

```xml
<D:href>/artifactory/example-repo-local/bfc28925-81ec-4b69-b494-e2ee0f957ee8/aba0a24a-3a2b-48e2-ae65-ef2dff027458%20!%40%23%24%25%5E%26()_%2B-%3D%3B'%2C.txt/bfc28925-81ec-4b69-b494-e2ee0f957ee8/aba0a24a-3a2b-48e2-ae65-ef2dff027458%20!@%23$%25%5E&()_+-=;',.txt</D:href>
```

`quick-xml` will recongize `&()_+-=;` as an XML entity thus lead to deserializate failure:

```shell
test panicked: stat must succeed: Unexpected (persistent) at stat => deserialize xml

Context:
   service: webdav
   path: 2c9433ab-450c-49e9-bb88-703cb2f327a0 !@#$%^&()_+-=;',.txt

Source:
   Error while escaping character at range 238..244: Unrecognized escape symbol: "()_+-=": Error while escaping character at range 238..244: Unrecognized escape symbol: "()_+-=": Error while escaping character at range 238..244: Unrecognized escape symbol: "()_+-="
```

## To Reproduce

deserializate content like

```xml
<?xml version=\"1.0\" encoding=\"utf-8\" ?>\n
[...]

PR #766, opened 2024-06-21
Rework handling general entity references (`&entity;`)

### LINKED ISSUE #667: How would I parse character references as literal bytes and not codepoints?
I have an element like this:
```xml
<element>&#240;&#159;&#152;&#131;</element>
```

If those characters are literally interpreted, they should be the byte sequence `f0 9f 98 83`, which should be U+1F603, or `😃`. Instead, it expands to `c3 b0 c2 9f c2 98 c2 83` (this sequence is not printable, but you may inspect it [here](https://unicode.link/inspect/utf8:c3.b0.c2.9f.c2.98.c2.83)).

This is very much how this is meant to work, and I am aware of that. Unfortunately this decision wasn't made nor is it controlled by me. So, I'd like to know if there's an obvious way to change how escapes are done without having to do it by just iterating through the bytes returned by a Text event.

Mingun:
This is a big change in handling [general entity references](https://www.w3.org/TR/xml11/#gen-entity) and character references. Open PR early to get feedback.

With this changes we can correctly parse document
```xml
<!DOCTYPE root [
  <!ENTITY root "<root/>">
]>
&root;
```
as equivalent normalized document
```xml
<root/>
```

The updated `custom_entities` example shows how it would be possible to implement requirement from the specification about [parsed general entities](https://www.w3.org/TR/xml11/#dt-parsedent). Serde deserializer did not updated yet, because this is not trivial part and probably that will be done in another PR.

Of course, such change probably makes the performance worse, I didn't measure impact yet.

Closes #667

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,7 +18,7 @@
 Now references to entities (as predefined, such as `&lt;`, as user-defined) reported as a new
 `Event::GeneralRef`.
 Caller can parse the content of the entity and stream events from it as it is required by the
-XML specification.
+XML specification. See the updated `custom_entities` example!
 
 ### New Features
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -34,6 +34,8 @@ XML specification. See the updated `custom_entities` example!
 
 - [#863]: Remove `From<QName<'a>> for BytesStart<'a>` because now `BytesStart` stores the
   encoding in which its data is encoded, but `QName` is a simple wrapper around byte slice.
+- [#766]: `BytesText::unescape` and `BytesText::unescape_with` replaced by `BytesText::decode`.
+  Now Text events does not contain escaped parts which are reported as `Event::GeneralRef`.
 
 [#766]: https://github.com/tafia/quick-xml/pull/766
 [#863]: https://github.com/tafia/quick-xml/pull/863
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,11 +13,20 @@
 
 ## Unreleased
 
+### Significant changes
+
+Now references to entities (as predefined, such as `&lt;`, as user-defined) reported as a new
+`Event::GeneralRef`.
+Caller can parse the content of the entity and stream events from it as it is required by the
+XML specification.
+
 ### New Features
 
 - [#863]: Add `Attributes::into_map_access(&str)` and `Attributes::into_deserializer()` when `serialize`
   feature is enabled. This will allow do deserialize serde types right from attributes. Both methods
   returns the same type which implements serde's `Deserializer` and `MapAccess` traits.
+- [#766]: Allow to parse resolved entities as XML fragments and stream events from them.
+- [#766]: Added new event `Event::GeneralRef` with content of [general entity].
 
 ### Bug Fixes
 
@@ -26,7 +35,9 @@
 - [#863]: Remove `From<QName<'a>> for BytesStart<'a>` because now `BytesStart` stores the
   encoding in which its data is encoded, but `QName` is a simple wrapper around byte slice.
 
+[#766]: https://github.com/tafia/quick-xml/pull/766
 [#863]: https://github.com/tafia/quick-xml/pull/863
+[general entity]: https://www.w3.org/TR/xml11/#gen-entity
 
 
 ## 0.37.5 -- 2025-04-27
--- a/Changelog.md
+++ b/Changelog.md
@@ -27,6 +27,9 @@ XML specification. See the updated `custom_entities` example!
   returns the same type which implements serde's `Deserializer` and `MapAccess` traits.
 - [#766]: Allow to parse resolved entities as XML fragments and stream events from them.
 - [#766]: Added new event `Event::GeneralRef` with content of [general entity].
+- [#766]: Added new configuration option `allow_dangling_amp` which allows to have
+  a `&` not followed by `;` in the textual data which is required for some applications
+  for compatibility reasons.
 
 ### Bug Fixes
 
--- a/benches/macrobenches.rs
+++ b/benches/macrobenches.rs
@@ -54,7 +54,7 @@ fn parse_document_from_str(doc: &str) -> XmlResult<()> {
                 }
             }
             Event::Text(e) => {
-                criterion::black_box(e.unescape()?);
+                criterion::black_box(e.decode()?);
             }
             Event::CData(e) => {
                 criterion::black_box(e.into_inner());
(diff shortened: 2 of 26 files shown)


Mingun:
I finished work on the base part of the entities support. In this PR new `Event::GeneralRef` is added together with new `BytesRef` struct which represents the any `&...;` reference, including entity references and character references. Character references can be resolved by call `BytesRef::resolve_char_ref()`, entity references can be resolved by mapping from content of `BytesRef` to replacement text. Both usages are shown in the updated `custom_entities` example.

dralley:
I won't have a chance to fully review this for a couple of days.  Quick question though, am I correct in thinking that this PR will mean that any time a text block contains one or more entity references, instead of the developer receiving one `Event::Text`, they will receive a series of `Event::Text` and `Event::GeneralRef` which they will then need to merge back together themselves?

Mingun:
Yes, you are correct. But that does not mean that he/she will needed to construct the complete text themselves. In the next PR I plan to rename `Reader` to the `RawReader` and add make it return borrow-only `RawEvent`s, and in in the another PR introduce new `Reader` which will automatically merge all consequent `Text`, `CData` and `GeneralRef` events. This should be much more convenient for the average user. `RawReader` will only be needed for very fine control.

Because renames affects very many places, I want to do that in a separate PR to reduce noise in PR with new `Reader`.

Borrow-only reader-only events will be useful also in that sense that I plan to add `offset` member to them to track event position in the stream. When you construct event for writing you are obviously does not have position and I think it is better to not have a dummy value for it, in order to you couldn't mistakenly use writer event in the reading context and get the wrong position.

Because new `Reader` will have a stack of the `RawReader`s, it will be simple recreate readers when we will need to change encoding, so I think, that #158 is very close to resolving.

src/reader/mod.rs
```diff
@@ -250,6 +250,25 @@ macro_rules! read_event_impl {
...
+                        Ok((bytes, true)) => {
+                            $self.state.state = ParseState::InsideText;
+                            Ok(Event::GeneralRef(BytesRef::wrap(bytes, $self.decoder())))
+                        },
+                        Ok((_, false)) => {
+                            // We want to report error at `&`, but offset was increased,
+                            // so return it back (-1 for `&`)
+                            $self.state.last_error_offset = start - 1;
+                            Err(Error::Syntax(SyntaxError::UnclosedReference))
+                        }
```
Mingun, line 268: I found a code in ruffle in https://github.com/ruffle-rs/ruffle/pull/10471 from @Aaron1011 that will be break, because currently dangling `&` will always return `SyntaxError::UnclosedReference`. I think I should add a new configuration option for this here

Mingun:
@dralley, what do you think about this?

Changelog.md
```diff
@@ -24,6 +24,9 @@ XML specification. See the updated `custom_entities` example!
 
 - [#766]: Allow to parse resolved entities as XML fragments and stream events from them.
 - [#766]: Added new event `Event::GeneralRef` with content of [general entity].
+- [#766]: Added new configuration option `allow_dangling_amp` which allows to have
+  a `&` not followed by `;` in the textual data which is required for some applications
+  for compatibility reasons.
```
dralley, line 29: Which applications?

Mingun: I meant case from #719 here

dralley: Would it not be potentially better to just work with raw text events without attempting to unescape them?

dralley:
Sorry for not reviewing this promptly.  Once the current set of PRs are merged, can you rebase?

Mingun:
@dralley, you can review this. Initially I thought not to merge it until I implement other changes (in the follow-up PRs) that reworks the parser to reduce amount of releases with breaking changes, but because we already have enough amount of breaking changes, I think, it can be merged and included in next release.

examples/custom_entities.rs
```diff
@@ -1,76 +1,207 @@
...
 
-const DATA: &str = r#"
+use pretty_assertions::assert_eq;
 
-    <?xml version="1.0"?>
-    <!DOCTYPE test [
-    <!ENTITY msg "hello world" >
-    ]>
-    <test label="&msg;">&msg;</test>
+struct MyReader<'i> {
```
dralley, line 26: We very much need to make this functionality usable without the user needing to write or copy over a custom wrapper, IMO.  I believe that was the plan, I just want to make sure.

benches/macrobenches.rs
```diff
@@ -54,7 +54,7 @@ fn parse_document_from_str(doc: &str) -> XmlResult<()> {
                 }
             }
             Event::Text(e) => {
-                criterion::black_box(e.unescape()?);
+                criterion::black_box(e.decode()?);
```
dralley, line 57: About `Reader` is returning merged events - does that mean that the `.decode()` goes away completely as the entities would be pre-expanded?  And would there be any way to, say, return the original raw unexpanded XML between two tags from that wrapper?

examples/custom_entities.rs
dralley: My point just being that the current example as-is is a bit much to expect people to implement or copy and paste, and I just want to check my understanding that it's not a long-term solution.

Mingun: I have not really decided yet whether to leave this example as it is to demonstrate how the reader stack can be implemented if for some reason the standard solution does not work, or rewrite it to a new API. If it remains as it is, then there will be a mention of the standard way

benches/macrobenches.rs
Mingun: Yes, I think, `decode` method will gone

> And would there be any way to, say, return the original raw unexpanded XML between two tags from that wrapper, or would that be impossible without dropping to the `RawReader` level?

I think, it could be possible to implement that, but only if something requested that. I think, that this is niche feature. The API could be a special method that need to call instead of `read_event()` to get an unparsed content.

dralley (approved):
Looks reasonable.

>Initially I thought not to merge it until I implement other changes (in the follow-up PRs) that reworks the parser to reduce amount of releases with breaking changes, but because we already have enough amount of breaking changes, I think, it can be merged and included in next release.

I personally would prefer to see the `Reader` / `RawReader` split happen prior to the next release, or even in this PR in subsequent commits, to reduce the amount of massively breaking API changes going on.  I'd rather not split the changes across many releases.

From a purely psychological level I'd be deeply annoyed by having to change code that uses `Reader` to handle the existence of references throughout the returned event stream, and then subsequently needing to change it back later  to something closer to the original approach.  That particular type of churn (needing to make a change as a user and then un-make it later) is best avoided.

Mingun:
Totally agree. Ok, then I'll merge it when follow-up PR will be ready

dralley:
@Mingun The Reader / RawReader split is coming soon, then?

devrosch:
@Mingun It looks like 0.38.0 has been released with this breaking change but without the Reader / RawReader split. Can you estimate when the Reader / RawReader split is going to be released? I'm asking so I can better assess whether to spend time to update my code and stitch together events (text, entities, CDATA) myself now or just wait for a little bit and have the new Reader already cover this.

PR #771, opened 2024-06-26
Fixes in unescape routine

Mingun:
This PR fixes 2 errors that I found when working on a new way to represent entity references:
1. Currently, `quick_xml` does not allow too long numbers to represent character references. It, however, does not takes in account, that any number of leading zeroes are allowed, so the real character entity length can be any. [Conformance](https://github.com/belingueres/xml-conformance) test suite contains such tests.
2. #739 introduced regression where `Attribute::unescape_value` no longer unescaped predefined entities. That was fixed and checked that no similar places remains

[Corresponding test](https://dev.w3.org/cvsweb/2001/XML-Test-Suite/xmlconf/xmltest/valid/sa/042.xml?rev=1.1.1.1) (xmltest\valid\sa\042.xml -- one of):
```xml
<!DOCTYPE doc [
<!ELEMENT doc (#PCDATA)>
]>
<doc>&#00000000000000000000000000000000065;</doc>
```

### DIFF
--- a/src/escape.rs
+++ b/src/escape.rs
@@ -12,8 +12,8 @@ use pretty_assertions::assert_eq;
 pub enum EscapeError {
     /// Entity with Null character
     EntityWithNull(Range<usize>),
-    /// Unrecognized escape symbol
-    UnrecognizedSymbol(Range<usize>, String),
+    /// Referenced entity in unknown to the parser.
+    UnrecognizedEntity(Range<usize>, String),
     /// Cannot find `;` after `&`
     UnterminatedEntity(Range<usize>),
     /// Cannot convert Hexa to utf8
@@ -36,11 +36,9 @@ impl std::fmt::Display for EscapeError {
                 "Error while escaping character at range {:?}: Null character entity not allowed",
                 e
             ),
-            EscapeError::UnrecognizedSymbol(rge, res) => write!(
-                f,
-                "Error while escaping character at range {:?}: Unrecognized escape symbol: {:?}",
-                rge, res
-            ),
+            EscapeError::UnrecognizedEntity(rge, res) => {
+                write!(f, "at {:?}: unrecognized entity `{}`", rge, res)
+            }
             EscapeError::UnterminatedEntity(e) => write!(
                 f,
                 "Error while escaping character at range {:?}: Cannot find ';' after '&'",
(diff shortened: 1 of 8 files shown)


src/escape.rs
```diff
@@ -1799,3 +1799,17 @@ fn parse_number(bytes: &str, range: Range<usize>) -> Result<char, EscapeError> {
...
+#[inline]
+fn from_str_radix(src: &str, radix: u32) -> Result<u32, ParseIntError> {
+    let codepoint = u32::from_str_radix(src, radix)?;
+
+    // If we here then `src` is not empty, otherwise u32::from_str_radix returns error
+    match src.as_bytes()[0] {
+        // We should not allow sign numbers, but u32::from_str_radix will accept them.
+        // Because we cannot directly construct necessary error, we use the function
+        // which would return it for us
+        b'+' | b'-' => u32::from_str_radix("=", 10),
```
dralley, line 1812: I don't like this very much, personally.  I can see the appeal of using `from_str_radix` and `ParseIntError`, the overlap is pretty significant, but this is definitely a hack around the mismatch in semantics.  I'd rather define our own `ParseIntError` (or equivalent) that is easily convertible from the original and can be easily returned on its own.  You can copy the variants and descriptions straight from the original.

Mingun: After thinking I realized that here we needed our own type of error in any case (`ParseCharRefError`), into which I moved the variants that make sense only when parsing the character reference.

src/escape.rs
```diff
@@ -5,30 +5,56 @@ use std::borrow::Cow;
...
+    UnexpectedSign,
+    /// Number cannot be parsed due to non-number characters or a numeric overflow.
+    InvalidNumber(ParseIntError),
+    /// Character reference represents not a valid unicode codepoint.
+    InvalidCodepoint(u32),
+    /// Character reference expanded to a not permitted character for an XML.
+    ///
+    /// Currently, only `0x0` character produces this error.
+    IllegalCharacter(u32),
+}
```
dralley, line 21: :+1:

src/escape.rs
```diff
@@ -1791,18 +1816,27 @@ pub const fn resolve_html5_entity(entity: &str) -> Option<&'static str> {
...
+    }
+}
+
+#[inline]
+fn from_str_radix(src: &str, radix: u32) -> Result<u32, ParseCharRefError> {
+    match src.as_bytes().first().copied() {
+        // We should not allow sign numbers, but u32::from_str_radix will accept `+`.
+        // We also handle `-` to be consistent in returned errors
+        Some(b'+') | Some(b'-') => Err(ParseCharRefError::UnexpectedSign),
+        _ => u32::from_str_radix(src, radix).map_err(ParseCharRefError::InvalidNumber),
```
dralley, line 1840: :+1: 

Much better, thanks

dralley (approved):

PR #780, opened 2024-07-04
Allow attributes in the `Event::End` and fix `.error_position()`

### LINKED ISSUE #776: Parsing malformed closing tags
For [Ruffle](https://ruffle.rs/) it would be useful if we could parse these two kinds of malformed XML:
```xml
<a></a/>
```
From https://github.com/ruffle-rs/ruffle/issues/15674. Note the extra `/`.
```xml
<a></a hello="world">
```
From https://github.com/ruffle-rs/ruffle/issues/16862. Note the attribute in the closing tag!

Excuse me if I missed a config option for this.

Mingun:
This PR allows us to have attributes in the `Event::End` (but they are not accessible via `BytesEnd` directly). As explained in https://github.com/tafia/quick-xml/issues/776#issuecomment-2208078867 we anyway can produce an event which `.name()` does not valid, so it is better ar least keep it useful for some scenarios. When validation API would be added (I'm working on it) we can validate all events and report an error.

Closes #776

When changing parser I noticed, that in one arm we do not set `last_error_offset`. I checked if we really cover that lines -- yes, but we actually does not check anything. `last_error_offset` was checked for `0`, but this is its default value, so was unclear is that code running or not. I improved check and fix the bug that was found due to that.

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -589,7 +589,7 @@ impl EncodingRef {
 ///     // when the input is a &str or a &[u8], we don't actually need to use another
 ///     // buffer, we could directly call `reader.read_event()`
 ///     match reader.read_event_into(&mut buf) {
-///         Err(e) => panic!("Error at position {}: {:?}", reader.buffer_position(), e),
+///         Err(e) => panic!("Error at position {}: {:?}", reader.error_position(), e),
 ///         // exits the loop when reaching end of file
 ///         Ok(Event::Eof) => break,
 ///
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -8,6 +8,7 @@ use std::ops::Range;
 use crate::encoding::Decoder;
 use crate::errors::{Error, Result, SyntaxError};
 use crate::events::Event;
+use crate::parser::{ElementParser, Parser, PiParser};
 use crate::reader::state::ReaderState;
 
 /// A struct that holds a parser configuration.
@@ -449,15 +450,11 @@ macro_rules! read_to_end {
 #[cfg(feature = "async-tokio")]
 mod async_tokio;
 mod buffered_reader;
-mod element;
 mod ns_reader;
-mod pi;
 mod slice_reader;
 mod state;
 
-pub use element::ElementParser;
 pub use ns_reader::NsReader;
-pub use pi::PiParser;
 
 /// Range of input in bytes, that corresponds to some piece of XML
 pub type Span = Range<u64>;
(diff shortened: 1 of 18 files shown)


src/reader/mod.rs
```diff
@@ -346,18 +346,28 @@ macro_rules! read_until_close {
...
+            // in close tags. While such tags are not allowed by the specification,
+            // we anyway allow to parse them because:
+            // - we do not check constrains during parsing. This is performed by the
+            //   optional validate step which user should call manually
+            // - if we just look for `>` we will parse `</tag attr=">" >` as end tag
+            //   `</tag attr=">` and text `" >` which probably no one existing parser
+            //   do. This is still invalid tag but it also has no practical meaning
+            //   to parse it in a such way. End tag `</tag attr=">" >`, however, can
+            //   be produced by some real-world parsers, for example, by the parser
+            //   of the Macromedia Flash player.
```
dralley, line 358: For what it's worth, other "non-validating" parsers like expat still blow up on this

>ExpatError: not well-formed (invalid token): line 1, column 11

Lxml and libxml2 also do

>XMLSyntaxError: expected '>', line 1, column 12

>parserError: xmlParseDoc() failed

I'm on-board with making this possible to support, but our default behavior should probably be throwing an `IllFormedError` just like we do for mismatched tags (and like other parsers do).

>ExpatError: mismatched tag: line 1, column 7

Mingun: I want to evolve library in this direction:
- rename `Reader` to `RawReader` (and `Event` to `RawEvent`)
- make `RawReader` as liberal as possible. It should not make any validation checks
- add a `validate` method to `RawEvent` which will perform all well-formedless and validity checks (I already grab list of all checks from the [specification](https://www.w3.org/TR/xml11), 35 VC + 16 WFC references, actually most of them about DTD). All validation must be run manually by calling `RawEvent::validate`. Only one check cannot be performed in this way: check that [start tag name matches end tag name](https://www.w3.org/TR/xml11/#GIMatch)
- move existing checks (for example, [`IllFormedError::DoubleHyphenInComment`](https://github.com/tafia/quick-xml/blob/22b3e458d193e6937882ba6b643ec8c31a671de5/src/errors.rs#L119C26-L119C27)) to the new `validate` method
- introduce a new `Reader` which will perform validation checks when flag in config is set (and process many other things, like dealing with encodings and expanding entity references as shown in the `custom_entities` example in #766). Add necessary config flags

I haven't thought through all the details yet and don't want to focus on them yet. There will be a separate PR where all this will be considered. It should be pretty soon (this year).

So, given that we are not currently performing most validity checks and I am planning a generic consistent solution for them, do you agree that it is not worth introducing them now?

Mingun: > Also line 358 should be rewritten - XML isn't "produced" by a parser. You can just say that such documents exist in the wild and are tolerated by some parsers such as the one used by Adobe Flash.

Any help are appreciated :smiley: . I feel that my English is no good to writing simple-to-read explanations.

src/reader/mod.rs
```diff
@@ -345,18 +345,27 @@ macro_rules! read_until_close {
                 }
             },
             // `</` - closing tag
+            // #776: We parse using ElementParser which allows us to have attributes
+            // in close tags. While such tags are not allowed by the specification,
+            // we anyway allow to parse them because:
+            // - we do not check constrains during parsing. This is performed by the
```
dralley, line 351: spelling: constraints

src/reader/mod.rs
```diff
@@ -345,18 +345,27 @@ macro_rules! read_until_close {
...
+            //   optional validate step which user should call manually
+            // - if we just look for `>` we will parse `</tag attr=">" >` as end tag
+            //   `</tag attr=">` and text `" >` which probably no one existing parser
+            //   do. This is still invalid tag but it also has no practical meaning
```
dralley, line 355: do -> does

src/reader/mod.rs
dralley: That all sounds good to me, but we probably shouldn't document it as a "new feature" given that we do want this to eventually fail for users of the `Reader` API.

src/reader/mod.rs
Mingun: Fixed!

dralley (approved):

Mingun:
Task CIFuzz was failed by unknown reasons -- I asked https://github.com/google/oss-fuzz/issues/12168 but no response yet... It not related to this changes, so merge

Issue #230, opened 2020-08-19
Allow documents starting with <!doctype html> (UnexpectedBang)


Hocuri:
Currently, quick-xml returns UnexpectedBang if a document starts with `<!doctype html>`, which caused this issue: https://github.com/deltachat/deltachat-core-rust/issues/1804. The problem is that it will refuse to read the rest of the file then.

jbg:
Given that that's not valid XML, I would not expect an XML library to parse it. It seems like there still isn't a clear answer for whether this library intends to be an XML parser or something else, though.

tafia:
This library does intend to parse valid xmls only by default yes. Html parsing could be supported on a case by case and best effort basis.

jbg:
IMO, given that there already exist quite good crates for HTML parsing in Rust (e.g. html5ever), and given that HTML parsing is very different from XML parsing (despite them looking similar), it would simplify things a lot and probably lead to a higher quality library if HTML parsing was explicitly out of scope for quick-xml.

PR #781, opened 2024-07-05
Start CDATA section only after uppercase `<![CDATA[`

Mingun:
Implement more strict check for the start of CDATA section. Incorrect case is a syntax error. In that sense this PR is differs from #780 where attributes in the end tag is validation error. So we should not be case tolerant in this case.

### DIFF
--- a/src/reader/state.rs
+++ b/src/reader/state.rs
@@ -128,7 +128,11 @@ impl ReaderState {
                     self.decoder(),
                 )))
             }
-            BangType::CData if uncased_starts_with(buf, b"![CDATA[") => {
+            // XML requires uppercase only:
+            // https://www.w3.org/TR/xml11/#sec-cdata-sect
+            // Even HTML5 required uppercase only:
+            // https://html.spec.whatwg.org/multipage/parsing.html#markup-declaration-open-state
+            BangType::CData if buf.starts_with(b"![CDATA[") => {
                 debug_assert!(buf.ends_with(b"]]"));
                 Ok(Event::CData(BytesCData::wrap(
                     // Cut of `![CDATA[` and `]]` from start and end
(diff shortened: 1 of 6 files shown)


src/reader/state.rs
```diff
@@ -128,14 +128,22 @@ impl ReaderState {
...
+            // https://html.spec.whatwg.org/multipage/parsing.html#markup-declaration-open-state
+            BangType::CData if buf.starts_with(b"![CDATA[") => {
                 debug_assert!(buf.ends_with(b"]]"));
                 Ok(Event::CData(BytesCData::wrap(
                     // Cut of `![CDATA[` and `]]` from start and end
                     &buf[8..len - 2],
                     self.decoder(),
                 )))
             }
+            // XML requires uppercase only, but we will check that on validation stage:
```
dralley, line 143: Is there a particular reason to keep this difference in behavior intact in this PR?

dralley (approved):

src/reader/state.rs
Mingun: HTML allows using mixed case for `DOCTYPE` and handling of this was intentionally added to quick-xml in [0.23.0](https://github.com/tafia/quick-xml/blob/master/Changelog.md#0230----2022-05-08) and related #230. In one of previous PRs you said that you prefer to keep ability to parse HTML documents, so I leaved this ability.

CDATA sections, however, are always started from uppercase variant, event in HTML. Corresponding links a couple of lines below in the comment.

I planned to work on validation after merging #766 (because it changes possible content of `Text` events and we will know something about content, and may do not validate that).

After validation PR will be merged, the user will be able to validate each event manually after reading it if it requires strict parser and skip validation if that is not required. The API would be:
```rust
match reader.read_event() {
  Text(e) => {
    // Get iterator over all validation problems and stop at first
    // You also may skip problems that you want to allow
    if e.validate_iter().any() {
      return Err(...);
    }
    // Shortcut for reporting the first problem as a `Result`
    e.validate()?;
  }
  // do the same for other events
}
```

Issue #260, opened 2021-02-06
Allow streaming read of text content


ajtribick:
In some cases it may be useful to read the content of a text node via the `Read` trait. An example might be if the node contained a large amount of base64-encoded binary data, which the application parses into its own data structures: needing to go via a byte array holding the textual content of the node would needlessly allocate memory.

It would be useful to have a method on `Reader` that returns an implementation of `Read` that returns the unescaped/decoded content until the specified end tag.

In the case that the application does not read to the end of the returned `Read`, I think it makes sense for the `Reader` to skip ahead to where it would be if the `Read` had been completed.

Mingun:
I thought about it and here's what I suggest:
- Introduce `Event::TextChunk(BytesText)` and `Event::CDataChunk(BytesCData)`
- Those events will be emitted only for buffered reader, because in borrowed reader (when you use `read_event` method) the all XML is already in memory and there are no reasons to split text / CDATA into chunks
- When parsing text or CDATA return chunk event after one call to `BufRead::fill_buf()`, the chunk will contain as much data as the `BufRead` implementation returns
- When we find the end of event (start of markup / `]]>`), `Event::Text` / `Event::CData` will be emitted
- The application should start incremental process when it gets a first chunk event
- The application should collect / incrementally process chunk events until it gets a finish event

dralley:
That sounds reasonable

One point: returning `Event::TextChunk(BytesText)` would require maintaining a little extra state in the parser so that it remembers that it's still in the middle of parsing an `Event::Text`.  Therefore, if the parser already has to go into a special mode, we could theoretically just return an `Event::TextHandle` / `Event::CDataHandle` (note: no payload on those) immediately, after which we could probably let the user grab from the reader a handle that would allow you to `Read` the contents of the `Event::Text` / `Event::CData` directly, up until the start of the next event.  

That's a little closer to what the user suggested and might be more convenient.  There are downsides to both approaches though.

Mingun:
Implementation with handle could be difficult to implement. In theory we can return handle here instead of `match`:
https://github.com/tafia/quick-xml/blob/8d38e4cc145b1911d67690a9e076c056ff2e6efe/src/reader/mod.rs#L282-L292

but then in order to continue parsing we should guarantee that text from handle will be consumed. We could implement `Drop` for handle which will do that, but then we would need to suppress all errors that could occur during read. Implementation with special events seems easy to implement. The `Read` stream will also could be possible to implement that way:
```rust
[...]

PR #783, opened 2024-07-12
Implement ability to read data directly from the underlying reader

### LINKED ISSUE #623: Support for embedded raw binary

Mingun:
Closes #623

This also implements #260, but the user should manually:
- ensure, that it is reading text
- stop reading when text ends

So I do not consider that #260 is solved.

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -567,6 +567,68 @@ impl EncodingRef {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+/// A direct stream to the underlying [`Reader`]s reader which updates
+/// [`Reader::buffer_position()`] when read from it.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless read or polled"]
+pub struct BinaryStream<'r, R> {
+    inner: &'r mut R,
+    offset: &'r mut u64,
+}
+
+impl<'r, R> BinaryStream<'r, R> {
+    /// Returns current position in bytes in the original source.
+    #[inline]
+    pub const fn offset(&self) -> u64 {
+        *self.offset
+    }
+
+    /// Gets a reference to the underlying reader.
+    #[inline]
+    pub const fn get_ref(&self) -> &R {
+        self.inner
+    }
+
+    /// Gets a mutable reference to the underlying reader.
+    ///
+    /// Avoid read from this reader because this will not update reader's position
+    /// and will lead to incorrect positions of errors. Read from this stream instead.
+    #[inline]
+    pub fn get_mut(&mut self) -> &mut R {
+        self.inner
+    }
+}
+
+impl<'r, R> io::Read for BinaryStream<'r, R>
+where
+    R: io::Read,
+{
+    #[inline]
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        let amt = self.inner.read(buf)?;
+        *self.offset += amt as u64;
+        Ok(amt)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/reader/mod.rs
```diff
@@ -759,6 +818,60 @@ impl<R> Reader<R> {
...
         self.state.decoder()
     }
+
+    /// Get the direct access to the underlying reader, but tracks the amount of
+    /// read data and update [`Reader::buffer_position()`] accordingly.
+    ///
+    /// # Example
+    ///
+    /// This example demonstrates, how it is possible to read embedded binary data.
+    /// Such XML documents are exist in the wild.
```
dralley, line 828: This example demonstrates how to read stream raw bytes from an XML document.  This could be used to implement streaming read of text, or to read raw binary bytes embedded in an XML document.  (Documents with embedded raw bytes are not valid XML, but XML-derived file formats exist where such documents are valid).

src/reader/mod.rs
```diff
@@ -759,6 +818,60 @@ impl<R> Reader<R> {
...
+    ///     (Event::Start(BytesStart::new("tag")), 5)
+    /// );
+    ///
+    /// // Reading directly from underlying reader will not update position
+    /// // let mut inner = reader.get_mut();
+    ///
+    /// // Reading from the stream() advances position
+    /// let mut inner = reader.stream();
+    ///
+    /// // Read binary data. We somehow should known its size
```
dralley, line 852: We must know its size

dralley: Also somewhere (maybe up above, not here), we should mention the two conditions that the user must uphold when using this API and the potential consequences if it's misused.

Mingun: Mmm... what two conditions? I added a note that read data will not be returned in subsequent `Event`s

dralley: > the user should manually:
> 
> -  ensure, that it is reading text (binary data)
> -  stop reading when text (binary data) ends
>

Mingun: @dralley, what do you think about my last comment? If you think, that we can write something clear do not hesitate to write it -- I decidedly have no thoughts. I mean, that just "when binary data ends" is not clear enough, because this literally can be any condition which is already clear.

dralley (approved):

Issue #785, opened 2024-07-15
serde De/Serializing Option<u64> fails


jonaspleyer:
Hi,

I have some code in which I need to (de)-serialize in the following way:
```rust
#[test]
fn issue785() {
    #[derive(Debug, Deserialize, Eq, PartialEq, Serialize)]
    struct Identifier {
        parent: Option<u64>,
    }
    let ident = Identifier {
        parent: None,
    };
    let mut save_string = String::new();
    let mut serializer = quick_xml::se::Serializer::new(&mut save_string);
    serializer.indent(' ', 4);
    ident.serialize(serializer).unwrap();
    let res: Identifier = quick_xml::de::from_str(&save_string).unwrap();
    assert_eq!(res, ident);
}
```

This test does however fail with the current implementation of quick-xml. I have narrowed it down to the `deserialize_option` function of the `MapValueDeserializer` struct in `src/de/map.rs`. Sadly, I do not understand enough about the internals of `quick-xml` to contribute anything meaningful. Please let know how I can further help if needed.

PS: I have created a PR https://github.com/tafia/quick-xml/pull/786 to work on this problem.

Mingun:
Duplicate of #497

PR #786, opened 2024-07-15
add new test for issue 785

jonaspleyer:
This PR is meant to work in issue https://github.com/tafia/quick-xml/issues/785. Please let know how I can be of further help.

### DIFF
--- a/tests/serde-issues.rs
+++ b/tests/serde-issues.rs
@@ -494,3 +494,20 @@ fn issue683() {
         }
     );
 }
+
+#[test]
+fn issue785() {
+    #[derive(Debug, Deserialize, Eq, PartialEq, Serialize)]
+    struct Identifier {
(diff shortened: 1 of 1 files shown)


Mingun:
That case [a quite hard to resolve](https://github.com/tafia/quick-xml/issues/497#issuecomment-1321220134) and maybe everything that we can do it is document how each possible behaviour could be achieved using serde attributes.

jonaspleyer:
Thanks for the fast feedback. I did not see this issue. Since I do not strictly rely on xml (de)-serialization, I am considering to simply drop this functionality. If it helps you with further development: I am mostly concerned with obtaining identical results via round-trips: Rust -> xml -> Rust

dralley:
The backing issue is closed as a duplicate of another issue which is now closed - closing this too.

PR #788, opened 2024-07-21
Serde support for serializing and deserializing binary blobs in XML files

elrnv:
This is another attempt to support binary blobs in serde, needed for [vtkio](https://github.com/elrnv/vtkio).

This attempts to address #623 
While initial support was added #783 this adds it for the serde api.

This PR also adjusts how trimming is handled in the serde API since this is important when dealing with binary. I tried to be consistent with the event based Reader/Writer APIs.

Please inspect the test failures to evaluate if this makes sense.


src/de/mod.rs
```diff
@@ -2217,7 +2244,16 @@ impl<'i, R: XmlRead<'i>, E: EntityResolver> XmlReader<'i, R, E> {
...
+                    match e
+                        .unescape_with(|entity| self.entity_resolver.resolve(entity))
+                        .map(|res| self.drain_text(res))
+                    {
+                        Ok(x) => x,
+                        // failed to escape treat as binary blob.
+                        Err(_) => Ok(DeEvent::Binary(Binary {
+                            text: e.into_inner(),
+                        })),
+                    }
```
Mingun: We definitely shouldn't rely on luck here. Binary should be explicitly requested for the field via flag in field name

elrnv: What is the best way/mechanism to maintain context in the code to keep track of flags like that?

dralley:
If someone wants to try to resurrect this, we're probably open to it so long as it doesn't require serious hacks.  But I'm closing this because it doesn't seem like it's going anywhere.

PR #796, opened 2024-08-21
unpin tokio Cargo.toml

j-mendez:


Mingun:
We cannot do that because quick-xml uses API that was introduced only since 1.10, and in tests we uses API that was introduced since 1.21.

Why you want to do that?

j-mendez:
> We cannot do that because quick-xml uses API that was introduced only since 1.10, and in tests we use API that was introduced since 1.21.
> 
> 
> 
> Why you want to do that?

No worries - was testing some minor things for v0

PR #797, opened 2024-08-23
fix namespace prefix deserialize

### LINKED ISSUE #591: problems with colon separators
```rust
fn main() {
    let xml = r#"<p:sldLayoutIdLst>
        <p:sldLayoutId id="2147483649" r:id="rId1"/>
        <p:sldLayoutId id="2147483650" r:id="rId2"/>
    </p:sldLayoutIdLst>"#;
    let result:CtSlideLayoutIdList = quick_xml::de::from_str(xml).unwrap();
    println!("{:?}",quick_xml::se::to_string(&result));
}

#[derive(Serialize,Deserialize,Debug)]
#[serde(rename(serialize = "p:sldLayoutIdLst", deserialize = "sldLayoutIdLst"))]
pub struct CtSlideLayoutIdList {
    #[serde(rename(serialize = "p:sldLayoutId", deserialize = "sldLayoutId"))]
    #[serde(skip_serializing_if = "Option::is_none")]
    pub sld_layout_id: Option<Vec<CtSlideLayoutIdListEntry>>,
}



#[derive(Serialize,Deserialize,Debug)]
pub struct CtSlideLayoutIdListEntry{
    #[serde(rename="@id")]
    pub id_attr:Option<String>,

    #[serde(rename="@r:id")]
    pub r_id_attr:Option<String>,
}
```
i got an error "thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("duplicate field `@id`")'"

KaiserY:
Similar to #539, namespace prefix (e.g. r:id) also map to `@r:id` instead of `@id`, possibly fix #591.

### DIFF
--- a/src/de/key.rs
+++ b/src/de/key.rs
@@ -41,7 +41,7 @@ fn decode_name<'n>(name: QName<'n>, decoder: Decoder) -> Result<Cow<'n, str>, De
 ///   to the identifier
 /// - put the decoded [`local_name()`] of a name to the identifier
 ///
-/// The final identifier looks like `[@]local_name`, or `@xmlns`, or `@xmlns:binding`
+/// The final identifier looks like `[@]local_name`, or `@xmlns`, or `@xmlns:binding`, or `@r:id`
 /// (where `[]` means optional element).
 ///
 /// The deserializer also supports deserializing names as other primitive types:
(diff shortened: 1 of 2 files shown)


Mingun:
The problem with that approach that in the real document prefix not necessary will be called `r`. This is different from the `xmlns` prefix, which is a special prefix that is always defined, cannot be redefined and its namespace cannot be assigned to another prefix. In other words, all of that are prohibited:
- `xmlns:xmlns="something-not-xmlns-namespace"`
- `xmlns:xmlns="http://www.w3.org/2000/xmlns/"`
- `xmlns:any-name="http://www.w3.org/2000/xmlns/"`

In the end that means that the only way to declare namespace prefix is to use `xmlns:prefix-name="namespace"`. `xmlns` is constant. That is why we have an exception for it, although you generally should not read namespace prefixes into the fields, because `prefix-name` can be arbitrary XML identifier and rely on the particular name for it is dangerous.

The right approach is to include namespace name to the `#[serde(rename)]` and during deserialization resolve actual prefixes into namespaces, compose a field name from them (for example, `@attr{namespace}`) and pass that name to the type's `Visitor`. That already was explored in #466 and we should move forward that way.

KaiserY:
Thanks for the full explanation. in my case i'm parsing ooxml, somehow the prefix seem fixed, so i think this will be a quick fix. I will dig more into #466, close this for now.

PR #802, opened 2024-09-19
Fix DocType closing bracket recognition in buffered reader

### LINKED ISSUE #533: Deserialization of a doctype with very long content fails
`quick_xml::de::from_reader()` parsing fails if the XML contains a doctype with content larger than the internal `BufRead` capacity.
For instance
```xml
<!DOCTYPE [
<!-- A very very long comment *snipped* -->
]>
<X></X>
```

Here is a minimal code to reproduce this issue. It fails with an `ExpectedStart` error.
```rust
use std::io::Write;
use serde::Deserialize;

#[derive(Deserialize)]
struct X {}

fn main() {
    {
        let mut file = std::fs::File::create("test.xml").unwrap();
        let header = &"<!DOCTYPE X [<!--";
        let footer = &"-->]><X></X>";
        let padding = 8192 - (header.len() + 2);
        write!(file, "{header}{:1$}{footer}", "", padding).unwrap();
    }

    let file = std::fs::File::open("test.xml").unwrap();
    let reader = std::io::BufReader::new(file);
    let _: X = quick_xml::de::from_reader(reader).unwrap();
}
```
`Cargo.toml` content
```toml
[package]
name = "test"
version = "0.1.0"
edition = "2021"

[dependencies]
quick-xml = { version = "0.27.1", features = ["serialize"] }
serde = { version = "1.0", features = ["derive"] }
```

* When decreasing the padding size, or using `BufReader::with_capacity()` to increase the buffer, even of 1 byte, there is no error.
* Other `BufRead` implementations don't have this issue (checked with `&[u8]` and `stdin`).
* Content does not have to be in one "block". The same issue occurs for a doctype split into multiple declarations and comments.
* With a longer doctype with real content, the error may be different. For instance it may complains about an invalid `!` from a `!ENTITY` tag.
* No issue with `serde-xml-rs`, even for larger comments.
* Tested on Windows, with rustc 1.66.0.

### LINKED ISSUE #590: Unexpected Bang Since 0.23.1

### LINKED ISSUE #801: Error when parsing !DOCTYPE longer than buffer of BufRead
## Description
When parsing `<!DOCTYPE []>` from a BufRead, there is a bug where it thinks a closing tag `'>'` that occurs as part of the doctype instead closes the DOCTYPE. This occurs when the entire doctype does not fit inside the bufread's buffer.

Because of that, when a closing tag '>' is mistakenly parsed as end of doctype, the content after that is parsed incorrectly.

## Steps to reproduce
```rust
use quick_xml::events::Event;
use quick_xml::Reader;
use std::io::BufReader;

fn main() {
    let xml = r#"<!DOCTYPE X [<!-- comment --><!ENTITY a "a">]>"#;
    let reader = BufReader::with_capacity(4, xml.as_bytes());
    let mut xml_reader = Reader::from_reader(reader);
    let mut buf = vec![];
    loop {
        match xml_reader.read_event_into(&mut buf).unwrap() {
            Event::Eof => break,
            _ => {}
        }
    }
    println!("Parsed xml without error");
}
```

Above code throws an error `Syntax(InvalidBangMarkup)`. This is because the closing tag for comment `<!-- comment -->` is mistakenly parsed as closing doctype, and so trying to parse `<!ENTITY>` throws an error.

However, things work fine if we increase the bufreader capacity to greater than text content (e.g. 1024).

I'll be opening a PR for the fix right away

BlueGreenMagick:
Fixes #533, fixes #590, fixes #801

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -1019,8 +1019,8 @@ enum BangType {
     CData,
     /// <!--...-->
     Comment,
-    /// <!DOCTYPE...>
-    DocType,
+    /// <!DOCTYPE...>. Contains balance of '<' (+1) and '>' (-1)
+    DocType(i32),
 }
 impl BangType {
     #[inline(always)]
@@ -1028,7 +1028,7 @@ impl BangType {
         Ok(match byte {
             Some(b'[') => Self::CData,
             Some(b'-') => Self::Comment,
-            Some(b'D') | Some(b'd') => Self::DocType,
+            Some(b'D') | Some(b'd') => Self::DocType(0),
             _ => return Err(Error::Syntax(SyntaxError::InvalidBangMarkup)),
         })
     }
@@ -1040,7 +1040,7 @@ impl BangType {
     /// - `buf`: buffer with data consumed on previous iterations
     /// - `chunk`: data read on current iteration and not yet consumed from reader
     #[inline(always)]
-    fn parse<'b>(&self, buf: &[u8], chunk: &'b [u8]) -> Option<(&'b [u8], usize)> {
+    fn parse<'b>(&mut self, buf: &[u8], chunk: &'b [u8]) -> Option<(&'b [u8], usize)> {
         match self {
             Self::Comment => {
                 for i in memchr::memchr_iter(b'>', chunk) {
@@ -1085,14 +1085,15 @@ impl BangType {
                     }
                 }
             }
-            Self::DocType => {
-                for i in memchr::memchr_iter(b'>', chunk) {
-                    let content = &chunk[..i];
-                    let balance = memchr::memchr2_iter(b'<', b'>', content)
-                        .map(|p| if content[p] == b'<' { 1i32 } else { -1 })
-                        .sum::<i32>();
-                    if balance == 0 {
-                        return Some((content, i + 1)); // +1 for `>`
+            Self::DocType(ref mut balance) => {
+                for i in memchr::memchr2_iter(b'<', b'>', chunk) {
+                    if chunk[i] == b'<' {
+                        *balance += 1;
+                    } else {
+                        if *balance == 0 {
+                            return Some((&chunk[..i], i + 1)); // +1 for `>`
+                        }
+                        *balance -= 1;
                     }
                 }
             }
@@ -1104,7 +1105,7 @@ impl BangType {
         match self {
             Self::CData => Error::Syntax(SyntaxError::UnclosedCData),
             Self::Comment => Error::Syntax(SyntaxError::UnclosedComment),
-            Self::DocType => Error::Syntax(SyntaxError::UnclosedDoctype),
+            Self::DocType(_) => Error::Syntax(SyntaxError::UnclosedDoctype),
         }
     }
 }
@@ -1414,7 +1415,7 @@ mod test {
                                 .unwrap();
                             assert_eq!(
                                 (ty, Bytes(bytes)),
-                                (BangType::DocType, Bytes(b"!DOCTYPE"))
+                                (BangType::DocType(0), Bytes(b"!DOCTYPE"))
                             );
                             assert_eq!(position, 10);
                         }
@@ -1488,7 +1489,7 @@ mod test {
                                 .unwrap();
                             assert_eq!(
                                 (ty, Bytes(bytes)),
-                                (BangType::DocType, Bytes(b"!doctype"))
+                                (BangType::DocType(0), Bytes(b"!doctype"))
                             );
                             assert_eq!(position, 10);
                         }
--- a/tests/issues.rs
+++ b/tests/issues.rs
@@ -171,6 +171,24 @@ mod issue514 {
     }
 }
 
+/// Regression test for https://github.com/tafia/quick-xml/issues/590
+#[test]
+fn issue590() {
+    let mut reader = Reader::from_reader(BufReader::with_capacity(
+        16,
+        &b"<!DOCTYPE t [<!1><!2>]>"[..],
+        // 0      7       ^15    ^23
+        //[                ] = capacity
+    ));
+    let mut buf = Vec::new();
+    loop {
+        match reader.read_event_into(&mut buf).unwrap() {
+            Event::Eof => break,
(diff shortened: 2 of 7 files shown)


BlueGreenMagick:
Modified the code so that `balance_buf` is computed only once.

Mingun (changes requested):
That PR will not fix the problem entirely, because correct DTD can contain unbalanced number of `<` and `>`, but I think that is rare case, so until proper fix will be implemented, a partial one also acceptable. The implementation, however, still may be improved. The problem here that we do not store state between invocations of `BangType::parse`. So the solution is to store balance inside `BangType::DocType`, then, I think, it will be unnecessary to calculate it over `buf` (because everything in `buf` previously was in `chunk`).

I also see the problem in the external `for i in memchr::memchr_iter(b'>', chunk)` loop. It will skip `<`s inside `chunk`s that does not contain `>`, so the balance will be calculated incorrectly. I think, if create a test carefully, it will catch such situation.

So I ask you to do the following changes:
- convert `BangType::DocType` -> `BangType::DocType(i32)`
- start with balance == 1 (it is `<` in `<!DOCTYPE`) here
https://github.com/tafia/quick-xml/blob/2f3824a1d265ac0b683aa485683ad00ddbc5c092/src/reader/mod.rs#L1031
- when calculate balance take into account saved balance
- emit event only when `BangType::DocType(0)` here
https://github.com/tafia/quick-xml/blob/2f3824a1d265ac0b683aa485683ad00ddbc5c092/src/reader/state.rs#L147
- add tests for the mentioned case (chunk with `<` + chunk with `>` which closes `<` from the first chunk). It is enough just craft such a string and use appropriate buffer size to slice string to desired chunks
- also add regression test from https://github.com/tafia/quick-xml/issues/590#issuecomment-1503927845
- try to check if this fixes #533. I think it should because it seems to be a duplicate of these issues, but need to check

BlueGreenMagick:
Thanks for the detailed feedback! I adjusted the PR.

> * start with balance == 1 (it is `<` in `<!DOCTYPE`) here
As the balance is calculated in up to chunk excluding current `>`, 1 is not added to balance initially.

> add tests for the mentioned case (chunk with < + chunk with > which closes < from the first chunk). It is enough just craft such a string and use appropriate buffer size to slice string to desired chunks

Are regression tests for issues 533, 590, 801 enough? Or are additional tests desired?

tests/issues.rs
```diff
@@ -171,6 +171,40 @@ mod issue514 {
...
+    #[derive(Deserialize)]
+    struct X {}
+
+    #[test]
+    fn test() {
+        let xml = "<!DOCTYPE X [<!-- -->]><X></X>";
+        let reader = BufReader::with_capacity(20, xml.as_bytes());
+        let _: X = quick_xml::de::from_reader(reader).unwrap();
+    }
+}
```
Mingun, line 188: Please put this test into `serde-issues.rs`. It compiles only when `serialize` feature is active so does not require feature gate.

src/reader/mod.rs
```diff
@@ -1020,15 +1020,15 @@ enum BangType {
     /// <!--...-->
     Comment,
     /// <!DOCTYPE...>
-    DocType,
+    DocType(i32),
```
Mingun, line 1023: Updating documentation also would be useful (language can be improved if required, I'm not native speaker):
```rust
    /// <!DOCTYPE...>. Contains a balance of `<` (+1) and `>` (-1)
    DocType(i32),
```

Mingun (approved):
> Are regression tests for issues 533, 590, 801 enough? Or are additional tests desired?

I'm not sure that tests still cover the case when we get four chunks with roughly such content:
1. `<!DOCTYPE`
2. `<` without any `>`
3. `>`, which would close `<` from point 2, not the doctype
4. `>` which should close doctype

If that case already covered, fine, although I would prefer to have test which explicitly guaranties such conditions (can tweak one of added regression tests for that).

And please add changelog entry to `Changelog.md` that people know that bug (#590) fixed. I'll close other as duplicate.

Mingun:
You also can squash all your changes if you wish

BlueGreenMagick:
Done! 

I adjusted regression test for issue801 so that all angle brackets are explicitly in different buffer (by lowering buffer size to 2 bytes).

I think it might be better (and easier with GitHub UI) to squash merge on your end, so others can follow the conversations and commits in the future.

Mingun:
Thanks!

Issue #227, opened 2020-07-31
quick_xml::se::to_string returns a DeError


ghost:
Name of the error doesn't fit, I would expect it to return something like SeError.

tafia:
Yes. Mind doing a PR?

ozgrakkurt:
yes, I can make a PR

dralley:
@ozgrakkurt I can't find any PR for this, have you opened one?

PR #810, opened 2024-09-28
Return `std::io::Error` from Writer methods

RedPhoenixQ:
I recently made a PR to update this crate in [another crate (inferno)](https://github.com/jonhoo/inferno/pull/332) and needed to remove this crates errors from the public API. While doing this I realized that the `Writer` could never return any error other than `Error::Io`. Wrapping this error in `Arc` and putting it in the `Error` enum feels unnecessary.

This does change the public API of `Writer` and introduces the `DeError::Io` kind which may not be desirable. In case this change will not happen a note in the documentation of `Writer` about not returning errors other than `Error::Io` may be useful (but may be hard to enforce, leading to outdated/incorrect docs).

I'd be more than willing to make any other changes if needed.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -159,7 +159,7 @@ impl std::error::Error for IllFormedError {}
 /// The error type used by this crate.
 #[derive(Clone, Debug)]
 pub enum Error {
-    /// XML document cannot be read from or written to underlying source.
+    /// XML document cannot be read from underlying source.
     ///
     /// Contains the reference-counted I/O error to make the error type `Clone`able.
     Io(Arc<IoError>),
@@ -318,8 +318,6 @@ pub mod serialize {
     pub enum DeError {
         /// Serde custom error
         Custom(String),
-        /// IO error from Writer
-        Io(Arc<IoError>),
         /// Xml parsing error
         InvalidXml(Error),
         /// Cannot parse to integer
@@ -351,13 +349,6 @@ pub mod serialize {
         /// store at current position, for example, attempt to deserialize `struct`
         /// from attribute or attempt to deserialize binary data.
         ///
-        /// Serialized type cannot be represented in an XML due to violation of the
-        /// XML rules in the final XML document. For example, attempt to serialize
-        /// a `HashMap<{integer}, ...>` would cause this error because [XML name]
-        /// cannot start from a digit or a hyphen (minus sign). The same result
-        /// would occur if map key is a complex type that cannot be serialized as
-        /// a primitive type (i.e. string, char, bool, unit struct or unit variant).
-        ///
         /// [XML name]: https://www.w3.org/TR/xml11/#sec-common-syn
         Unsupported(Cow<'static, str>),
         /// Too many events were skipped while deserializing a sequence, event limit
@@ -370,7 +361,6 @@ pub mod serialize {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             match self {
                 DeError::Custom(s) => write!(f, "{}", s),
-                DeError::Io(e) => write!(f, "{}", e),
                 DeError::InvalidXml(e) => write!(f, "{}", e),
                 DeError::InvalidInt(e) => write!(f, "{}", e),
                 DeError::InvalidFloat(e) => write!(f, "{}", e),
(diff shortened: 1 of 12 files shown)


Mingun:
I think, it would be better to introduce a new error for serializer and close #227. Having variant in <tt><i>De</i>Error</tt> just for _writer_ looks completely disheartening to me.

RedPhoenixQ:
> I think, it would be better to introduce a new error for serializer and close #227. Having variant in DeError just for _writer_ looks completely disheartening to me.

I agree and did think of making such a big change. Have added changes to split out `SeError` from `DeError`. While doing this I also see an opportunity to remove `errors::Error` from `DeError` aswell by also spitting out a `DecodeError` (from `encoding::Decoder` impl) and `ReadError` (from `reader::read_event_impl`) that could be included separately. This would make it so that no error type has more variants than needed. @Mingun Do you think this is too much for this one PR or wanted at all?

> Also, please add a changelog entry (do not forgot to add explicit link to section with links).

I will be sure to add this when the scope of these changes are finalized

src/errors.rs
```diff
@@ -458,10 +438,75 @@ pub mod serialize {
...
+    }
+
+    impl fmt::Display for SeError {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            match self {
+                SeError::Custom(s) => write!(f, "{}", s),
+                SeError::Io(e) => write!(f, "I/O error: {}", e),
+                SeError::Fmt(e) => write!(f, "Formatting error: {}", e),
+                SeError::Unsupported(s) => write!(f, "Unsupported value: {}", s),
+                SeError::NonEncodable(e) => write!(f, "Malformed UTF-8: {}", e),
```
Mingun, line 472: For better composition of different types of errors, it is better to start the error texts with a small letter. In general, we need to correct the texts for other errors too
```suggestion
                SeError::Fmt(e) => write!(f, "formatting error: {}", e),
                SeError::Unsupported(s) => write!(f, "unsupported value: {}", s),
                SeError::NonEncodable(e) => write!(f, "malformed UTF-8: {}", e),
```

src/se/key.rs
```diff
@@ -1,4 +1,4 @@
-use crate::errors::serialize::DeError;
+use crate::errors::serialize::SeError;
```
Mingun, line 1: In all other places, you used this path to error. Let's here too, for consistency:
```suggestion
use crate::se::SeError;
```

tests/serde-se.rs
```diff
@@ -1,7 +1,7 @@
 use quick_xml::de::from_str;
 use quick_xml::se::Serializer;
 use quick_xml::utils::Bytes;
-use quick_xml::DeError;
+use quick_xml::SeError;
```
Mingun, line 4: For consistency, this also could be imported as `quick_xml::se::SeError` in line 2

src/writer.rs
```diff
@@ -280,8 +278,8 @@ impl<W: Write> Writer<W> {
     /// # use serde::Serialize;
     /// # use quick_xml::events::{BytesStart, Event};
     /// # use quick_xml::writer::Writer;
-    /// # use quick_xml::DeError;
-    /// # fn main() -> Result<(), DeError> {
+    /// # use quick_xml::SeError;
```
Mingun, line 281: For consistency
```suggestion
    /// # use quick_xml::se::SeError;
```

Mingun (approved):
Great work. Only several minor style corrections are preferred and a changelog entry.

RedPhoenixQ:
I have work on the changes to narrow other return values, like from the `encoding` module. I'll submit a separate PR with these changes if that would be of interest

Mingun:
Unrelated changes in changelog.md is unwanted. Could you revert them? Especially two blank lines used to separate releases.

> I have worked on some changes to narrow other return values, like from the encoding module. I'll submit a separate PR after this has landed with these changes if that would be of interest

Any PR is welcome

RedPhoenixQ:
> Unrelated changes in changelog.md is unwanted. Could you revert them? Especially two blank lines used to separate releases.

My bad, didn't realize some formatter had kicked in. Fixed

Mingun:
Thanks!

Mingun:
@RedPhoenixQ, for the reference: I will wait for your other PR for a some time (to the end of next week) before release a new version.

PR #811, opened 2024-09-29
Make Errors more "narrow" 

RedPhoenixQ:
This was discussed in #810 [here](https://github.com/tafia/quick-xml/issues/810#issuecomment-2381341534).

This splits up error types so that there is almost one type for each module, which narrows the amount of error variants that is returned from each function.

For the functions where two or more error types may be returned the new `combined_error!` macro is used to create a new error `enum` that holds these variants with most error `impl`'s automatically done. 

The old `errors::Error` is still present and public. All other new errors implement `From<_> for errors::Error` so that the provided `errors::Result` type can still be used to try (`?`) any function from this crate. 

I haven't spent much time adding/editing docs for these changes since most of them still apply without changes. Maybe the docs for `errors::Error` should be changed to make it clear that it will not be given out as an error from anywhere directly. There are also some names of the new error types that might need changing

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -358,16 +358,16 @@ pub mod serialize {
                 DeError::InvalidXml(e) => write!(f, "{}", e),
                 DeError::InvalidInt(e) => write!(f, "{}", e),
                 DeError::InvalidFloat(e) => write!(f, "{}", e),
-                DeError::InvalidBoolean(v) => write!(f, "Invalid boolean value '{}'", v),
-                DeError::KeyNotRead => write!(f, "Invalid `Deserialize` implementation: `MapAccess::next_value[_seed]` was called before `MapAccess::next_key[_seed]`"),
+                DeError::InvalidBoolean(v) => write!(f, "invalid boolean value '{}'", v),
+                DeError::KeyNotRead => write!(f, "invalid `Deserialize` implementation: `MapAccess::next_value[_seed]` was called before `MapAccess::next_key[_seed]`"),
                 DeError::UnexpectedStart(e) => {
-                    f.write_str("Unexpected `Event::Start(")?;
+                    f.write_str("unexpected `Event::Start(")?;
                     write_byte_string(f, e)?;
                     f.write_str(")`")
                 }
-                DeError::UnexpectedEof => write!(f, "Unexpected `Event::Eof`"),
+                DeError::UnexpectedEof => write!(f, "unexpected `Event::Eof`"),
                 #[cfg(feature = "overlapped-lists")]
-                DeError::TooManyEvents(s) => write!(f, "Deserializer buffers {} events, limit exceeded", s),
+                DeError::TooManyEvents(s) => write!(f, "deserializer buffered {} events, limit exceeded", s),
             }
         }
     }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,13 +1,11 @@
 //! Error management module
 
-use crate::encoding::Decoder;
+use crate::encoding::{Decoder, EncodingError};
 use crate::escape::EscapeError;
 use crate::events::attributes::AttrError;
 use crate::name::{NamespaceError, QName};
 use std::fmt;
 use std::io::Error as IoError;
-use std::str::Utf8Error;
-use std::string::FromUtf8Error;
 use std::sync::Arc;
 
 /// An error returned if parsed document does not correspond to the XML grammar,
@@ -165,13 +163,10 @@ pub enum Error {
     Syntax(SyntaxError),
     /// The document is not [well-formed](https://www.w3.org/TR/xml11/#dt-wellformed).
     IllFormed(IllFormedError),
-    /// Input decoding error. If [`encoding`] feature is disabled, contains `None`,
-    /// otherwise contains the UTF-8 decoding error
-    ///
-    /// [`encoding`]: index.html#encoding
-    NonDecodable(Option<Utf8Error>),
     /// Attribute parsing error
     InvalidAttr(AttrError),
+    /// Encoding error
+    Encoding(EncodingError),
     /// Escape error
     EscapeError(EscapeError),
     /// Parsed XML has some namespace-related problems
@@ -211,19 +206,11 @@ impl From<IllFormedError> for Error {
     }
 }
 
-impl From<Utf8Error> for Error {
-    /// Creates a new `Error::NonDecodable` from the given error
+impl From<EncodingError> for Error {
+    /// Creates a new `Error::EncodingError` from the given error
     #[inline]
-    fn from(error: Utf8Error) -> Error {
-        Self::NonDecodable(Some(error))
-    }
-}
-
-impl From<FromUtf8Error> for Error {
-    /// Creates a new `Error::Utf8` from the given error
-    #[inline]
-    fn from(error: FromUtf8Error) -> Error {
-        error.utf8_error().into()
+    fn from(error: EncodingError) -> Error {
+        Self::Encoding(error)
     }
 }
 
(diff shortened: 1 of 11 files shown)


RedPhoenixQ:
Sorry for the force pushes, it keeps running my usage example as a doc test and the annotations (like `'''rust`) mean different things in different rust version apparently

src/name.rs
```diff
@@ -3,13 +3,58 @@
...
+        /// Namespace to which prefix tried to be bound
+        namespace: Vec<u8>,
+    },
+}
+
+impl fmt::Display for NamespaceError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::UnknownPrefix(prefix) => {
+                f.write_str("Unknown namespace prefix '")?;
```
Mingun, line 39: As already mentioned, if we bring order in errors, make all texts with a small letter:
```suggestion
                f.write_str("unknown namespace prefix '")?;
```

src/name.rs
```diff
@@ -3,13 +3,58 @@
...
+                write_byte_string(f, prefix)?;
+                f.write_str("'")
+            }
+            Self::InvalidPrefixBind { prefix, namespace } => {
+                f.write_str("The namespace prefix '")?;
```
Mingun, line 44: Same here
```suggestion
                f.write_str("the namespace prefix '")?;
```

src/errors.rs
```diff
@@ -1,16 +1,125 @@
...
+macro_rules! combined_error {
+    (@choose ($a:tt) ($b:tt)) => { $b };
+    (@choose ($a:tt)) => { $a };
+    (
+        $(#[$($attrss:tt)*])*
+        $name:ident;
+        $($variant:ident($error:path $(, $inner_type:path)?) $fmt_str:literal),+ $(,)?
+    ) => {
+        #[derive(Debug)]
+        #[allow(missing_docs)]
```
Mingun, line 38: I would like to update macro and write a documentation for each generated variant under which circumstances it will be returned. That is not always obvious from the name and description of the inner error.
```suggestion
```
Also, I would prefer to have more traditional syntax for defining enums, in order to in the end the written code will look like:
```rust
combined_error! {
  /// Doc
  // derives
  pub enum SpecificError {
    /// Variant 1 doc
    Variant1(Variant1Error) => "display 1 text",
    /// Variant 2 doc
    Variant2(Variant2Error) => "display 2 text",
  }
}
```
Please also derive `Debug` for each error type.

Mingun (changes requested):
Some doc links became broken, that need to be fixed
```
> cargo doc --all-features
warning: unresolved link to `encoding`
  --> src\encoding.rs:22:18
   |
22 | /// If feature [`encoding`] is disabled, the EncodingError is always `DecodeError::Utf8`:
   |                  ^^^^^^^^ no item named `encoding` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: unresolved link to `Error::IllFormed`
   --> src\reader\buffered_reader.rs:308:75
    |
308 |     /// If a corresponding [`End`] event is not found, an error of type [`Error::IllFormed`]
    |                                                                           ^^^^^^^^^^^^^^^^ no item named `Error` in scope

warning: unresolved link to `Error::IllFormed`
  --> src\reader\slice_reader.rs:88:75
   |
88 |     /// If a corresponding [`End`] event is not found, an error of type [`Error::IllFormed`]
   |                                                                           ^^^^^^^^^^^^^^^^ no item named `Error` in scope

warning: `quick-xml` (lib doc) generated 3 warnings
```

I'm fine with the first 3 commits, but I'm unsure about the 4th (other two commits just the follow-ups fixing the compilation errors). I'm not sure that such a fine breakdown of errors will be convenient for work. In most cases, you will not be interested in a specific type of error and it will still be thrown higher, but the presence of many types of errors will complicate such throwing (it's not for nothing that things like `anyhow::Error` appeared).

Do you have opposite experience?

---

You can fix things and make a force push, GitHub UI provides a way to compare between force pushes, so it is fine). I would prefer to have a history without fix-up commits, so it would be great if each commit:
- in a compilable state on CI (which means it is compiled fine with all CI tested combinations of flags, in practice usually check of `cargo test` and `cargo test --all-features` is enough)
- `cargo doc --all-features` does not report warnings
- each commit updates changelog with the relevant changes (so it is easely to understand later in which commit that change was made)

RedPhoenixQ:
> Some doc links became broken, that need to be fixed

Compiling the docs was an oversight. Will fix.

> I'm fine with the first 3 commits, but I'm unsure about the 4th (other two commits just the follow-ups fixing the compilation errors). I'm not sure that such a fine breakdown of errors will be convenient for work. In most cases, you will not be interested in a specific type of error and it will still be thrown higher, but the presence of many types of errors will complicate such throwing (it's not for nothing that things like `anyhow::Error` appeared).
> 
> 
> 
> Do you have opposite experience?

I'm mostly interested in not having irrelevant options when calling lower level apis. I will agree that the `attributes` methods are inconvenient when returning `ReadError` (like the `read_node` example). I still think that having these methods return `AttrError` is better for when only using those apis.

I'm torn on whether this is better, especially when it's common for `ReadError` and `AttrError` appear in the same function. I still prefer returning `AttrError` but I agree it may not be worth it.

If this split beyond `EncodingError` and `NamespaceError` is not desired, the `combined_error!` macro could also be removed entirely.

> You can fix things and make a force push, GitHub UI provides a way to compare between force pushes, so it is fine). I would prefer to have a history without fix-up commits, so it would be great if each commit:
>
> - in a compilable state on CI (which means it is compiled fine with all CI tested combinations of flags, in practice usually check of `cargo test` and `cargo test --all-features` is enough)
> 
> - `cargo doc --all-features` does not report warnings
>
> - each commit updates changelog with the relevant changes (so it is easely to understand later in which commit that change was made)

Absolutely. The docs was an oversight and I know that one commit doesn't pass the tests. I will go back and rework all of these commits.

RedPhoenixQ:
When the errors are being changed anyway, there's an opportunity to be consistent about whether Error enum variants should be named `SomeError::Io` or `SomeError::IoError`. Any preference here?

Mingun:
> I still think that having these methods return `AttrError` is better for when only using those apis.

When we declare in API a more wide error that is really could happen, I'm fine to narrow the result type, but introducing new fine-granulated error types for that, I think, would be overkill.

The problem is that the API may not be well-established, and with the introduction of validation checks, it may turn out that some functions will return more errors. Some Linux package maintainers already complained, that quick-xml API changes too quickly :).

> If this split beyond `EncodingError` and `NamespaceError` is not desired, the `combined_error!` macro could also be removed entirely.

Yes. So for now please left only the changes from the first 3 commits. Maybe in time the 4th commit also would be welcomed, who knows :)?

> Any preference here?

`SomeError::Io`

src/name.rs
```diff
@@ -3,13 +3,58 @@
...
+    ///
+    /// This error returned in following cases:
+    /// - the XML document attempts to bind `xml` prefix to something other than
+    ///   `http://www.w3.org/XML/1998/namespace`
+    /// - the XML document attempts to bind `xmlns` prefix
+    /// - the XML document attempts to bind some prefix (except `xml`) to
+    ///   `http://www.w3.org/XML/1998/namespace`
+    /// - the XML document attempts to bind some prefix to
+    ///   `http://www.w3.org/2000/xmlns/`
+    InvalidPrefixBind {
```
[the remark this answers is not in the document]
RedPhoenixQ: > If we split error into small parts, maybe make a dedicated variant for each listed variant?

I will add a commit for this soon. Need to understand the code and the linked standard to see which error applies were which may take a bit longer.

RedPhoenixQ:
I belive I have fixed all issues from the previous review. The changelog is now incrementally updated every commit and all commits pass `cargo test` and `cargo doc` (with `--all-features`). 

I have still included the `AttrError` change as I didn't understand if it should be discarded or not. Will remove it if you want.

src/name.rs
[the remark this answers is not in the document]
RedPhoenixQ: Done in commit 8002cc6

Mingun (approved):
I polished the PR slightly:
- Normalized all error messages to start from lowercase
- Made error-related code more consistent: removed unnecessary usages of `write!` macro, used `Self` instead of type name where possible (that two changes was related to the errors so it was logical to made them in that PR)
- Implemented `PartialEq` and `Eq` for `NamespaceError` to simplify testing of the methods returning results with that error
- That allowed to change tests from manual matching on result to use `assert_eq!` which will provide nice diff if failed
- You forgot to return `Some` from `Error::source` for `Error::Namespace` -- fixed
- Used tuple form instead of struct form for `NamespaceError` errors. Usually struct form with one field is not used
- Removed `BangType` from changelog. It is internal type not visible to users

Mingun:
Thanks!

PR #814, opened 2024-10-01
Implemented change implementing "Keep specific element to single line while using indent with serde serializer#655"

RedIODev:
Added checks to adjust indentation in case only attributes or empty elements are found.

### DIFF
--- a/src/se/element.rs
+++ b/src/se/element.rs
@@ -238,7 +238,7 @@ impl<'w, 'k, W: Write> Serializer for ElementSerializer<'w, 'k, W> {
         self.ser.writer.write_char('<')?;
         self.ser.writer.write_str(self.key.0)?;
         Ok(Struct {
-            start_whitespace: None,
+            //start_whitespace: None,
             contains_non_attribute_keys: false,
             ser: self,
             children: String::new(),
@@ -371,7 +371,7 @@ impl<'w, 'k, W: Write> SerializeTupleVariant for Tuple<'w, 'k, W> {
 ///   serializer
 pub struct Struct<'w, 'k, W: Write> {
     ser: ElementSerializer<'w, 'k, W>,
-    start_whitespace:Option<String>,
+    //start_whitespace:Option<String>,
     contains_non_attribute_keys:bool,
     /// Buffer to store serialized elements
     // TODO: Customization point: allow direct writing of elements, but all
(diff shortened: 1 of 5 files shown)


src/se/element.rs
```diff
@@ -368,6 +370,7 @@ impl<'w, 'k, W: Write> SerializeTupleVariant for Tuple<'w, 'k, W> {
 ///   serializer
 pub struct Struct<'w, 'k, W: Write> {
     ser: ElementSerializer<'w, 'k, W>,
+    contains_non_attribute_keys:bool,
```
Mingun, line 373: Could you add a bit of documentation, explaining how this field influence serialization (when it is `true` and when is `false`, what those values means)
```suggestion
    contains_non_attribute_keys: bool,
```

src/se/element.rs
```diff
@@ -475,9 +482,21 @@ impl<'w, 'k, W: Write> SerializeStruct for Struct<'w, 'k, W> {
...
+                let mut indentation = String::new();
+                self.ser.ser.indent.increase();
+                self.ser.ser.indent.write_indent(&mut indentation)?;
+                self.ser.ser.indent.decrease();
 
-            self.ser.ser.indent.write_indent(&mut self.ser.ser.writer)?;
+                let children = self.children.split_at(indentation.len()).1;
+
+                self.ser.ser.writer.write_str(children)?;
+            }
```
Mingun, line 499: So the idea is to cut written indentation. I wonder, is it possible to defer intendation write until we will sure that it is necessary?

RedIODev: That would require storing the serialized elements structured with their indentation separated instead of just the single string for all child elements. 

I would guess the overhead of storing separated indentation and recombining all these for every struct would be quite large compared to removing indentation in 1 minor case.
I would guess more than 75% of structs use the other if branch that uses the normal behavior.
Since we only care about the case where the TEXT (\<any>_TEXT_\</any>) part is a single statement this was the easiest implementation I could think of. If you have a better idea to get the desired behavior feel free. 

What I would suggest to take a second look is the new format for the tests mentioned above. It is questionable rather the formatting should be this way or rather they require a special case treatment.

Great regards
RedIODev

RedIODev: I can't get it to work. The Indentation::current().len() and Indentation::current_indent_len both are off by 1 byte in some but not all cases. I can't see another way to ask Indent for its length.
Great regards
RedIODev

Mingun: Ok, I'll look then myself. Thanks for PR!

Mingun:
I implemented alternative solution in #820. It handles all cases where text field may arise and does not require editing already written data. So I close this PR, but anyway thanks for submitting it!

PR #820, opened 2024-10-11
Do not write excess indents arount text fields when serialize with serde

### LINKED ISSUE #655: Keep specific element to single line while using indent with serde serializer
Hi, thanks for making this library; I'm getting a lot of use out of it.
However, I've run into an issue where I have to format some code a certain way that a proprietary reader expects it to be. Namely, I've got these structures:

```rs
#[derive(Deserialize, Serialize, Debug)]
pub struct TextureCoordinates {
    #[serde(rename = "@dimension")]
    pub dimension: String,
    #[serde(rename = "@channel")]
    pub channel: String,
    #[serde(rename = "$value")]
    pub elements: String,
}
#[derive(Deserialize, Serialize, Debug)]
pub struct VertexBuffer {
    pub positions: String,
    pub normals: String,
    #[serde(skip_serializing_if = "Option::is_none", default)]
    pub texture_coordinates: Option<TextureCoordinates>,
}
```

I'm writing like this, though admittedly with some context missing:
```rs
let mut ser = Serializer::with_root(&mut buffer, None).unwrap();
ser.indent(' ', 2);
ser.expand_empty_elements(true);

obj.serialize(ser).unwrap();
```
The texture coordinates serialise to something like this:
```xml
<VertexBuffer>
  <Positions>319.066 -881.28705 7.71589</Positions>
  <Normals>-0.0195154 -0.21420999 0.976593</Normals>
  <TextureCoordinates dimension="2D" channel="0">
    0.752494 0.201033,0.773967 0.201033
  </TextureCoordinates>
</VertexBuffer>

```
However, the program I need to serialize for requires this format instead:
```xml
<VertexBuffer>
  <Positions>319.066 -881.28705 7.71589</Positions>
  <Normals>-0.0195154 -0.21420999 0.976593</Normals>
  <TextureCoordinates dimension="2D" channel="0">752494 0.201033,0.773967 0.201033</TextureCoordinates>
</VertexBuffer>
```

Is there some way to achieve this that I'm not aware of?

I'm currently working around it by just not using indentation, so everything is written in a single line, but it would be nice to have.

Mingun:
This PR alternative and more full implementation of the fix for excess indents around text fields. It implements the [Python](https://github.com/tafia/quick-xml/issues/655#issuecomment-2395506831) scheme of writing indents:
```rust
struct Xml {
  before: String,// = "value1"
  #[serde(rename = "$text")] // or $value
  text: String,// = "text"
  after: String,// = "value2"
}
```
`before`, `$text` and `after`
```xml
<Xml>
  <before>value1</before>text<after>value2</after>
</Xml>
```
`$text` and `after`
```xml
<Xml>text<after>value2</after>
</Xml>
```
`before` and `$text`
```xml
<Xml>
  <before>value1</before>text</Xml>
```
`$text` only
```xml
<Xml>text</Xml>
```
The implementation never write indent after `Option` fields, because `Option` can contain a type that is sensible to the surrounding text content, for example `Option<String>`. Because serde does not provide the type of `None` when serialize it, it is impossible to be smart here and write indent in one cases but not other, so I choosed the safe variant.

As a side effect, `to_writer`, `to_writer_with_root` and `Serializer` not returns `bool` on success. The `true` means that is will be safe to write text (in particular, indent) after the serialized object, it will not be interpreted as a part of content during deserialization. If `false` will be returned, then writing additional text content may change data what deserializer will see.

Closes #655, closes #814

### DIFF
--- a/src/se/content.rs
+++ b/src/se/content.rs
@@ -3,7 +3,7 @@
 use crate::de::TEXT_KEY;
 use crate::se::element::{ElementSerializer, Struct, Tuple};
 use crate::se::simple_type::{QuoteTarget, SimpleTypeSerializer};
-use crate::se::{Indent, QuoteLevel, SeError, XmlName};
+use crate::se::{Indent, QuoteLevel, SeError, WriteResult, XmlName};
 use serde::ser::{
     Impossible, Serialize, SerializeSeq, SerializeTuple, SerializeTupleStruct, Serializer,
 };
@@ -15,7 +15,7 @@ macro_rules! write_primitive {
         #[inline]
         fn $method(self, value: $ty) -> Result<Self::Ok, Self::Error> {
             self.into_simple_type_serializer().$method(value)?;
-            Ok(())
+            Ok(WriteResult::Text)
         }
     };
 }
@@ -24,7 +24,9 @@ macro_rules! write_primitive {
 
 /// A serializer used to serialize content of an element. It does not write
 /// surrounding tags. Unlike the [`ElementSerializer`], this serializer serializes
-/// enums using variant names as tag names, i. e. as `<variant>...</variant>`
+/// enums using variant names as tag names, i. e. as `<variant>...</variant>`.
+///
+/// Returns the classification of the last written type.
 ///
 /// This serializer does the following:
 /// - numbers converted to a decimal representation and serialized as naked strings;
@@ -107,7 +109,7 @@ impl<'w, 'i, W: Write> ContentSerializer<'w, 'i, W> {
 
     /// Writes `name` as self-closed tag
     #[inline]
-    pub(super) fn write_empty(mut self, name: XmlName) -> Result<(), SeError> {
+    pub(super) fn write_empty(mut self, name: XmlName) -> Result<WriteResult, SeError> {
         self.write_indent()?;
         if self.expand_empty_elements {
             self.writer.write_char('<')?;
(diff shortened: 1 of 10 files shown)


Mingun:
Updated: even better solution. Now serialization returns the classification result (instead of just boolean) and you can decide what to do with it.

src/se/content.rs
```diff
@@ -1410,19 +1367,20 @@ pub(super) mod tests {
                 => Unsupported("character `<` is not allowed at the start of an XML name `<\"&'>`"));
 
             value!(newtype: Newtype(42) => "42");
-            value!(enum_newtype: Enum::Newtype(42) => "<Newtype>42</Newtype>");
+            value!(enum_newtype: Enum::Newtype(42) => "\n  <Newtype>42</Newtype>\n  ");
 
             // Note that sequences of primitives serialized without delimiters!
-            value!(seq: vec![1, 2, 3] => "1\n  2\n  3");
-            value!(seq_empty: Vec::<usize>::new());
+            value!(seq: vec![1, 2, 3] => "123");
```
dralley, line 1368: How is  `vec![1, 2, 3]` distinguished from `vec![123]`

src/se/content.rs
```diff
@@ -1410,19 +1367,20 @@ pub(super) mod tests {
...
+            value!(seq_empty: Vec::<usize>::new() => "");
             value!(tuple: ("<\"&'>", "with\t\n\r spaces", 3usize)
-                => "&lt;&quot;&amp;&apos;&gt;\n  \
-                    with\t\n\r spaces\n  \
+                => "&lt;&quot;&amp;&apos;&gt;\
+                    with\t\n\r spaces\
                     3");
-            value!(tuple_struct: Tuple("first", 42) => "first\n  42");
+            value!(tuple_struct: Tuple("first", 42) => "first42");
```
dralley, line 1374: I'm definitely missing something here.

How would `Tuple("first42", 0)` be distinguished from `Tuple("first", 420)`

src/se/content.rs
Mingun: Unfortunately, in any way. This is what happens now, if you do not use indents.

src/se/content.rs
Mingun: The same answer as above. This is restriction of current implementation. I have a TODO to forbid serialization consequent text fields:
https://github.com/tafia/quick-xml/blob/6eea6bbcb361a3ab51ccd1f5917d855d8a195c71/src/se/content.rs#L75

dralley: Is there an issue written up for that already, and if not can you write one and reference it

Mingun: No, there is no issue yet

src/se/content.rs
Mingun: In #823 I change the return value in that case to error

src/se/content.rs
Mingun: In #823 I changed the return value in that case to error

dralley (approved):

PR #823, opened 2024-10-15
Do not allow consequent primitives in `$value` fields and top-level

Mingun:
This PR changes serialization of `$value` fields with types like `Vec<String>` or `(u32, u32, u32)`. Previously this fields would be serialized without delimiters between elements which is unlikely what you want. Now serialization of that types would return `SeError::Unsupported`.

Example:
```rust
struct Xml {
  #[serde(rename = "$value")]
  seq: Vec<String>,

  #[serde(rename = "$value")]
  tuple: (bool, usize, String),
}
```

Currently it is based on #820, only the last commit implements the change.

This also partially handle this https://github.com/tafia/quick-xml/issues/497#issuecomment-2393236747.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -30,12 +30,15 @@
 - [#811]: Narrow down error return type from `Error` where only one variant is ever returned:
   attribute related methods on `BytesStart` and `BytesDecl` returns `AttrError`
 - [#820]: Classify output of the `Serializer` by returning an enumeration with kind of written data
+- [#823]: Do not allow serialization of consequent primitives, for example `Vec<usize>` or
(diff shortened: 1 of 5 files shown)


dralley (approved):
Might be helpful to mention the "consequent primitives" caveat in the documentation.

Mingun:
I think, it is already mentionen in [`$value`/Primitives and sequences of primitives](https://docs.rs/quick-xml/latest/quick_xml/de/index.html#primitives-and-sequences-of-primitives) part of documentation (but I forgot to update this part, now fixed). If you have better place where we can put it or describe it better, make a PR. Perhaps my gaze is too blurry.

Issue #824, opened 2024-10-18
How to deserialize a Struct with only Attributes from `BytesStart`?


C0D3-M4513R:
I'm currently trying to build a Project, which potentially deals with pretty big xml files.
Other conventional readers/applications (written in other languages) have trouble dealing with those xml files, due to running out of ram.
I want to future proof against the even bigger growing xml files and build a deserializer(/serializer) by using quick_xml's Reader, instead of just reading it all into memory.

However I can find no good solution to actually de-serializing structs, which only have attributes.
(I automatically generate two variants of my structs: one with only attributes and a builder [which is not be important for this issue])

Mingun:
It is unclear, what the troubles you have? Get the [`BytesStart::attributes`](https://docs.rs/quick-xml/latest/quick_xml/events/struct.BytesStart.html#method.attributes) iterator, read attributes and map them to you struct. Something like that:
```rust
struct YourStruct {
  attr1: String,
  attr2: String,
}
match reader.read_event()? {
  Event::Start(start) => {
    let attr1 = None;
    let attr2 = None;
    for attr in start.attributes() {
      match attr.name {
        QName(b"attr1") => attr1 = Some(attr.value),
        QName(b"attr2") => attr2 = Some(attr.value),
        _ => {} // skip unknown attributes
      }
    }
    YourStruct {
      attr1: attr1.unwrap(),
      attr2: attr2.unwrap(),
    }
  }
  _ => {} // other events
}
```

If you want, you may to write a serde deserializer that would deserialize type from the `BytesStart` that will do approximately the same in the `deserialize_map` method. Actually, implementing such deserializer in quick-xml is a good thing for PR.

C0D3-M4513R:
> It is unclear, what the troubles you have? Get the [`BytesStart::attributes`](https://docs.rs/quick-xml/latest/quick_xml/events/struct.BytesStart.html#method.attributes) iterator, read attributes and map them to you struct. Something like that:
> 
> ```rust
> struct YourStruct {
>   attr1: String,
>   attr2: String,
> }
> match reader.read_event()? {
>   Event::Start(start) => {
>     let attr1 = None;
>     let attr2 = None;
>     for attr in start.attributes() {
>       match attr.name {
>         QName(b"attr1") => attr1 = Some(attr.value),
>         QName(b"attr2") => attr2 = Some(attr.value),
>         _ => {} // skip unknown attributes
>       }
>     }
>     YourStruct {
[...]

PR #825, opened 2024-10-19
Make `SimpleTypeDeserializer` public

C0D3-M4513R:
Fixes: #824

I was able to get #824 working with the `SimpleTypeDeserializer`. Thank you so much for that tipp!


Mingun:
Thanks for PR! But I already working on that, besides, it requires a little more work then just add a reexport (update documentation, fix some quirks)

PR #827, opened 2024-10-20
Make `escape` and it variants take a `impl Into<Cow<str>>` argument and implement `From<(&'a str, Cow<'a, str>)>` on `Attribute`

Tpt:


Tpt:
Handy to easily create `Attribute` from owned values without having to play with lifetimes.

Issue #831, opened 2024-11-17
`]]>` in `CDATA` is not escaped


Turbo87:
I'm using the `rss` crate in the crates.io codebase, which uses `quick-xml` under the hood and for certain fields performs:

```rust
self.write_event(Event::Start(BytesStart::new(name)))?;
self.write_event(Event::CData(BytesCData::new(text.as_ref())))?;
self.write_event(Event::End(BytesEnd::new(name)))?;
```

According to https://en.wikipedia.org/wiki/CDATA#Nesting in a `CDATA` section `]]>` needs to be encoded as `]]]]><![CDATA[>` to not accidentally close the `CDATA` section.

When I'm using my test string (`"let's try & break this <item> ]]>, shall we?"`) however, I currently see this in the output:

```xml
<![CDATA[let's try & break this <item> ]]>, shall we?]]>
```

Apologies if this was already reported, but I couldn't find a matching issue. Is the escaping of `]]>` the responsibility of the caller in this case? 

I noticed that the `BytesCData` documentation says:

> Note, that inner text will not contain `]]>` sequence inside

but that does not seem to be targeted at people using the library for serializing `CDATA` sections.

Related:

- https://github.com/rust-syndication/rss/issues/173
- https://github.com/tafia/quick-xml/pull/374

Mingun:
Yes, creating some events requires manual escape of illegal data. When I starting documenting writing methods (that attempts never was public) I found this issue and planned to fix it, but never had time to do that. We need to change creation API for these events to return an iterator over events instead of single event. Something like:
```rust
impl BytesCData {
  pub fn new(content: &str) -> impl Iterator<BytesCData> { ... }
}
```
Comment event also suffers from that problem.

PR to implement this API is welcomed.

PR #832, opened 2024-11-17
Implement `BytesCData::escaped()` fn

Turbo87:
This PR should resolve https://github.com/tafia/quick-xml/issues/831 by implementing an `escaped()` fn on the `BytesCData` struct as suggested in https://github.com/tafia/quick-xml/issues/831#issuecomment-2481236477.

I opted for keeping the `new()` fn around and introducing another function instead because the `new()` fn is used in various places in the `quick-xml` test suite and I didn't want to break anything. This should also allow us to ship this PR in a non-breaking version release since the existing `new()` fn keeps working as-is. We could consider deprecating the `new()` fn though to make it easier for users to understand that they should use `escaped()` instead. We could also add a `raw()` fn as an alias for `new()` if we would like to keep the behavior for some users while still deprecating `new()`.

Related:

- https://github.com/tafia/quick-xml/issues/831
- https://github.com/rust-syndication/rss/issues/173
- https://github.com/rust-syndication/rss/pull/174
- https://github.com/tafia/quick-xml/pull/374

/cc @Mingun

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -41,6 +41,7 @@ pub mod attributes;
 use encoding_rs::Encoding;
 use std::borrow::Cow;
 use std::fmt::{self, Debug, Formatter};
+use std::iter::FusedIterator;
 use std::mem::replace;
 use std::ops::Deref;
 use std::str::from_utf8;
@@ -53,7 +54,7 @@ use crate::escape::{
 use crate::name::{LocalName, QName};
 #[cfg(feature = "serialize")]
 use crate::utils::CowRef;
-use crate::utils::{name_len, trim_xml_end, trim_xml_start, write_cow_string};
+use crate::utils::{name_len, trim_xml_end, trim_xml_start, write_cow_string, Bytes};
 use attributes::{AttrError, Attribute, Attributes};
 
 /// Opening tag data (`Event::Start`), with optional attributes: `<name attr="value">`.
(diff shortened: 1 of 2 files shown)


src/events/mod.rs
```diff
@@ -833,6 +846,49 @@ impl<'a> arbitrary::Arbitrary<'a> for BytesCData<'a> {
...
+    type Item = BytesCData<'a>;
+
+    fn next(&mut self) -> Option<BytesCData<'a>> {
+        if self.finished {
+            return None;
+        }
+
+        // If there are no more `]]>`, yield the remaining text and
+        // mark the iterator as finished.
+        let Some((next, _)) = self.iter.next() else {
```
Mingun, line 873: I think, that regular `match` would be more readable here. From my point of view `if let else` construction is hard to understand

src/events/mod.rs
```diff
@@ -1441,4 +1497,26 @@ mod test {
...
+        assert_eq!(cdata[1].as_ref(), b"> can be difficult and <invalid>");
+
+        let content = "foo]]>bar]]>baz]]>quux";
+        let cdata = BytesCData::escaped(content).collect::<Vec<_>>();
+        assert_eq!(cdata.len(), 4);
+        assert_eq!(cdata[0].as_ref(), b"foo]]");
+        assert_eq!(cdata[1].as_ref(), b">bar]]");
+        assert_eq!(cdata[2].as_ref(), b">baz]]");
+        assert_eq!(cdata[3].as_ref(), b">quux");
+    }
```
Mingun, line 1521: Move this test to doc-test of `escaped` method. It is enough to have only doc tests in such simple cases as writing.

src/events/mod.rs
```diff
@@ -833,6 +846,49 @@ impl<'a> arbitrary::Arbitrary<'a> for BytesCData<'a> {
...
+/// This iterator is created by the [`BytesCData::escaped`] method.
+///
+/// If a string contains `]]>`, it needs to be split into multiple `CDATA`
+/// sections, splitting the `]]` and `>` characters. This iterator yields
+/// each of those sections.
+pub struct CDataIterator<'a> {
+    content: &'a str,
+    iter: MatchIndices<'a, &'a str>,
+    start: usize,
+    finished: bool,
```
Mingun, line 860: Do we need a separate `finished` field? Isn't enough to rely on `MatchIndices` iterator returning `None`? [`MatchIndices` is `FusedIterator`](https://doc.rust-lang.org/std/str/struct.MatchIndices.html#impl-FusedIterator-for-MatchIndices%3C'a,+P%3E)

Also, the iterator always continues to yield `None` when exhausted, so implement `FusedIterator` for it.

src/events/mod.rs
```diff
@@ -833,6 +846,49 @@ impl<'a> arbitrary::Arbitrary<'a> for BytesCData<'a> {
...
 }
 
+/// Iterator over `CDATA` sections in a string.
+///
+/// This iterator is created by the [`BytesCData::escaped`] method.
+///
+/// If a string contains `]]>`, it needs to be split into multiple `CDATA`
+/// sections, splitting the `]]` and `>` characters. This iterator yields
+/// each of those sections.
+pub struct CDataIterator<'a> {
```
Mingun, line 856:
```suggestion
#[derive(Clone, Debug)]
pub struct CDataIterator<'a> {
```

src/events/mod.rs
```diff
@@ -833,6 +846,49 @@ impl<'a> arbitrary::Arbitrary<'a> for BytesCData<'a> {
...
+    content: &'a str,
```
Mingun, line 857: `MatchIndices` already contains something like that, but unfortunately does not give access to it. It could be possible to implement this iterator using memchr, but the current implementation is also fine.

src/events/mod.rs
```diff
@@ -700,12 +700,25 @@ impl<'a> BytesCData<'a> {
...
     ///
-    /// `content` must not contain the `]]>` sequence.
+    /// `content` must not contain the `]]>` sequence. You can use
+    /// [`BytesCData::escaped`] to escape the content instead.
     #[inline]
     pub fn new<C: Into<Cow<'a, str>>>(content: C) -> Self {
         Self::wrap(str_cow_to_bytes(content), Decoder::utf8())
     }
 
+    /// Creates an iterator of `BytesCData` from a string.
```
Mingun, line 710: Could you move tests here as examples and also explain a bit why this method returns iterator (i. e. describe that we need to split user data at `]]>` to create well-formed XML)

Mingun (changes requested):
Thanks for PR. Some minor improvements can be made here and there, and also please add changelog entry. Use link to the issue #831 in section for issues and brief description about added method in **New Features** section. See the other issues in the file for an example.

src/events/mod.rs
Turbo87: > Do we need a separate `finished` field? Isn't enough to rely on `MatchIndices` iterator returning `None`?

AFAICT that is not enough, but I might be wrong. the problem is that the `MatchIndices` gives us the positions of the N splitters (`]]>`), but we need to yield the N+1 parts around the splitters. it might be possible to use a different base iterator that already yields the N+1 relevant parts, but I haven't found one yet.

Turbo87:
thanks for the quick review. I just pushed a couple of commits and I think I addressed all of the comments :)

Mingun (approved):
Good. Only add changelog entry please.

Turbo87:
> Only add changelog entry please.

done!

Mingun:
In the end I actually rewrite the whole implementation to use `memchr`. I think, it is now more straightforward and easy for understanding. But thanks anyway for your PR! The changes will be released soon in `0.37.1`

Issue #499, opened 2022-10-28
The `se::to_writer` function now requires `std::fmt::Write` instead of `std::io::Write`


calops:
Is there a reason for this change? It makes it impossible to use this function with a `Vec<u8>` (which does implement `std::io::Write`), while it is still possible in the current released version.

Mingun:
Yes, the reason is because XML is a text-based format which could use many encodings. You should manually encode the string result to bytes for now. If you want an UTF-8 encoded data in `Vec<u8>`, it is easily and zero-cost to get it:
```rust
let ser = Serializer::new(String::new());
let string: String = data.serialize(ser).unwrap();
let bytes: Vec<u8> = string.into_bytes();
```

calops:
That's a fair point. I have a question though:

I create a `Writer`, which requires a `Write` argument that implements `std::io::Write`. But if I want to pass that writer around and append to what it's doing, I can only call `to_writer(writer.inner())`, which won't work, because `to_writer` requires `std::fmt::Write. Right now I'm having to use an intermediate type that implements the `fmt` version and forwards it to `io` (or vice versa).

I feel like the actual solution to my use-case would be the ability to call the `Writer` directly so that it can serialize a random argument I throw at it, without having to go through `create_element`, which requires defining a root tag (which I don't want in this context).

Mingun:
You want to mix serde serializer with event-based Writer API? Could you describe your case better? Also, I suppose you use master, right?

calops:
Basically, I write a bunch of nodes myself through the event-based API. But then the inner part is serialized through `serde`, so I end up calling `to_writer(ToFmtWrite(writer.inner()))` in the middle of my events, `ToFmtWrite` being a dumb newtype with the correct `fmt::Write` implementation for anything that implements `io::Write`. I'm using a `Vec<u8>` as the buffer for the writer (but I'd use a `String` just as well if it meant I don't have to do this `Write` trickery).

Being able to replace something like that

```rust
writer.write_event(Event::Start(BytesStart::new("foo")))?;
quick_xml::se::to_writer(ToWriteFmt(writer.inner()), &Bar { baz: 42 })?;
writer.write_event(Event::End(BytesEnd::new("foo")))?;
```

with something like this

```rust
writer
    .create_element("foo")
    .write_serialized_content(Bar { baz: 42 })?;
```

would be the ideal implementation for my use-case.

Mingun:
> with something like this
> 
> ```rust
> writer
[...]

PR #836, opened 2024-12-29
Add to_utf8_io_writer helper serializing into std::io::Write

pronebird:
This PR is related to #499 and adds a helper method to dump XML into `std::io::Write` restricted to utf-8 output. The name is a bit verbose trying to convey that a. it's utf-8 only b. it's `io::Write`r and not `fmt::Write` to avoid confusion

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,9 +15,13 @@
 
 ### New Features
 
+- [#836]: Add `se::to_utf8_io_writer()` helper compatible with `std::io::Write` and restricted to UTF-8 encoding.
+
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


Mingun (approved):
Could you also add a changelog entry under **New Features**? Do not forgot to add a link to PR, like other entries. All links placed at the end of corresponding version section (**Unreleased** in that case)

pronebird:
> Could you also add a changelog entry under **New Features**? Do not forgot to add a link to PR, like other entries. All links placed at the end of corresponding version section (**Unreleased** in that case)

Sure thing. Done.

Issue #839, opened 2025-01-17
`xs:list` deserialization does not split on all whitespace


yuriks:
Despite the [documentation](https://docs.rs/quick-xml/0.37.2/quick_xml/de/index.html#basics) explicitly mentioning that "According to the XML Schema specification, delimiters for elements is one or more space (` `, `\r`, `\n`, and `\t`) character(s).", `ListIter` only actually splits on ` ` (space): [`simple_type.rs`](https://github.com/tafia/quick-xml/blob/8f91a9c20eb67666eaccbc4e37fbe5e3adde3a44/src/de/simple_type.rs#L364).

Reproducer:
```rust
use serde::Deserialize;
use quick_xml::de::from_str;

#[derive(Deserialize)]
struct Root {
    #[serde(rename = "$text")]
    items: Vec<String>,
}

fn main() {
    const TEST_XML: &str = "<Root>one two\nthree four\tfive six</Root>";

    let root: Root = from_str(TEST_XML).unwrap();
    assert_eq!(
        root.items,
        ["one", "two", "three", "four", "five", "six"]
    );
    // assertion `left == right` failed
    //  left: ["one", "two\nthree", "four\tfive", "six"]
    // right: ["one", "two", "three", "four", "five", "six"]
}
```

Mingun:
Duplicate of #674

PR #843, opened 2025-02-01
Resolve "xs:list deserialization does not split on all whitespace"

JGM01:
Changed DELIMITER (u8) into an array DELIMITERS (u8; 4) that holds 4 possible delimiters.

This was implemented with issue #839 in mind, so I covered the following whitespaces: ` `, `\t`, `\r`, `\n`. It is fairly easy to include more if necessary.

### DIFF
--- a/src/de/simple_type.rs
+++ b/src/de/simple_type.rs
@@ -8,7 +8,6 @@ use crate::encoding::Decoder;
 use crate::errors::serialize::DeError;
 use crate::escape::unescape;
 use crate::utils::CowRef;
-use memchr::memchr;
 use serde::de::value::UnitDeserializer;
 use serde::de::{
     DeserializeSeed, Deserializer, EnumAccess, IntoDeserializer, SeqAccess, VariantAccess, Visitor,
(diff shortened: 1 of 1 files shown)


src/de/simple_type.rs
```diff
@@ -361,14 +360,17 @@ impl<'de, 'a> SeqAccess<'de> for ListIter<'de, 'a> {
...
+            const DELIMETERS: [u8; 4] = [b' ', b'\t', b'\r', b'\n']; 
 
             loop {
                 let string = content.as_str();
                 if string.is_empty() {
                     return Ok(None);
                 }
-                return match memchr(DELIMITER, string.as_bytes()) {
+
+                let first_delimiter = string.as_bytes().iter().position(|c| DELIMETERS.contains(c));
```
adamreichold, line 371: [`str::find`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.find) will accept an array of characters and give you the first matching position which is likely faster than iterating bytes.

Alternatively, two calls to [`memchr2`](https://docs.rs/memchr/latest/memchr/fn.memchr2.html) might be an alternative that might be a bit faster yet.

src/de/simple_type.rs
```diff
@@ -391,7 +393,7 @@ impl<'de, 'a> SeqAccess<'de> for ListIter<'de, 'a> {
                     // `content` started with a space, skip them all
                     Some(0) => {
                         // Skip all spaces
-                        let start = string.as_bytes().iter().position(|ch| *ch != DELIMITER);
+                        let start = string.as_bytes().iter().position(|c| !DELIMETERS.contains(c));
```
adamreichold, line 396: Similar to the above, [`str::find`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.find) also accepts a closure checking whether a character matches.

src/de/simple_type.rs
Mingun: I think, it will need to run benchmarks to say what is faster. Probably the difference would be not significant because the lists rarely are big in real applications.

Mingun (changes requested):
Thanks for PR. I would like to have:
- a new test that ensures that bug is fixed
- a changelog entry informing users that bug is fixed

src/de/simple_type.rs
adamreichold: > Probably the difference would be not significant because the lists rarely are big in real applications.

In that case,

```rust
let first_delimiter = string.find(DELIMITERS);
```

is certainly simpler.

JGM01: Rewrote to use `str::find`.

src/de/simple_type.rs
JGM01: Rewrote to use `str::find`.

JGM01:
Sorry for necro but has this been looked at? Was just looking through my old repos and saw the fork.

Issue #353, opened 2022-01-26
Serializing CDATA


keithmss:
My apologies if this is an inappropriate place to raise this issue.

I have the following structure that I am trying to serialize:

```rust
use serde::Serialize;

/// `Description` field for `Tag`.
#[derive(Serialize)]
pub(super) struct Description {
    #[serde(rename = "$value")]
    value: String,
}

impl Description {
    /// Create a new `Description`.
    pub(super) fn new(input: &str) -> Self {
        let value = format!("<![CDATA[{}]]>", input);
        Self { value }
    }
}
```

As you can see I'm wrapping it in `CDATA` as per a file format that I'll be writing that I have no control over. This structure is actually just part of a larger structure:

```rust
#[derive(Serialize)]
#[serde(rename_all = "PascalCase")]
pub(crate) struct Tag {
    name: String,
    tag_type: TagType,
    data_type: DataType,
    dimensions: Option<usize>,
    radix: Radix,
    constant: Constant,
    external_access: ExternalAccess,
    description: Description,
}

impl Tag {
    /// Return an XML `String` representation of `Tag`.
    pub(crate) fn to_xml(&self) -> Result<String, DeError> {
        to_string(self)
    }
}
```

Which is serializing perfectly (except for the description field):

```xml
<Tag Name="test_dint" TagType="Base" DataType="DINT" Radix="Decimal" Constant="false" ExternalAccess="Read Only"><Description>&lt;![CDATA[Test DINT]]&gt;</Description></Tag>
```

It looks like when it serializes the CDATA portion is being escaped:

```xml
<Description>&lt;![CDATA[Test DINT]]&gt;</Description>
```

Is there anything I can do to unescape it ?

Mingun:
Actually, the serializer worked as expected. You should not write `<![CDATA[]]>` manually as part of your data, because it is a part of format. I suppose that currently serializer never uses CDATA when serializers strings. That can be improved.

Because we are limited in a way to pass options from structure to the deserializer due to serde design, the possible solutions could be a setting in a serializer how to serialize strings:
- text always
- CDATA always
- a function to select Text/CDATA representation
- an automatic choice based on the content, as [XmlBeans](https://xmlbeans.apache.org/), for example, do. That choice could be implemented as a default format selection function from the previous point

PR is welcome!

kulame:
any update for this issue?

Mingun:
[...]

PR #849, opened 2025-03-07
WIP: Serde CData

anton-dutov:
@Mingun 
I made some draft for $cdata serialization, deserialization the same as $text field, can you review?

### DIFF
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -5,7 +5,7 @@ use crate::{
     de::resolver::EntityResolver,
     de::simple_type::SimpleTypeDeserializer,
     de::text::TextDeserializer,
-    de::{DeEvent, Deserializer, XmlRead, TEXT_KEY, VALUE_KEY},
+    de::{DeEvent, Deserializer, XmlRead, TEXT_KEY, CDATA_KEY, VALUE_KEY},
     encoding::Decoder,
     errors::serialize::DeError,
     errors::Error,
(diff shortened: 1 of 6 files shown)


Mingun (changes requested):
I would prefer to not introduce `$cdata` rename, because serializing text as CDATA is a rather serializer choice instead of datatype choice. I already outlined my thoughts [here](https://github.com/tafia/quick-xml/issues/353#issuecomment-1339685379).

Also, you implementation just wrong, because it not escaping `]]>` in the serialized string and escapes `&`, `<` and `>`. The essential point of any CDATA-related PR should be test of those corner cases.

anton-dutov:
> I would prefer to not introduce `$cdata` rename, because serializing text as CDATA is a rather serializer choice instead of datatype choice. I already outlined my thoughts [here](https://github.com/tafia/quick-xml/issues/353#issuecomment-1339685379).
>
I understand your point, but unfortunately some software, e.g. implementing [IATA CUPPS](https://www.iata.org/en/programs/passenger/common-use/) understand only CDATA in some elements, there are many of them and you can't change them. I can't explicitly tell the serializer that the data should be serialized as CDATA, so I'm looking for a solution.  As I understand I am not the only one with this problem. 

> Also, you implementation just wrong, because it not escaping `]]>` in the serialized string and escapes `&`, `<` and `>`. The essential point of any CDATA-related PR should be test of those corner cases.

Adding tests and fixing escaping is not a big deal, it's more a question of how to idiomatically add this functionality

Mingun:
Ok, if a real use case exist for that, that we can add `$cdata` special rename in addition to the setting to the serializer. Would you like to implement variant with setting too?

anton-dutov:
I apologize for bothering you. After a long correspondence, the software manufacturers acknowledged the problem and promised to solve it within a year. Then I will roll back to a temporary solution (waiting for a miracle), and close the PR

PR #852, opened 2025-03-30
Derive implementations

lukasfri:


lukasfri:
I seem to accidentally have published this PR a bit early, I planned on making some more improvements to the NsReader to justify this PR, but personally I really only need Clone implemented for my own uses, and added Debugs due to me not seeing any reason why they _wouldn't_ be implemented.

Mingun:
Yes, I agree that the lack of implementation of those traits is an oversight. If you plan to add something to this PR, please also add a changelog entry about added trait implementations

lukasfri:
Sorry, I don't know how I accidentally wrote the wrong PR number. That's dumb.

PR #856, opened 2025-04-22
Add a `read_text` implementation for async readers

BitSyndicate1:
This adds a `read_text_into_async` method for Readers which allows writing the content in between nested tags to an arbitrary `AsyncWriter`, analogous to `Reader::read_text` and `NsReader::read_text` respectively.


Mingun (changes requested):
Need to add tests to ensure that this will work as expected (especially check that trailing spaces are included in the text). Doc tests will be enougth

PR #857, opened 2025-04-26
Make BytesCData decode method public

mematthias:
Currently, retrieving the decoded string content of a `quick_xml::events::Event::CData(cdata)` requires manually obtaining the Decoder instance from the Reader and calling `reader.decoder().decode(cdata.as_ref())`

This PR addresses this by changing the visibility of the internal decode method associated with BytesCData from `pub(crate)` to `pub`.

This enables users to directly call the decoding logic on the BytesCData instance itself:
```rust
match cdata.decode() {
    Ok(decoded_string) => { /* ... */ }
    Err(err) => { /* ... */ }
}
```

### DIFF
--- a/src/events/mod.rs
+++ b/src/events/mod.rs
@@ -843,8 +843,13 @@ impl<'a> BytesCData<'a> {
         ))
     }
 
-    /// Gets content of this text buffer in the specified encoding
-    pub(crate) fn decode(&self) -> Result<Cow<'a, str>, EncodingError> {
+    /// Decodes the raw input byte content of the CDATA section into a string,
+    /// without performing XML entity escaping.
+    ///
+    /// When this event produced by the XML reader, it uses the encoding information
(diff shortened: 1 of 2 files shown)


Mingun (approved):
Could you also add a changelog entry similarly to existing entries?

src/events/mod.rs
```diff
@@ -844,7 +844,7 @@ impl<'a> BytesCData<'a> {
     }
 
     /// Gets content of this text buffer in the specified encoding
-    pub(crate) fn decode(&self) -> Result<Cow<'a, str>, EncodingError> {
+    pub fn decode(&self) -> Result<Cow<'a, str>, EncodingError> {
```
Mingun, line 847: Also need to change doc string, because it is currently outdated. Will be used an encoding of a reader from which this event was produced.

Mingun:
Thanks!

PR #860, opened 2025-05-04
Simplify Deserializer::peek to avoid unreachable

Dr-Emann:
Use `Option::insert` which allows getting a reference to the value just inserted into the Option.

Unfortunately, we can't simplify this to just `Option::get_or_insert_with` because we can fail to generate the value to insert (if reader.next() fails).


Mingun (approved):
Good catch. Just add missing comma and I merge it.

PR #865, opened 2025-05-29
Yet another approach to disable trim

ggodlewski:
I implemented disabling trim by modifying Text struct.
See: https://github.com/tafia/quick-xml/issues/285#issuecomment-2863592598

I need this for parsing `.odt` files.
Can anybody give me some feedback? This issue is 4 years old and I'm willing to fix it if possible.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -20,6 +20,9 @@ Now references to entities (as predefined, such as `&lt;`, as user-defined) repo
 Caller can parse the content of the entity and stream events from it as it is required by the
 XML specification. See the updated `custom_entities` example!
 
+Implement whitespace behavior in the standard in `Deserializer`, which says string primitive
+types should preserve whitespace, while all other primitives have collapse behavior.
+
 ### New Features
 
 - [#863]: Add `Attributes::into_map_access(&str)` and `Attributes::into_deserializer()` when `serialize`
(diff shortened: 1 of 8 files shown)


Mingun (changes requested):
Good job! Because you anyway will add changes I would like to see that PR force pushed and to have at least 2 commits:
- the first one introduces new fields and API for `Text` (with mention in changelog), adapting code to creating `Text`s and nothing more
- the second uses this API, changes tests and adds changelog about fixed issue

Mingun (approved):
I fixed problems that I found:
- we need to use definition of "whitespace" that is used by XML specification. That is only 4 characters: space, tab, newline and carridge return
- I asked for the documentation with examples for the new methods of `Text`, I added them
- excess count of `skip_whitespaces` calls. I left only necessary minimum. I suspect, that you tried to fix tests that checks that all events are consumed. I just rewrite the check (and also fixed a bug in the `is_empty()` method)
- you forgot to add an actual link to the issue in the changelog. GitHub automatically detects some links, but the changelog should be also readable without GitHub
- after changes, unused code appeared, I deleted it

@ggodlewski, could you check that the rewrited PR solves you problems? I deleted your test, because we already have similar tests and I try to not repeat.

@dralley, could you also look at this?

PR #866, opened 2025-06-06
flatten_enum.rs added serialize 

dizzyi:
Added enum serializing to `example/flattened_enum.rs`
- added implementation of `Serialize` to `Elem`
- updated comment
- added comment of expected output under `println!` statement

### note
still trying to familiarize myself with this crate. 
I am not sure if `@type` key should count toward `len` of map here, it does in this example, not sure if that matter.

### run
```shell
cargo run --example flattened_enum --features="serialize"
```
*output*
```
Model { elem: [Foo(Foo { a: "1", subfoo: SubFoo { a1: "2", a2: "42", a3: "1337" } }), Bar(Bar { b: "22" })] }
<model><elem type="foo"><a>1</a><subfoo><a1>2</a1><a2>42</a2><a3>1337</a3></subfoo></elem><elem type="bar"><b>22</b></elem></model>
```

### DIFF
--- a/examples/flattened_enum.rs
+++ b/examples/flattened_enum.rs
@@ -1,5 +1,5 @@
-//! This example demonstrates how to deserialize enum nodes using an intermediate
-//! custom deserializer.
+//! This example demonstrates how to deserialize and serialize enum nodes using an intermediate
+//! custom deserializer and seralizer.
 //! The `elem` node can either be a `Foo` or a `Bar` node, depending on the `type`.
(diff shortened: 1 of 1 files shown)


Mingun:
> I am not sure if @type key should count toward len of map here, it does in this example, not sure if that matter.

Yes, it should. Usually serialization part with serde are straightforward and not needs additional explanation in addition to official serde guidem, but I'm not against it.

PR #870, opened 2025-06-18
Bump `criterion` version

eirnym:
Built and tested, 

please note: I haven't run benches or MSRV check locally

`criterion::black_box` is replaced to follow directions in deprecation message `Deprecated: use std::hint::black_box() instead`


eirnym:
A note about failing MSRV:

if you expect that an example app will work with your library with a given version, do exactly that.

PR #872, opened 2025-06-21
Update `compare` project to latest versions

Mingun:
Allow to benchmarking latest available versions. Because performance of maybe_xml degraded by 2 times, I kept version 0.10 which was (and still is) the fastest XML parser.

### DIFF
--- a/compare/Cargo.toml
+++ b/compare/Cargo.toml
@@ -29,5 +29,9 @@ serde_derive = "1.0"
 pretty_assertions = "1.4"
 
 [[bench]]
-name = "bench"
+name = "low-level"
+harness = false
+
+[[bench]]
+name = "serde"
 harness = false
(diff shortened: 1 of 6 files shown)


dralley (approved):

dralley:
Why did `maybe_xml` performance degrade?  Because if it was perhaps for some spec-compliance reason, IMO the thing that would make the most sense to track the latest version - because it's not like we subject `quick-xml` to the same handicap (though we obviously don't have 100% compliance)

Mingun:
> Why did `maybe_xml` performance degrade?

I don't know, but when I studied the source code of `maybe_xml` 0.10, it seemed to me very logical and straightforward. It almost correctly parse [well-formed](https://www.w3.org/TR/xml11/#sec-well-formed) documents ([except for inline DTD declarations](https://github.com/bluk/maybe_xml/issues/9)). It could be ideal for situations where you are sure that you have valid documents without DTD (which is anyway rarely used because replaced by XML Schema). That is why I want to keep it in the comparison as an ideal to strive for.

The slowdown is most likely due to the fact that in version 0.11 the parser now checks the validity of characters during parsing in each state, and not just looks for boundaries of events (like it did in 0.10 and as `quick-xml` does).

@bluk, just that you know

Issue #654, opened 2023-09-20
Unable to serialize a struct that deserialization works for


tatupesonen:
I'm trying to serialize/deserialize some data that an API responds with. This is what the XML from the API looks like:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<soapenv:Envelope xmlns:soapenc="http://schemas.xmlsoap.org/soap/encoding/"
    xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/"
    xmlns:xsd="http://www.w3.org/2001/XMLSchema"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <soapenv:Body>
        <LoginResponse xmlns="urn:vim25">
             <returnval>
                <userName>username_here</userName>
                <loggedInAt>datetime_here</loggedInAt>
            </returnval>
        </LoginResponse>
    </soapenv:Body>
</soapenv:Envelope>
```

And this is my code:
```rust
#[derive(Deserialize, Debug, Serialize)]
#[serde(rename = "soapenv:Envelope")]
pub struct SoapResponse {
    #[serde(rename = "@xmlns:soapenc")]
    soapenc: Option<String>,
    #[serde(rename = "@xmlns:soapenv")]
    soapenv: Option<String>,
    #[serde(rename = "@xmlns:xsd")]
    xsd: Option<String>,
    #[serde(rename = "@xmlns:xsi")]
    xsi: Option<String>,
    #[serde(rename = "$value")]
    pub body: SoapBody,
}

#[derive(Deserialize, Debug, Serialize)]
pub struct SoapBody {
    #[serde(rename = "$value")]
    pub response_type: Response,
}

#[derive(Deserialize, Debug, Serialize)]
pub enum Response {
    #[serde(rename = "LoginResponse")]
    LoginResponse {
        #[serde(rename = "@xmlns")]
        xmlns: String,
        #[serde(rename = "$value")]
        returnval: LoginResponse,
    },
    #[serde(rename = "Other")]
    SomeOther {},
}

#[derive(Deserialize, Debug, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct LoginResponse {
    logged_in_at: String,
    user_name: String,
}
```

Deserialization works fine. However, when I try to serialize a structure like this:
```rust
use quick_xml::se::to_string;

let thing = SoapResponse {
    soapenc: None,
    soapenv: None,
    xsd: None,
    xsi: None,
    body: {
[...]

PR #878, opened 2025-07-16
feat: support serializing structs into `$value`-named fields

Syndelis:
## About this PR

This PR implemensts support for serializing `$value`-named struct fields. These fields can already be correctly deserialized, but causes this runtime error when attempted to be serialized:

```
serialization of struct `StructName` is not supported in `$value` field
```

## Rationale

For an integration I'm working on, I've got to consume a relatively slim SOAP API. All requests are expected to be wrapped around `<soapenv:Envelope><soapenv:Body> ... </soapenv:Body></soapenv:Envelope>`. 

This is very similar to the issue in #654, but that could be solved by using enums. However, for various reasons, my specific use case would benefit from using a generic field inside the `SoapEnvBody` struct, like so:

```rust
#[derive(Serialize)]
#[serde(rename = "soapenv:Envelope")]
struct SoapEnvEnvelope<T> {
    #[serde(rename = "soapenv:Body")]
    body: SoapEnvBody<T>,
}

#[derive(Serialize)]
struct SoapEnvBody<T> {
    body: T,
}
```

However, that would produce an unwanted `<body>` tag. To get rid of it, I attempted to use `#[serde(flatten)]`, but that unexpectedly gets rid of one of the inner tags as well. To add to the previous example:

<details>
<summary>Long example</summary>

```rust
#[derive(Serialize)]
#[serde(rename = "soapenv:Envelope")]
struct SoapEnvEnvelope<T> {
    #[serde(rename = "soapenv:Body")]
    body: SoapEnvBody<T>,
}

#[derive(Serialize)]
struct SoapEnvBody<T> {
    #[serde(flatten)]
    body: T,
}

#[derive(Serialize)]
struct A {
    b: B,
}

#[derive(Serialize)]
struct B {
    field: &'static str,
}

println!(
    "{}",
    to_string(&SoapEnvEnvelope {
        body: SoapEnvBody {
            body: A {
                b: B { field: "my string" }
            }
        }
    })
    .unwrap()
);
```
</details>

The output of that example is:
```xml
<soapenv:Envelope><soapenv:Body><b><field>my string</field></b></soapenv:Body></soapenv:Envelope>
```

This output is missing `A`'s tag. With the implementation provided in this PR for `$value`, the above example can be modified like so:

<details>
<summary>Previous example, swapping `flatten` for `rename = "$value"`</summary>

```rust
#[derive(Serialize)]
#[serde(rename = "soapenv:Envelope")]
struct SoapEnvEnvelope<T> {
    #[serde(rename = "soapenv:Body")]
    body: SoapEnvBody<T>,
}

#[derive(Serialize)]
struct SoapEnvBody<T> {
    #[serde(rename = "$value")]
    body: T,
}

#[derive(Serialize)]
struct A {
    b: B,
}

#[derive(Serialize)]
struct B {
    field: &'static str,
}

println!(
    "{}",
    to_string(&SoapEnvEnvelope {
        body: SoapEnvBody {
            body: A {
                b: B { field: "my string" }
            }
        }
    })
    .unwrap()
);
```

</details>

To produce the desired output (notice the `<A>` tag):
```xml
<soapenv:Envelope><soapenv:Body><A><b><field>my string</field></b></A></soapenv:Body></soapenv:Envelope>
```

## Closing thoughts

Admittedly, this is a very simple solution, which just reuses what's already implemented. A lot of test comments mention how we "cannot" and "should not" wrap structs in this way, and that got me wondering if the comment was just referring to the current state, or if I'm missing any specific reason this should not be implemented. Considering the only tests that failed are exactly the tests that expect this behavior not to be implemented, I believe it's the former, but I'm open to hearing otherwise.

Also, some test modules include both `err!` and `serialize_as!` macros which acutally behave differently, usually with `serialize_as!` wrapping the input around a struct, map or something else, while the corresponding `err!` macro does not. This looks silly when reviewing the code changes, as it almost looks like I removed some tests and replaced them with completely different tests. That's not the case.

## Regarding Documentation

I have not updated the repository's documentation. I'd first like to see whether this is a wanted feature or not, before investing time into rewriting the docs. I'm fully open to updating them before merging this PR, though, I just wanted to get a general review first.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -1577,15 +1577,24 @@
 //! uses that name. This will allow you to switch XML crates more smoothly if required.
 //! </div>
 //!
-//! Representation of primitive types in `$value` does not differ from their
-//! representation in `$text` field. The difference is how sequences are serialized.
-//! `$value` serializes each sequence item as a separate XML element. The name
-//! of that element is taken from serialized type, and because only `enum`s provide
-//! such name (their variant name), only they should be used for such fields.
+//! The representation of primitive types in `$value` does not differ from their
+//! representation in `$text` fields. The difference is how sequences are serialized
+//! and deserialized. `$value` serializes each sequence item as a separate XML element.
+//! How the name of the XML element is chosen depends on the field's type. For
+//! `enum`s, the variant name is used. For `struct`s, the name of the `struct`
+//! is used.
 //!
-//! `$value` fields does not support `struct` types with fields, the serialization
-//! of such types would end with an `Err(Unsupported)`. Unit structs and unit
-//! type `()` serializing to nothing and can be deserialized from any content.
+//! During deserialization, if the `$value` field is an enum, then the variant's
+//! name is matched against. That's **not** the case with structs, however, since
+//! `serde` does not expose type names of nested fields. This does mean that **any**
+//! type could be deserialized into a `$value` struct-type field, so long as the
+//! struct's fields have compatible types (or are captured as text by `String`
+//! or similar-behaving types). This can be handy when using generic types in fields
+//! where one knows in advance what to expect. If you do not know what to expect,
+//! however, prefer an enum with all possible variants.
+//!
+//! Unit structs and unit type `()` serialize to nothing and can be deserialized
+//! from any content.
 //!
 //! Serialization and deserialization of `$value` field performed as usual, except
 //! that name for an XML element will be given by the serialized type, instead of
@@ -1644,6 +1653,38 @@
 //! #     AnyName { field: Enum::B },
 //! #     quick_xml::de::from_str("<root><B/></root>").unwrap(),
 //! # );
+//! ```
+//!
+//! The next example demonstrates how generic types can be used in conjunction
+//! with `$value`-named fields to allow the reuse of wrapping structs. A common
+//! example use case for this feature is SOAP messages, which can be commmonly
+//! found wrapped around `<soapenv:Envelope> ... </soapenv:Envelope>`.
+//!
+//! ```rust
+//! # use serde::Serialize;
+//! # use pretty_assertions::assert_eq;
+//! #
+//! #[derive(Serialize)]
+//! struct Envelope<T> {
+//!     body: Body<T>,
+//! }
+//!
+//! #[derive(Serialize)]
+//! struct Body<T> {
+//!     #[serde(rename = "$value")]
+//!     inner: T,
+//! }
+//!
+//! #[derive(Serialize)]
+//! struct Example {
+//!     a: i32,
+//! }
+//!
+//! assert_eq!(
+//!     quick_xml::se::to_string(&Envelope { body: Body { inner: Example { a: 42 } } }).unwrap(),
+//!     // Notice how `inner` is not present in the XML
+//!     "<Envelope><body><Example><a>42</a></Example></body></Envelope>",
+//! );
 //! ```
 //!
 //! ### Primitives and sequences of primitives
--- a/src/se/content.rs
+++ b/src/se/content.rs
@@ -179,7 +179,7 @@ impl<'w, 'i, W: Write> Serializer for ContentSerializer<'w, 'i, W> {
     type SerializeTupleStruct = Seq<'w, 'i, W>;
     type SerializeTupleVariant = Tuple<'w, 'i, W>;
     type SerializeMap = Impossible<Self::Ok, Self::Error>;
-    type SerializeStruct = Impossible<Self::Ok, Self::Error>;
+    type SerializeStruct = Struct<'w, 'i, W>;
     type SerializeStructVariant = Struct<'w, 'i, W>;
 
     write_primitive!(serialize_bool(bool));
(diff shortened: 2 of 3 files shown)


Mingun:
> A lot of test comments mention how we "cannot" and "should not" wrap structs in this way, and that got me wondering if the comment was just referring to the current state, or if I'm missing any specific reason this should not be implemented.

This constraint was added with the idea that the data type name is not part of a type. So if serialize them, we actually introduce strings from air, which could lead to the oddity. When we serialize top-level struct, there are no problems: root tag name is a part of _XML document_, not part of the _XML type_ (as defined by XSD) of this document. And we map rust types to XSD types.

Type aliases also makes things difficult for understanding. Unlike type name, field names cannot be aliased. For example, in you example you would be able to deserialize two different types
```rust
#[derive(Serialize)]
struct A {
  c: String,
}
#[derive(Serialize)]
struct B {
  c: String,
}

#[derive(Serialize)]
struct SoapEnvBody<T> {
    #[serde(rename = "$value")]
    body: T,
}
```
from both XMLs to which each of them are serialized:
```xml
<soapenv:Envelope>
  <soapenv:Body>
    <A><c>Data</c></A>
  </soapenv:Body>
</soapenv:Envelope>

<soapenv:Envelope>
  <soapenv:Body>
    <B><c>Data</c></B>
  </soapenv:Body>
</soapenv:Envelope>
```
If you are not embarrassed by this, then I am ready to accept this PR, but I want explicitly clarify in the [documentation](https://docs.rs/quick-xml/latest/quick_xml/de/index.html#value), that although struct name is serialized, any element name anyway will be accepted as valid on deserialization, and mark, that this is supported only for use with generics. You can past your example with SOAP here. For usual you should use correct field names.

I also want to point that you problem may be solved by manual implementation of the `Serialize` trait for `SoapEnvBody`. That is one-shot action and that should not be too difficult. first of all, look at the expanded macros for flatten variant:
```rust
impl<T: serde::Serialize> serde::Serialize for SoapEnvBody<T> {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
        where S: serde::Serializer,
    {
        let mut state = serializer.serialize_map(None)?;
        self.body.serialize(serde::__private::ser::FlatMapSerializer(&mut state))?;
        serde::ser::SerializeMap::end(state)

        // For reference, without #[serde(flatten)]:
        // let mut state = serializer.serialize_struct("SoapEnvBody", 1)?;
        // state.serialize_field("body", &self.body)?;
        // serde::ser::SerializeStruct::end(state)
    }
}
```
`FlatMapSerializer` does not record the type name for structs. So you need to write a copy of that serializer that will write it, in the same way as `FlatMapSerializer::serialize_struct_variant` does that.

Syndelis:
Thanks for taking your time and reviewing this!

> This constraint was added with the idea that the data type name is not part of a type. So if serialize them, we actually introduce strings from air, which could lead to the oddity. When we serialize top-level struct, there are no problems: root tag name is a part of _XML document_, not part of the _XML type_ (as defined by XSD) of this document. And we map rust types to XSD types.

I see! Thanks for clarifying that.

> Type aliases also makes things difficult for understanding. Unlike type name, field names cannot be aliased. For example, in you example you would be able to deserialize two different types
> [...]
> If you are not embarrassed by this, then I am ready to accept this PR, but I want explicitly clarify in the [documentation](https://docs.rs/quick-xml/latest/quick_xml/de/index.html#value), that although struct name is serialized, any element name anyway will be accepted as valid on deserialization, and mark, that this is supported only for use with generics. You can past your example with SOAP here. For usual you should use correct field names.

Sounds fair to me; I'll add a piece to the documentation mentioning this behavior. To be clear, this PR only introduces the serialization part of this behavior. The fact that deserialization would accept any element name is already present in the current version of the library. You should be able to verify this by running this example in `master`:

```rust
// Run this example in `master`, not this PR's branch

struct A {
    c: String,
}
struct B {
    c: String,
}

struct Body<T> {
    #[serde(rename = "$value")]
    body: T,
}

struct Envelope<T> {
    body: Body<T>,
}

let a = "<Envelope><body><A><c>Data</c></A></body></Envelope>";

let big_a: Envelope<A> = from_str(a).unwrap(); // Ok!
let big_b: Envelope<B> = from_str(a).unwrap(); // Ok!
```
 
> I also want to point that you problem may be solved by manual implementation of the `Serialize` trait for `SoapEnvBody`. That is one-shot action and that should not be too difficult. first of all, look at the expanded macros for flatten variant:
> [...]
> `FlatMapSerializer` does not record the type name for structs. So you need to write a copy of that serializer that will write it, in the same way as `FlatMapSerializer::serialize_struct_variant` does that.

Thanks for pointing that out. Of course, generally speaking, I could write a custom deserializer for most issues I'd encounter. In this particular case, I'd need to write it more than twice, as I have at least 3 structs which would need to support the same type of serialization as is fixed by this PR. At that point, I could write a macro to automate that, or contribute to the library itself. Seeing as it does not break any current behavior, I assumed it was fine to move this into the library itself. Please let me know otherwise! For the moment being, I'll be updating the docs.

Syndelis:
@Mingun I've edited the documentation of `$value` to mention how it works for structs in 353f890. Let me know what you think

Issue #761, opened 2024-06-17
Issue with serialization when using serde(flatten) 


Dekker1:
I ran into an issue where trying to use serde to serialize using quick-xml, does not work when you use `#[serde(flatten)]` in the root structure. However, the code seems to work when the struct is used further down in a nested structure, and moving the members of the struct into the original it works as well.

The following test case currently fails:
```rust
	#[test]
	fn flatten_serialize() {
		#[derive(Debug, PartialEq, Serialize)]
		struct A {
			a: i32,
			#[serde(flatten)]
			b: B,
		}
		#[derive(Debug, PartialEq, Serialize)]
		struct B {
			#[serde(rename = "@b")]
			b: i32,
		}

		let a = A {
			a: 1,
			b: B { b: 2 },
		};
		let output = quick_xml::se::to_string(&a).unwrap();
		assert_eq!(output, r#"<A a="1" b="2"/>"#);
	}
```

Mingun:
This is a fundamental flaw of serde flatten design and can be fixed only in serde. Duplicate of #714

Dekker1:
Thank you for the information, and the quick response. Fingers crossed the serde issue will get resolved at some point.

Mingun:
Actually, that is slight different problem, but also due to the flatten design and unavoidable without changes in serde. In you case error is `Unsupported("cannot serialize map without defined root tag")`. That message say that you try to serialize map which does not have a type name, so you must provide the one. [Map](https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_map) serialization instead of [struct](https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_struct) serialization is used because your struct contains flatten fields and therefore the overall set of fields for serialization is unknow at syntax level (at which proc macro works).

With two changes in you struct the serialization is work as expected, but deserialization will not work due to mentioned design flaw:
```rust

#[test]
fn issue761() {
    #[derive(Debug, PartialEq, Serialize)]
    struct A {
        #[serde(rename = "@a")]// <<<< Add (you want attribute)
        a: i32,
        #[serde(flatten)]
        b: B,
    }
    #[derive(Debug, PartialEq, Serialize)]
    struct B {
        #[serde(rename = "@b")]
        b: i32,
    }

    let a = A {
        a: 1,
        b: B { b: 2 },
    };
    let output = quick_xml::se::to_string_with_root("A", &a).unwrap();// <<<< Change
[...]

PR #879, opened 2025-07-16
Feat: Flatten Attributes

Syndelis:
## About this PR

This PR introduces a new special value (akin to `$value` and `$text`) that is supposed to work exclusively with attributes, aptly named `$attributes`. This allows users to flatten structs of attributes into other structs.

## Rationale

I have an use case in which 2 distinct APIs are implementing the same standard. Unfortunately, the standard is not a whole lot strict, and both APIs are in disagreement regarding a couple dozen of attributes, both in types and in names. Both APIs implement a standard from 2007, so I'm in no luck asking either to change their current behavior.

Now, this could be circumvented by:
1. Making attributes present in one API but not the other optional;
2. Wrapping attributes with different type expectancies around untagged enumerators;

However, I'd argue that's not really an elegant solution, and judging by the likes of #761, there is case for flattening attributes. This implementation solves the described issue by allowing generic types to be used and flattened to provide the attributes in either API's expected names and types, minimizing code duplication.

### Alternatives

- `#[serde(flatten)]` can be used, so long as the tag with the flattened field is not the root tag, at which point you'd need to provide the root tag's name explicitly by using `to_string_with_root`, as described by @Mingun [in this response](https://github.com/tafia/quick-xml/issues/761#issuecomment-2172694270) to #761.
  - This does not make it easier in contexts that deal with generic structs. Imagine a wrapper struct that is used in all API requests
- My other PR #878, which adds support for `$value`-named struct fields, would serialize/deserialize the attributes from the "flattened" struct into a new element with the name matching that of the "flattened" struct, not solving the issue described.

## Closing thoughts

This PR lacks a lot of documentation regarding this new feature. I will add the necessary documentation, but I'd like to get reviews on the approach and whether this feature would be accepted before investing time into documenting it.

Examples of this feature's introduced behavior can be found in the tests in `serde-de.rs` and `serde-se.rs`


Mingun:
I prefer to avoid introducing new special renames, and in this case your needs may be solved by regular `#[serde(flatten)]` attribute. Actually, `#[serde(flatten)]` is broken, so you need to implement flattening using other approaches. One of them is to use alternative derive macro which will correctly do flattening or implement it manually. Probably I publish [my implementation](https://github.com/serde-rs/serde/compare/master...Mingun:serde:flatten) on crates.io soon (well, I hope it this year). It is not finished yet, because it was planned as PR to the serde and I tried to not introduce breaking changes (except requiring to implement new `FromFlatten` trait for flattenable types), but serde maintenance very poor at the moment, so merging it is unlikely. So I think, I may just publish my implementation of `serde_derive` crate, where I will not need to solve the complex problem of maintaining full backward compatibility (currently I have not solution for flattening of untagged enums).

Syndelis:
> I prefer to avoid introducing new special renames

I can understand not wanting to introduce new special renames.

> and in this case your needs may be solved by regular `#[serde(flatten)]` attribute. Actually, `#[serde(flatten)]` is broken, so you need to implement flattening using other approaches. One of them is to use alternative derive macro [...] It is not finished yet [...]

So your suggestion is to do a manual implementation, as you point to `serde(flatten)` being broken and your alternative derive macro won't be accepted by serde and is currently unfinished to be released as an individual crate.

With all due respect that honestly doesn't look like an ideal solution, as manually implementing `Serialize`/`Deserialize` is both time consuming and subject to issues when updating struct fields, types, etc. leading to an increase in errors caught at runtime. I'd argue that is less than ideal.

Could you suggest an alternative to the special renaming that could be introduced to this crate?

Mingun:
I do not want to introduce any renamings to this crate, because the problem in the flatten deriving. It should be solved either by serde, or (more realistically) by another derive (+runtime for it).

Most preparation work for this derive I've already [done](https://github.com/serde-rs/serde/compare/master...Mingun:serde:flatten): `FromFlatten` trait and proof-of-concept derivation for
- `struct`s
- externally tagged enums
- adjacently tagged enums

Internally tagged and untagged enums requires more work. I don't mind if someone takes my job from the [mentioned branch](https://github.com/serde-rs/serde/compare/master...Mingun:serde:flatten) and makes alternative `serde_derive` crate (if no one does, I will do that someday).

Actually, two new crates are required:
- runtime crate which introduces `FromFlatten` trait and its implementation for basic types
- derive crate that implements it for user types
- optionally new `Deserialize` derive, that will use `FromFlatten` implementation. Otherwise this small part can be written manually

I understand, that in short-term this solution more resource intensive for you, but in long-term it is better for maintaining this library. So I try to reduce my work too :innocent:

Syndelis:
@Mingun 

> I do not want to introduce any renamings to this crate, because the problem in the flatten deriving. It should be solved either by serde

While I agree this should be solved in `serde`, you probably know it better than me that this is not getting fixed, considering the current maintenance status of the crate. 

> or (more realistically) by another derive (+runtime for it).

Unfortunately, what you described would be unfeasible for my use case. 

> I understand, that in short-term this solution more resource intensive for you, but in long-term it is better for maintaining this library. So I try to reduce my work too 😇

Of course, I completely understand where you're coming from! Having read through the whole `serde` thread, I understand that's more or less what they're doing too, thus causing this situation where the maintainers have no need for the new features, and, thus, do not want to spend time maintaining them. No judgements here!

I do still believe this is a valuable feature for the `quick-xml` crate, and that it blends nicely with other features (such as `$value` and `$text`). If that would make you reconsider adding the feature to `quick-xml`, I volunteer myself to maintaining anything around it, dealing with bug reports around `$attributes` and even expanding support to other use cases if anyone's interested in it. After all, this is a feature I want to use at work, so it sounds fair that I'd invest time in maintaining it. Let me know what you think.

Issue #896, opened 2025-08-30
Lingering serde whitespace bug


jamwil:
Thanks to the work of @ggodlewski, whitespace is now preserved by default according to the xml specification when deserializing with serde. I note, however, one lingering apparent inconsistency in the case of deserializing a string consisting of only whitespace.

I've isolated the issue in [these contrasting passing and failing tests](https://github.com/tafia/quick-xml/compare/master...jamwil:quick-xml:space-issue).

Is this intended behaviour?

ggodlewski:
The reason is in here: https://github.com/tafia/quick-xml/blob/master/src/de/mod.rs#L2792

Check the implementation of `is_blank` https://github.com/tafia/quick-xml/blob/master/src/de/mod.rs#L2253

I think it's intended: https://github.com/tafia/quick-xml/blob/master/src/de/mod.rs#L3368

jamwil:
Thanks for that context. So if I'm understanding correctly, this is actually due to the `xml:space` attribute not yet being implemented?

My interpretation of the xml spec (and I could be misreading this) from section [2.10 White Space Handling](https://www.w3.org/TR/REC-xml/REC-xml-20081126-review.html#sec-white-space), is that an xml processor _must_ pass the whitespace through to the application, and the application gets to decide how to handle the `xml:space` attribute. If quick-xml is a processor, shouldn't the determination of what is significant or insignificant whitespace be done downstream?

Looping in @Mingun since he authored https://github.com/tafia/quick-xml/commit/2b85835968764820d48002dc75789cdd7a0bbe42 which handles this case.

Mingun:
> My interpretation of the xml spec (and I could be misreading this) from section [2.10 White Space Handling](https://www.w3.org/TR/REC-xml/REC-xml-20081126-review.html#sec-white-space), is that an xml processor _must_ pass the whitespace through to the application, and the application gets to decide how to handle the `xml:space` attribute.

Yes, you're absolutely right. `Reader` is an "xml processor", but `Deserializer` is an "application". Currently there are no ways to configure some aspects of it, including handling of whitespace-only text events. Handling of `xml:space` is one of such tasks (part of #464).

jamwil:
@Mingun What do you think about the potential solution in #897? It assumes a text event that consists exclusively of whitespace **and** contains a line break is insignificant. A whitespace event with no line breaks is treated as significant and preserved.

All existing tests pass without modification using this heuristic.

jamwil:
[...]

PR #897, opened 2025-08-31
de: Preserve whitespace for all nodes containg a text event

jamwil:
Potential solution (or intermediate solution) to #896.

### DIFF
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -273,7 +273,14 @@ where
                 QNameDeserializer::from_attr(QName(&slice[key]), decoder, &mut self.de.key_buf)?;
             seed.deserialize(de).map(Some)
         } else {
-            self.skip_whitespaces()?;
+            // If we have dedicated "$text" field, whitespace-only text may be significant.
+            // That also means, that type with `$text` fields behaves as if its element has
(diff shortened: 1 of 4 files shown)


Mingun:
No, I think, this is bad decision. We already have poor designed `make` utility, where spaces handled differently than tabs and that counterintuitive. I think, the [originally suggested](https://github.com/tafia/quick-xml/issues/896#issuecomment-3240039477) implementation, which may be enabled by setting, would be better.

jamwil:
I'm fine with the original suggestion as it certainly will satisfy my requirements, I only propose this route as it accomplishes the same thing but is not destructive in the way you indicated. I do appreciate that broken but predictable is a better position to be in than using opinionated heuristics like this.

jamwil:
Updated as requested.

dralley:
@Mingun Do you plan to make those changes?

dralley:
If this is close to completion already, it would be nice to get it into the next release.

Along with (to whatever degree they are separate and not overlapping issues) https://github.com/tafia/quick-xml/issues/896 and https://github.com/tafia/quick-xml/issues/900 and https://github.com/tafia/quick-xml/issues/984

dralley:
@Mingun Would you be able to either make the changes, or describe the changes that you would like to see made?

PR #899, opened 2025-09-15
Chore: cleanup clippy lints

dsgallups:
I noticed that some of the tests weren't updated. Specifically this line:

```rust
        serialize_as!(struct_: Struct { key: "answer", val: (42, 42) }
```
```console
missing structure fields:
- ser
- children
- write_indent
(rust-analyzer E0063)
```
This value may be found in many places, not sure what to do!

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -2812,6 +2812,7 @@ where
     fn skip(&mut self) -> Result<(), DeError> {
         let event = self.next()?;
         self.skip_event(event)?;
+        // Skip all subtree, if we skip a start event
         if let Some(DeEvent::Start(e)) = self.write.back() {
             let end = e.name().as_ref().to_owned();
             let mut depth = 0;


dralley (approved):

src/de/mod.rs
```diff
@@ -2815,35 +2812,32 @@ where
     fn skip(&mut self) -> Result<(), DeError> {
         let event = self.next()?;
         self.skip_event(event)?;
-        match self.write.back() {
-            // Skip all subtree, if we skip a start event
```
dralley, line 2819: Why did this comment disappear?

dralley (approved):

dsgallups:
note: the original comment at the top appears not to be caught by the checks. Not sure if that's something to consider..I don't quite know how to fix it, but it appears those structures are a bit out of date at this point.

src/de/key.rs
```diff
@@ -155,10 +155,8 @@ impl<'de, 'd> Deserializer<'de> for QNameDeserializer<'de, 'd> {
     deserialize_num!(deserialize_u32, visit_u32);
     deserialize_num!(deserialize_u64, visit_u64);
 
-    serde_if_integer128! {
-        deserialize_num!(deserialize_i128, visit_i128);
-        deserialize_num!(deserialize_u128, visit_u128);
-    }
+    deserialize_num!(deserialize_i128, visit_i128);
+    deserialize_num!(deserialize_u128, visit_u128);
```
Mingun: This change will be done in #908 with necessary bump of the dependency version. It is not required here, or you may rebase your PR over #908

src/encoding.rs
```diff
@@ -257,7 +257,7 @@ pub fn detect_encoding(bytes: &[u8]) -> Option<(&'static Encoding, usize)> {
         // without BOM
         _ if bytes.starts_with(&[0x00, b'<', 0x00, b'?']) => Some((UTF_16BE, 0)), // Some BE encoding, for example, UTF-16 or ISO-10646-UCS-2
         _ if bytes.starts_with(&[b'<', 0x00, b'?', 0x00]) => Some((UTF_16LE, 0)), // Some LE encoding, for example, UTF-16 or ISO-10646-UCS-2
-        _ if bytes.starts_with(&[b'<', b'?', b'x', b'm']) => Some((UTF_8, 0)), // Some ASCII compatible
+        _ if bytes.starts_with(b"<?xm") => Some((UTF_8, 0)), // Some ASCII compatible
```
Mingun: I prefer to leave it in its current form to keep nice formatting

src/escape.rs
```diff
@@ -2181,7 +2181,7 @@ mod normalization {
                     if ch == '\u{0085}' {
                         assert_eq!(normalize_xml11_eols(input), "\n", "{}", description);
                     } else {
-                        let mut expected = utf8.clone();
+                        let mut expected = utf8;
```
Mingun: That clone is intentional. We do not want to change `utf8` content, because that will change the `input`. I even doubt if that could be compiled.

PR #901, opened 2025-09-19
fix tests on 32-bit architecture

alexanderkjall:
as 42000000000000 is outside the valid range of usize/isize on 32-bit machines


Mingun:
Just remove excess zeroes instead of feature-gate tests.

PR #904, opened 2025-09-25
Implement serializing CDATA

### LINKED ISSUE #353: Serializing CDATA

Ninja3047:
Closes: #353

### DIFF
--- a/src/se/simple_type.rs
+++ b/src/se/simple_type.rs
@@ -3,15 +3,14 @@
 //! [simple types]: https://www.w3schools.com/xml/el_simpletype.asp
 //! [as defined]: https://www.w3.org/TR/xmlschema11-1/#Simple_Type_Definition
 
-use crate::escape::_escape;
+use crate::escape::escape_char;
 use crate::se::{QuoteLevel, SeError};
 use serde::ser::{
     Impossible, Serialize, SerializeSeq, SerializeTuple, SerializeTupleStruct,
     SerializeTupleVariant, Serializer,
 };
 use serde::serde_if_integer128;
-use std::borrow::Cow;
-use std::fmt::Write;
+use std::fmt::{self, Write};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum QuoteTarget {
@@ -23,14 +22,37 @@ pub enum QuoteTarget {
     SingleQAttr,
 }
 
+fn escape_into<W, F>(mut writer: W, value: &str, escape_chars: F) -> fmt::Result
+where
+    W: Write,
+    F: Fn(u8) -> bool,
+{
+    let bytes = value.as_bytes();
+    let mut iter = bytes.iter();
+    let mut pos = 0;
+    while let Some(i) = iter.position(|&b| escape_chars(b)) {
+        let new_pos = pos + i;
+        escape_char(&mut writer, value, pos, new_pos)?;
+        pos = new_pos + 1;
+    }
+
+    if let Some(raw) = value.get(pos..) {
+        writer.write_str(raw)?;
+    }
+    Ok(())
+}
+
 /// Escapes atomic value that could be part of a `xs:list`. All whitespace characters
 /// additionally escaped
-fn escape_item(value: &str, target: QuoteTarget, level: QuoteLevel) -> Cow<'_, str> {
+fn escape_item<W>(writer: W, value: &str, target: QuoteTarget, level: QuoteLevel) -> fmt::Result
+where
+    W: Write,
+{
     use QuoteLevel::*;
     use QuoteTarget::*;
 
     match (target, level) {
-        (_, Full) => _escape(value, |ch| match ch {
+        (_, Full) => escape_into(writer, value, |ch| match ch {
             // Spaces used as delimiters of list items, cannot be used in the item
             b' ' | b'\r' | b'\n' | b'\t' => true,
             // Required characters to escape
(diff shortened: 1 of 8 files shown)


Mingun (changes requested):
Looks good, but need to check how lists will be serialized and would we able to deserialize them back.

src/se/simple_type.rs
```diff
@@ -93,6 +102,37 @@ fn escape_item(value: &str, target: QuoteTarget, level: QuoteLevel) -> Cow<'_, s
     }
 }
 
+/// Writes content as CDATA section, handling the `]]>` sequence by splitting into multiple CDATA sections
+fn write_cdata_content<W: Write>(writer: &mut W, value: &str) -> Result<(), SeError> {
+    if value.is_empty() {
+        writer.write_str("<![CDATA[]]>")?;
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
Ninja3047: mostly for trying to match behavior with C# which apparently does emit empty CDATA sections https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlwriter.writecdata?view=net-9.0#remarks

it _probably_ won't break if i leave them out. not sure how common it is for other lanaguages and the xml specification doesn't seem to say whether or not they should be there but it does make it so i can't roundtrip with the same format which is annoying

src/se/simple_type.rs
```diff
@@ -1219,4 +1288,44 @@ mod tests {
...
+        serialize_cdata_as!(with_cdata_end: "foo]]>bar" => "<![CDATA[foo]]]]><![CDATA[>bar]]>");
+        serialize_cdata_as!(multiple_cdata_ends: "a]]>b]]>c" => "<![CDATA[a]]]]><![CDATA[>b]]]]><![CDATA[>c]]>");
+        serialize_cdata_as!(starts_with_cdata_end: "]]>hello" => "<![CDATA[]]]]><![CDATA[>hello]]>");
+        serialize_cdata_as!(ends_with_cdata_end: "hello]]>" => "<![CDATA[hello]]]]><![CDATA[>]]>");
+        serialize_cdata_as!(only_cdata_end: "]]>" => "<![CDATA[]]]]><![CDATA[>]]>");
+
+        serialize_cdata_as!(seq_basic: vec!["foo", "bar", "baz"] => "<![CDATA[foo]]> <![CDATA[bar]]> <![CDATA[baz]]>");
+        serialize_cdata_as!(seq_with_markup_chars: vec!["<tag>", "&entity", "\"quoted\"", "with space"] => "<![CDATA[<tag>]]> <![CDATA[&entity]]> <![CDATA[\"quoted\"]]> <![CDATA[with space]]>");
+        serialize_cdata_as!(seq_with_cdata_end_split: vec!["foo]]>bar", "test"] => "<![CDATA[foo]]]]><![CDATA[>bar]]> <![CDATA[test]]>");
+        serialize_cdata_as!(tuple_cdata: ("first", 42, "third") => "<![CDATA[first]]> <![CDATA[42]]> <![CDATA[third]]>");
```
Mingun, line 1329: It would be better, if this serialized as one CDATA section. Probably, that is possible?

Ninja3047: yeah agreed, need to figure out exactly how to do that nicely with serde

JeanMertz:
Great to see this being merged soon 👍 

I gave this implementation a try. One thing that confused me was that the example implies that the `CDATA` wrapper is only applied to fields that are marked with `#[serde(rename = "$text")]`, but I had other `String` fields that also got the wrapper, even though they didn't have the `$text` attribute applied to them. Is this a bug, or working as expected?

JeanMertz:
Also, I think the following test case should be added:

```
baz ]] baz 
```

As I think(?) it generates incorrect XML:

```
<![CDATA[baz ]] baz ]]>
```

JeanMertz:
> I think not. Imagine, if you want to store HTML in the `String` field.

Well, then you are free to mark it as `$text`, but if you want to store a string that you _know_ is safe, this PR does not give you a way to serialize that string without the CDATA tags.

I'm not saying it's a blocker, I was mostly curious why it isn't only applied to `$text` fields.

Mingun:
@Ninja3047, thanks on working on this! I finally polish the PR and make it much more simple. The release with the change will be soon

PR #907, opened 2025-10-22
Fix serialization of structs with only text content as attributes

### LINKED ISSUE #906: struct-based list can be serde-serialized as child but not as attribute
Hi, cool project!

Consider these two types using serde (de)serialize:
```rs

#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct FooType {
    #[serde(rename = "a-list")]
    pub a_list: ListType,
}

#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
pub struct ListType{
    #[serde(default, rename = "$text")]
    pub content: Vec<String>
}
```

This creates xml such as `<test><a-list>A B</a-list></test>`.

Now consider this slight adaptation:

```rs
#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct BarType {
    #[serde(default = "rename = "@a-list")]
    pub a_list: ListType,
}
``` 

I expect this to create xml such as `<test a-list="A B"/>`, but it fails with 
> Unsupported("cannot serialize struct `ListType` as an attribute or text content value")

I know that if I change `ListType` to simply

```rs
type ListType = Vec<String>
```

the second case works. But then, the first case does not serialize correctly: `<test><a-list>A</a-list><a-list>B</a-list></test>`

Is there a way to define a single `ListType` in such a way that both 
- `<test><a-list>A B</a-list></test>`
- `<test a-list="A B"/>`

can be serialized?
Thanks a lot!!

Kakadus:
close #906 

allows structs that contain only one $text field to be serialized as an attribute list.

As this is my first contribution, I may have overlooked something. The tests seem to pass though :)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,9 @@
 
 ### New Features
 
+- [#907]: Allow to serialize and deserialize structs which contains only one `$text` field
+  as an attribute value or a text content value.
+
 ### Bug Fixes
 
 ### Misc Changes
@@ -25,6 +28,7 @@
   of `NsReader`. Use `.resolver().bindings()` and `.resolver().resolve()` methods instead.
 - [#913]: `Attributes::has_nil` now accepts `NamespaceResolver` instead of `Reader<R>`.
 
+[#907]: https://github.com/tafia/quick-xml/pull/907
 [#908]: https://github.com/tafia/quick-xml/pull/908
 [#913]: https://github.com/tafia/quick-xml/pull/913
 
--- a/src/de/simple_type.rs
+++ b/src/de/simple_type.rs
@@ -3,13 +3,13 @@
 //! [simple types]: https://www.w3schools.com/xml/el_simpletype.asp
 //! [as defined]: https://www.w3.org/TR/xmlschema11-1/#Simple_Type_Definition
 
-use crate::de::Text;
+use crate::de::{Text, TEXT_KEY};
 use crate::encoding::Decoder;
 use crate::errors::serialize::DeError;
 use crate::escape::unescape;
 use crate::utils::{trim_xml_spaces, CowRef};
 use memchr::memchr;
-use serde::de::value::UnitDeserializer;
+use serde::de::value::{MapDeserializer, SeqAccessDeserializer, UnitDeserializer};
 use serde::de::{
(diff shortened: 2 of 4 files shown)


Mingun:
I think, that makes sense.

Kakadus:
Thanks for taking a look! It seems that I simply forgot about deserialization :)

Writing custom (de)serializers always work, although I'd really like to see first party support for all usages of the text structs. Each struct would need their own implementation, and you'd have to track their usage and apply the with appropriately. Coming from code generators like https://docs.rs/xsd-parser/latest/xsd_parser/, this makes it very complicated IMHO. What do you think about the `SimpleTypeDeserializer` extension I commited instead?

> As you known, some serde attributes may change struct representation to map representation and therefore we will lose our hack here, because maps does not report their keys.

Do you mean e.g. `#[serde(flatten)]`? A short experiment shows that deserialization of `AsElement` fails already (which is not affected by this PR):

```rs
        /// Regression test for https://github.com/tafia/quick-xml/issues/906.
        #[test]
        fn issue906() {
            #[derive(Debug, PartialEq, Deserialize, Serialize)]
            struct AsElement {
                #[serde(rename = "a-list")]
                a_list: TextContent,
            }

            #[derive(Debug, PartialEq, Deserialize, Serialize)]
            pub struct TextContent {
                #[serde(flatten)]
                content: Inner,
            }

            #[derive(Debug, PartialEq, Deserialize, Serialize)]
            pub struct Inner {
                #[serde(default, rename = "$text")]
                content: Vec<String>,
            }

            let foo = AsElement {
                a_list: TextContent {
                    content: Inner {
                        content: vec!["A".to_string(), "B".to_string()],
                    },
                },
            };

            let buffer = to_string_with_root("test", &foo).unwrap();
            assert_eq!(buffer, "<test><a-list>A B</a-list></test>");
            // thread 'with_root::list::issue906' panicked at tests/serde-se.rs:2347:53:
            // called `Result::unwrap()` on an `Err` value: Custom("invalid type: string \"A B\", expected a sequence")
            let foo2: AsElement = from_str(&buffer).unwrap();
            assert_eq!(foo2, foo);
        }
```

Kakadus:
build error with old serde version should be fixed now

Kakadus:
@Mingun implementing IntoDeserializer for ListIter directly allows this to compile with serde<1.0.214

PR #909, opened 2025-10-29
Avoid allocations in Serializer when write numbers and characters

Mingun:


src/se/simple_type.rs
```diff
@@ -460,9 +464,14 @@ pub struct SimpleTypeSerializer<W: Write> {
 }
 
 impl<W: Write> SimpleTypeSerializer<W> {
+    #[inline]
```
dralley: I'm pretty sure these `#[inline]` annotations do not actually accomplish anything.  All it does is make the function bodies available to dependent crates for optimization purposes.  But all of this code is internal to the crate anyway, and hence inline-able, and the compiler doesn't care about the `#[inline]` annotation when deciding whether or not to inline code within a compilation unit.

dralley (approved):

src/se/simple_type.rs
dralley: I've heard otherwise, but it's possible I'm wrong or that things have changed since I read about that.

PR #913, opened 2025-11-14
Improve API of `Attributes::has_nil` and deprecate some methods for namespace resolution

Mingun:


src/reader/ns_reader.rs
```diff
@@ -130,6 +130,7 @@ impl<R> NsReader<R> {
     /// # quick_xml::Result::Ok(())
     /// ```
     #[inline]
+    #[deprecated = "Use `.resolver().bindings()` instead"]
```
dralley: If you have a concrete idea of which release it will be removed in completely, it might be good to list that.  Up to you.

However either there should be an issue filed to remove the deprecated bits.

Issue #916, opened 2025-12-10
Split SyntaxError::UnclosedPIOrXmlDecl into 2 separate variants


SuchAFuriousDeath:
The SyntaxError enum does not currently differentiate between and unclosed processing instruction and an unclosed xml tag: 
https://github.com/tafia/quick-xml/blob/646ce4b866ee574f9f7e85de1496eea519ac2166/src/errors.rs#L22

I was trying to make some tests pass in [Ruffle](https://github.com/ruffle-rs/ruffle) and it turns out that in order for the tests to pass, we need to differentiate between those 2 errors and throw different errors based on which one occured. 

I was wondering whether it would be acceptable to split the SyntaxError::UnclosedPIOrXmlDecl variant into 2 separate variants. This would of course be a breaking change. I could do a PR if it is deemed acceptable.

Mingun:
Yes, that is acceptable

SuchAFuriousDeath:
Ok, I will do a PR when I find the time, thanks for the quick answer.

PR #924, opened 2025-12-20
Split SyntaxError::UnclosedPIOrXmlDecl into separate enum variants

SuchAFuriousDeath:
Implements #916

The SyntaxError enum is public, so this is a breaking change. 
The Parser trait is also public, so the method signature change is also a breaking change.


Mingun (changes requested):
Probably, that the right way to do that. I haven't looked in detail yet, but I already see a problem with the potential handling of currently unsupported encodings. I would like to see at least `#[ignore]`d tests added for these cases and FIXMEs in problematic lines.

Also, such change needs changelog entry. I prefer if you rebase your branch over current master instead of merging (that is needed to prevent merge conflict in `Changelog.md`).

Issue #923, opened 2025-12-20
`Reader::read_event_into` panics because of incorrect DTD handling


tayu0110:
Hello, I would like to report an issue because of incorrect DTD handling.

This is a well-formed and valid XML document.
```xml
<?xml version="1.0"?>
<!-- sample.xml -->
<!DOCTYPE root [
        <!ENTITY ent ">">
        <!ELEMENT root (#PCDATA)>
]>
<root>&ent;</root>
```

For the above XML document, write the following simple code,
```rust
use quick_xml::{Reader, events::Event};

fn main() {
    let path = std::env::args().nth(1).unwrap();
    let mut reader = Reader::from_file(path).unwrap();
    let mut buf = vec![];
    while reader.read_event_into(&mut buf).unwrap() != Event::Eof {}
}
```

and run it
```
cargo run -- sample.xml
```

As a result, `Reader::read_event_into` panics with the following message.
```
thread 'main' (427815) panicked at src/main.rs:7:44:
called `Result::unwrap()` on an `Err` value: Syntax(InvalidBangMarkup)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Furthermore, the following XML document is also well-formed.
```xml
<?xml version="1.0"?>
<!-- sample.xml -->
<!DOCTYPE root [
        <!ENTITY ent "<">
]>
<root />
```

The execution results for this are as follows.
```
thread 'main' (435643) panicked at src/main.rs:7:44:
called `Result::unwrap()` on an `Err` value: Syntax(UnclosedDoctype)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

According to the XML specification, including `<` and `>` in an entity value literal is entirely legal. Merely checking the correspondence between `<` and `>` is insufficient to skip over the DTD.  
In this sense, the current implementation's DTD handling is incorrect.

PR #925, opened 2025-12-21
Dtd handling correction for issue #923

tayu0110:
Hello, I have implemented the fix for issue #923 and would like to merge it into the master branch.

The PR primarily consists of fixes to `reader::BangType::DocType` and `reader::BangType::parse`, along with the addition of a small test to `tests/issues.rs`.

I believe these fixes will at least enable correct skipping of well-formed DTDs.

While adding checks for ill-formed DTDs might not be impossible, the current implementation lacks a mechanism for `reader::BangType::parse` to report errors. Since this would require significant changes, this update focuses solely on skipping well-formed DTD internal subsets without errors.

Thank you for your confirmation.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -27,6 +27,8 @@ struct and can be applied at once. When `serde-types` feature is enabled, config
 
 ### Bug Fixes
 
+- [#923]: Implement correct skipping of well-formed DTD.
+
 ### Misc Changes
 
 - [#908]: Increase minimal supported `serde` version from 1.0.139 to 1.0.180.
@@ -41,6 +43,7 @@ struct and can be applied at once. When `serde-types` feature is enabled, config
 [#908]: https://github.com/tafia/quick-xml/pull/908
 [#913]: https://github.com/tafia/quick-xml/pull/913
 [#924]: https://github.com/tafia/quick-xml/pull/924
+[#923]: https://github.com/tafia/quick-xml/issues/923
 
 
 ## 0.38.4 -- 2025-11-11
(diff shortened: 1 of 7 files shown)


Mingun:
I squashed all commits into the one and slightly change main `match` by matching by `Some(...)` and `None` in main `match` instead of inside the arms. I think, that this improves readability. Still reviewing...

What about your initial commit 96dba4d52859d6dd2c38ee061638a45b5117f451? It is fine to bump dependency version if you need some specific features (especially, if that will not increase our MSRV, but we also can bump it if needed)

adamreichold:
> If there's no significant performance difference, I think it's fine to leave it at the latest commit. (Since it's at most 3-character search, I don't expect a huge difference.)

Note that `memmem` has significant setup costs compared to `memchr` and might end up a pessimization (especially w.r.t. code density) if used in cold paths. Over at [`robinson`](https://github.com/adamreichold/robinson/blob/669c85bd663d11be344bc025a3961b84dbe14f60/src/memchr.rs#L51), I even found that matching substrings using the naive approach based on `memchr` for a single byte and looping was actually faster than `memmem` for the XML files I am interested in.

(`memchr`, particularly as implemented in the `memchr` crate is really optimized for finding rare needles large haystacks which is where SIMD is most effective. If a subparser is expected to only move forward a few bytes, scalar code using `match` which compilers can often implement using highly efficient bit test instruction can be more efficient than SIMD.)

Mingun:
At first I wanted to introduce a new kind of error to indicate that an unknown markup was found in the DTD, but I decided to postpone this until we would not just skip the DTD, but return its contents in the form of events (since we still parse it, so why not). Otherwise, it is difficult to point to a specific position where the error actually occured.

Mingun:
> Is there a possibility that support for reporting markup within DTDs as events will be added in the future?

Yes

PR #933, opened 2026-01-25
Add compare sync async benchmark

mohe2015:
It seems like the async implementation is much slower that the synchronous implementation. As a first start, add an benchmark to more easily compare them and be able to track improvements / regressions.

My actual usage does not read from static data but I would expect that reading from static data should have very similar performance for sync and async.

```
cargo bench --features async-tokio --bench compare_sync_async
   Compiling quick-xml v0.39.0 (/home/moritz/Documents/quick-xml)
    Finished `bench` profile [optimized + debuginfo] target(s) in 30.10s
     Running benches/compare_sync_async.rs (target/release/deps/compare_sync_async-ded794078866a808)
Gnuplot not found, using plotters backend
compare_sync/sample_rss.xml
                        time:   [103.25 µs 103.51 µs 103.77 µs]
                        thrpt:  [1.7749 GiB/s 1.7794 GiB/s 1.7839 GiB/s]
                 change:
                        time:   [−0.5326% −0.2991% −0.0553%] (p = 0.02 < 0.05)
                        thrpt:  [+0.0553% +0.2999% +0.5355%]
                        Change within noise threshold.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

compare_async/sample_rss.xml
                        time:   [176.42 µs 176.71 µs 177.05 µs]
                        thrpt:  [1.0403 GiB/s 1.0423 GiB/s 1.0440 GiB/s]
                 change:
                        time:   [−0.1489% +1.0627% +2.0229%] (p = 0.05 < 0.05)
                        thrpt:  [−1.9828% −1.0515% +0.1491%]
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -57,7 +57,7 @@ jobs:
     - name: Build
       run: cargo build
     - name: Build benchmarks
-      run: cargo bench --no-run
+      run: cargo bench --features async-tokio --no-run
     - name: Build benchmarks (compare)
       working-directory: compare
       run: cargo bench --no-run
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ regex = "1"
 # serde does not follow semver in numbering and their dependencies, so we specifying patch here
 serde_derive = { version = "1.0.206" }
 serde-value = "0.7"
-tokio = { version = "1.21", default-features = false, features = ["macros", "rt", "rt-multi-thread"] }
+tokio = { version = "1.21", default-features = false, features = ["macros", "rt"] }
 tokio-test = "0.4"
 
 [lib]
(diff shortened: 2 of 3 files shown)


Mingun (changes requested):
> that reading from static data should have very similar performance for sync and async.

async have some runtime costs to handle all the machinery and if data is static (i.e no I/O) it is expected that it would be _slower_ than sync interface. async is not about increasing performance of an individual task, but about using less threads to process more I/O bound tasks.

tafia:
> I still think, that it would be better to place this benchmark into compare project.

I think it makes sense to be in the core crate because it is comparing against itself.

> async have some runtime costs to handle all the machinery and if data is static (i.e no I/O) it is expected that it would be slower than sync interface. async is not about increasing performance of an individual task, but about using less threads to process more I/O bound tasks.

I agree with @Mingun. I am not sure how useful that benchmark is. I have a feeling that you are benchmarking async tokio vs std, not really quick-xml. Except if you could add some realistic scenario where async is preferable (e.g. loading multiple files large in parallel with a single CPU etc ...)

PR #935, opened 2026-02-12
Change parser state only after .await points to prevent state corruption

Mingun:
See explanation in https://github.com/tafia/quick-xml/issues/624#issuecomment-3892689132

Unfortunately, I cannot create stable test for this, so this bugfix without tests.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,7 +18,11 @@
 
 ### Bug Fixes
 
+- [#624]: Fix corruption of parser state after cancelling of async reading.
+
 ### Misc Changes
+
+[#624]: https://github.com/tafia/quick-xml/issues/624
 
 
 ## 0.39.0 -- 2026-01-11
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -405,10 +405,15 @@ macro_rules! read_until_close {
         $reader:expr
         $(, $await:ident)?
     ) => {{
-        $self.state.state = ParseState::InsideText;
-
         let start = $self.state.offset;
-        match $reader.peek_one() $(.$await)? {
+        // It is important to change state only after .await point, because future
+        // may be cancelled only at .await points
+        //
+        // See https://github.com/tafia/quick-xml/issues/624
+        let lookahead = $reader.peek_one() $(.$await)?;
+
+        $self.state.state = ParseState::InsideText;
+        match lookahead {
             // `<!` - comment, CDATA or DOCTYPE declaration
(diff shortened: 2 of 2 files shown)


dralley (approved):

Mingun:
I don't quite understand the question. `async` is needed to wait for data from somewhere, when parsing we do not need to await anything, except the data, but getting data is independent of its parsing. For example, when we parse from slice we already have all needed data in the memory.

adamreichold:
> It is still unclear for me how the async may help with managing parse states. Could you provide, for example, how parsing of `<tag attribute1=">" attribute2='<'>` could look? We need 3 states here:
> 
>     * Outside -- outside of any quoted region
> 
>     * DoubleQ -- inside double-quoted region
> 
>     * SingleQ -- inside single-quoted region

Instead of a signature like

```rust
fn feed(&mut self, bytes: &[u8]) -> Option<usize>;
```

which tracks state via `self`, you'd track state via control flow, e.g.

```rust
async fn parse_quoted(read: &mut AsyncBufRead) -> Result<Vec<u8>> {
    let quote = expect_oneof(read, [b'\'', b'"']).await?;

    let mut buf = Vec::new();
    read.read_until(quote, &mut buf).await?;

    Ok(buf)
}
```

where the compiler will effectively create that state enum for you when it prepares the future to suspend after each suspension/await point.

This is just how I/O-based recursive descent parsers have been written forever and their main upside is that the code flow represents what symbols are expected where and when which is usually simpler for humans to understand than an explicit state machine.

The main trick here would to wrap this using a custom implementation of `AsyncBufRead` and trivial executor, so that one would could present this as a sans I/O interface on the outside, e.g.

```rust
fn feed(&mut self, bytes: &[u8]) -> Result<usize> {
   self.bytes = bytes;

   let mut cx = std::task::Context::from_waker(std::task::Waker::noop());

   match self.parser.as_mut().poll(&cx) {
     Poll::Pending => { Ok(self.consumed) }
     Poll::Ready(Ok(())) => { self.finished = true; Ok(self.consumed) }
     Poll::Ready(Err(err))) => { Err(err) }
   }
}
```

with some suitably constructed future stored in `self.parser` which was passed an `AsyncBufRead` implementation that is essentially backed by `self.bytes` and `self.consumed`. (Yes, there is a bit of hand-waving to safely making the sharing happen behind the scenes here.)

This would turn this around making the parser async-first and then derive the sync version from it, but just as a means to get state-management-via-control-flow known from classic recursive descent parsers.

Issue #939, opened 2026-02-16
UnclosedTag on validated XML file.


LegoWolf:
Starting with v0.39.1 of quick_xml, I get a `SyntaxError::UnclosedTag` when parsing the XML files that I get from Apple Health Kit's export feature.

I've attached a sample export that I've trimmed down as small as I can while still reproducing the error. I shrunk it by removing large chunks of the data tags in between the HealthData tags, then scrubbed it of personally identifying information and checked its validity with XML ValidatorBuddy Plus. You can reproduce the error by creating the following project and placing the attached file [exporter-test.xml](https://github.com/user-attachments/files/25330636/exporter-test.xml) in the same folder as Cargo.toml. 

Cargo.toml:
```toml
[package]
name = "testxml"
version = "0.1.0"
edition = "2024"

[dependencies]
quick-xml = "=0.39.1"
```

main.rs
```rust
use std::error::Error;
use std::fs::File;
use std::io::BufReader;
use quick_xml::events::Event;
use quick_xml::reader::Reader;

const BUFFER_SIZE: usize = 4 * 1024 * 1024;

fn main() -> Result<(), Box<dyn Error>> {
    let xml_file = BufReader::new(File::open("exporter-test.xml")?);
    let mut xml_file = Reader::from_reader(xml_file);
    let mut buf = Vec::with_capacity(BUFFER_SIZE);
    loop {
        match xml_file.read_event_into(&mut buf)? {
            Event::Start(_e) => {},
            Event::Empty(_e) => {},
            Event::End(_e) => {},
            Event::Eof => break,
            _ => { },
        }
        buf.clear();
    }
    Ok(())
}
```

I noticed along the way that the XML line where the failure occurs shifts wildly depending on what seem like innocuous edits to the XML file. e.g. Changing the string length of the value of one attribute moved the error from line ~1700 to line ~4400. I suspect that this error is related to buffer management or something rather than parsing as a result, but I don't know the code well enough to diagnose.

I've tested this example code on Windows and Amazon Linux 2023 with the same result.

Mingun:
Minimal reproducer:
```rust
/// Regression test for https://github.com/tafia/quick-xml/issues/939
#[test]
fn issue939() {
    let xml_file = BufReader::with_capacity(4, &b"<r><t/></r>"[..]);
    let mut reader = Reader::from_reader(xml_file);
    let mut buf = Vec::new();

    assert_eq!(
        reader.read_event_into(&mut buf).unwrap(),
[...]

PR #947, opened 2026-02-22
Provide access to a wrapper that validates that bytes being read are valid UTF-8

dralley:
This is a small stepping stone towards https://github.com/tafia/quick-xml/issues/158


src/errors.rs
```diff
@@ -221,7 +221,12 @@ impl From<IoError> for Error {
     /// Creates a new `Error::Io` from the given error
     #[inline]
     fn from(error: IoError) -> Error {
-        Self::Io(Arc::new(error))
+        match error.kind() {
+            IoErrorKind::InvalidData => Self::Encoding(error.downcast::<EncodingError>().expect(
```
dralley, line 225: Apparently stabilized in 1.79 https://doc.rust-lang.org/std/io/struct.Error.html#method.downcast

So we would need to bump MSRV.

tests/encodings.rs
```diff
@@ -27,7 +29,70 @@ mod decode {
...
 }
+#[cfg(not(feature = "encoding"))]
+mod validate {
+    use super::*;
+    use quick_xml::encoding::{EncodingError, Utf8ValidationError};
+    use quick_xml::errors::Error;
+
+    #[test]
+    fn test_validation_fails_on_utf16le_input() {
+        let src = include_bytes!("documents/encoding/utf16le-bom.xml").as_ref();
```
dralley: I also tried with the KOI-R sample, but it appears that our XML validates under UTF-8.  As do the non-BOM UTF16 samples.  They're just too short and don't contain any incompatible sequences.

So, the next commit in this series will try to enable `detect_encoding` without the `encoding` feature, so that we can throw an error faster if the encoding doesn't match UTF-8.

src/encoding.rs
```diff
@@ -264,3 +318,720 @@ pub fn detect_encoding(bytes: &[u8]) -> Option<(&'static Encoding, usize)> {
...
+        &mut self.inner
+    }
+
+    /// Consumes this reader and returns the underlying reader
+    pub fn into_inner(self) -> R {
+        self.inner
+    }
+}
+
+impl<R: Read> Read for Utf8ValidatingReader<R> {
```
dralley: I need to do more testing and review on this implementation myself as well, I don't trust it yet.

dralley:
@Mingun Do you have any thoughts on the overall design direction or plan of action?

src/reader/mod.rs
```diff
@@ -792,6 +792,35 @@ impl<R> Reader<R> {
...
+    /// use std::io::Cursor;
+    ///
+    /// let data = Cursor::new(b"<tag>content</tag>".to_vec());
+    /// let reader = Reader::from_reader_validating(data);
+    /// // Reader will validate UTF-8 as it reads
+    /// ```
+    ///
+    /// [`Utf8BytesReader`]: crate::encoding::Utf8BytesReader
+    #[cfg(not(feature = "encoding"))]
+    pub fn from_reader_validating(reader: R) -> Self {
```
dralley, line 821: New constructors vs. opt-in feature is an open question, I have no particularly strong feelings.  I leaned towards this route because `Reader::from_reader()` is frequently used with slices, and in that case we would be throwing a buffer into the mix which would require signatures changing, etc... I wanted to avoid changing public API as much as possible for now.

src/encoding.rs
```diff
@@ -264,3 +318,720 @@ pub fn detect_encoding(bytes: &[u8]) -> Option<(&'static Encoding, usize)> {
...
+                            }
+                            // Otherwise fall through to read more data
+                        }
+                    }
+                }
+            }
+
+            // Read more data from the underlying reader
+            let read_size = buf.len().max(64); // Read at least 64 bytes for efficiency
+            let mut temp = vec![0u8; read_size];
```
Mingun: Creating a new `Vec` in the loop in hot path will ruin performance. You already have `self.buffer`, why not read into it? Or at least read into a fixed-size buffer. Or even better, require `BufRead` and get buffers from it.

src/encoding.rs
```diff
@@ -264,3 +318,720 @@ pub fn detect_encoding(bytes: &[u8]) -> Option<(&'static Encoding, usize)> {
...
+    }
+}
+
+#[cfg(test)]
+mod utf8_bytes_reader_tests {
+    use super::*;
+    use std::io::{BufRead, Read};
+
+    #[test]
+    fn test_basic_read() {
```
Mingun: I prefer not to name the testing functions with `test_`, because this is already clear.

Mingun (commented):
It seems to me, rather that implement own decoding stream, we may use [encoding_rs_io](https://docs.rs/encoding_rs_io/latest/encoding_rs_io/) or similar library to do the conversion.

I also just created #948 to show you where we may change the reader from the reader that guessing encoding to the reader that transparently decodes. It seems to me that it is at the level of a slightly higher level `XmlReader` that honest decoding should be introduced.

src/encoding.rs
dralley: Yes, that is correct - as mentioned this is really a proof of concept level implemenation, it needs some iteration before being production ready.  But I can address this in the next push.

src/reader/mod.rs
```diff
@@ -267,7 +267,7 @@ macro_rules! read_event_impl {
                     #[cfg(feature = "encoding")]
                     if let Some(encoding) = $reader.detect_encoding() $(.$await)? ? {
                         if $self.state.encoding.can_be_refined() {
-                            $self.state.encoding = crate::reader::EncodingRef::BomDetected(encoding);
+                            $self.state.encoding = crate::reader::EncodingRef::BomDetected(encoding.encoder());
```
dralley: I'm not sure if this was ever strictly correct, `detect_encoding()` would return `Some((encoder, bom_len)` even when there was no BOM per-se.  So for e.g. an encoding similar to or compatible with UTF-8 that is not UTF-8, I would think this is providing false certainty about BOM presence.

edit: to be clear, if it was wrong before, it's still wrong.

src/encoding.rs
```diff
@@ -300,25 +298,63 @@ pub fn decode_into(
 /// |`00 3C 00 3F`|UTF-16 BE or ISO-10646-UCS-2 BE or similar 16-bit BE (use declared encoding to find the exact one)
 /// |`3C 00 3F 00`|UTF-16 LE or ISO-10646-UCS-2 LE or similar 16-bit LE (use declared encoding to find the exact one)
 /// |`3C 3F 78 6D`|UTF-8, ISO 646, ASCII, some part of ISO 8859, Shift-JIS, EUC, or any other 7-bit, 8-bit, or mixed-width encoding which ensures that the characters of ASCII have their normal positions, width, and values; the actual encoding declaration must be read to detect which of these applies, but since all of these encodings use the same bit patterns for the relevant ASCII characters, the encoding declaration itself may be read reliably
-#[cfg(feature = "encoding")]
-pub fn detect_encoding(bytes: &[u8]) -> Option<(&'static Encoding, usize)> {
+pub fn detect_encoding(bytes: &[u8]) -> Option<DetectedEncoding> {
```
dralley, line 312: Implements https://github.com/tafia/quick-xml/issues/158#issuecomment-2953986471

tests/encodings.rs
dralley: Done

src/reader/state.rs
```diff
@@ -280,6 +280,8 @@ impl ReaderState {
             if content.starts_with(b"xml") && (len == 3 || is_whitespace(content[3])) {
                 let event = BytesDecl::from_start(BytesStart::wrap(content, 3, self.decoder()));
 
+                // TODO: once we can assume that the parser is operating on UTF-8, then we can throw
+                // an error here if we see a non-UTF-8 encoding... if encoding/decoding is not enabled.
```
dralley, line 284: We can't assume this yet because it's still possible to create non-validating readers under `[cfg(not(feature = "encoding"))]

src/encoding.rs
dralley: Done in latest commit

dralley:
@Mingun Ignore for the time being the exact details of the `Utf8ValidatingReader` implementation - I already mentioned it still needs some additional work.

Architecturally, do you have any objections to the direction of this work?  The general approach, how it's layered, etc.

Or the (long-term) goal of taking all decoding, encoding detection and/or validation, and BOM detection and/or stripping (and maybe EOL normalization) OUT of the parser itself and into a pre-processor, provided doing so either improves performance or has minimal cost?

> It seems to me, rather that implement own decoding stream, we may use [encoding_rs_io](https://docs.rs/encoding_rs_io/latest/encoding_rs_io/) or similar library to do the conversion.

Yes, I agree, that's the path I was going down on the initial implementation.  But assuming we want to continue keeping a separate `encoding` feature, the non `encoding` implementation can still provide the same guarantees with validation, and that then simplifies everything written on top.  You could get rid of the API divergence between `encoding` and `not(encoding)`, safely move to a `str` / `String` based API, etc.

edit: probably this path could also be implemented with `encoding_rs`, it would just require carrying that dependency always.  I am flexible on the details

> I also just created https://github.com/tafia/quick-xml/pull/948 to show you where we may change the reader from the reader that guessing encoding to the reader that transparently decodes. It seems to me that it is at the level of a slightly higher level XmlReader that honest decoding should be introduced.

I still think the better approach is to just push all preprocessing down underneath the parsing code, and take advantage of the simplifications that makes possible.  And the ability to be able to parse UTF-16.

Yes maybe you can skip over performing some validation that way, but there is also a cost to running the validation (and maybe allocations) many times over small buffers instead of once over a large buffer.  Not to mention the additional internal complexity of `Decoder` infecting so many different object types (which also increases the size of all of the structs), even if it was otherwise hidden from the user.

dralley:
Feature-wise, this PR is now complete.  It's just a matter of improving the quality of `Utf8ValidatingReader` and deciding whether the new constructors are how we want it to be exposed. 

Global decoding, improving APIs, etc. -- all of that is for future PRs.  It should be able to be bolted onto this infrastructure though.

src/encoding.rs
```diff
@@ -246,21 +309,1059 @@ pub fn decode_into(
...
+    }
+
+    fn consume(&mut self, amt: usize) {
+        self.reader.consume(amt)
+    }
+}
+
+/// A reader wrapper that ensures only valid UTF-8 bytes are read.
+///
+/// This reader uses [`str::from_utf8()`] and [`Utf8Error::valid_up_to()`] to validate
```
dralley: I'd like to also try [`simdutf8`](https://github.com/rusticstuff/simdutf8)

Probably this could also be done with `encoding_rs`.  If you are open to requiring `encoding_rs` always, then we could ditch this implementation completely.  I just figured that we probably wouldn't want to do that.

dralley:
I'm working on a more efficient implementation that assumes `R: BufRead`.  In the meantime, could you please give your opinion on the overall architecture?  Everything except for the `Utf8ValidatingReader` itself is ready for review.

dralley:
@Mingun I left the original implementation commit intact, but the new implementation in "[Make Utf8ValidatingReader use BufRead](https://github.com/tafia/quick-xml/pull/947/commits/92e888137a207bf225098d335cd24a5b1f543c60)" validates UTF-8 at 30 GB per second (as per the included benchmarks), which is more than 15x faster than the original implementation.  Total memory bandwidth through BufReader is otherwise about 90 GBps.

Since parsing benchmarks rarely breach 500 MBps, I would say that is insignificant overhead.

Though, I'm now working on the decoding variant using encoding_rs, and the implementation is simpler and still performs well.  So maybe just using the one implementation w/ encoding_rs is better and more maintainable, but would be at the expense of always carrying that dependency.

dralley:
@Mingun Please take a look at my comments when you get a moment, I'd like to move this work along.

As mentioned I remain pretty flexible on the implementation (e.g. if you'd prefer to just use `encoding_rs` for everything) and basic API (I don't like the `from_*_validating()` constructors much, it was just convenient to get testing going without committing to API breakage)

Mingun:
I will try to find time to review this and other latest issues / PRs in the end of week.

src/errors.rs
Mingun: If you will rebase you PR for whatever reason, it would be good to add this to the commit that bumps MSRV so we quickly could find reasons if required

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+            }
+            let len = self
+                .chunk_size
+                .min(buf.len())
+                .min(self.data.len() - self.pos);
+            buf[..len].copy_from_slice(&self.data[self.pos..self.pos + len]);
+            self.pos += len;
+            Ok(len)
+        }
+    }
```
Mingun, line 792: That is not necessary, you may use `BufReader::with_capacity`, the effect will be the same.

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+            let mut result = Vec::new();
+
+            loop {
+                let mut buf = [0u8; 10];
+                let n = reader.read(&mut buf).unwrap();
+                if n == 0 {
+                    break;
+                }
+                result.extend_from_slice(&buf[..n]);
+            }
```
Mingun, line 916: You may use more strict check here: we expect, that we will need only two calls of `read` to get the data and the 3rd call should return 0.

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
```
Mingun, line 936: Same here: 3 + 1 calls

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
```
Mingun, line 956: Same here: 4 + 1 calls

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+
+            let mut result = Vec::new();
+            loop {
+                let mut buf = [0u8; 20];
+                let n = reader.read(&mut buf).unwrap();
+                if n == 0 {
+                    break;
+                }
+                result.extend_from_slice(&buf[..n]);
+            }
```
Mingun, line 1010: Same here and below

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+pub struct Utf8BytesReader<R> {
+    #[cfg(feature = "encoding")]
+    reader: io::BufReader<R>,
+    #[cfg(not(feature = "encoding"))]
+    reader: io::BufReader<Utf8ValidatingReader<io::BufReader<R>>>,
+}
+
+// TODO: Utf8BytesReader should manage encoding detection and BOM stripping - that responsibility
+// can then be removed from the readers, with perhaps an exception for slice reader.
+impl<R: io::Read> Utf8BytesReader<R> {
```
Mingun, line 385: Why use `io::` prefix if `Read` was `use`d above?

```suggestion
impl<R: Read> Utf8BytesReader<R> {
```

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+        let data = "Hello, 世界! 😀".as_bytes();
+        let mut reader = Utf8BytesReader::new(&data[..]);
+        let mut result = Vec::new();
+        reader.read_to_end(&mut result).unwrap();
+        assert_eq!(result, data);
+        assert_eq!(std::str::from_utf8(&result).unwrap(), "Hello, 世界! 😀");
+    }
+
+    #[test]
+    fn bufread_interface() {
```
Mingun, line 729: I think, need to add some negative tests for `BufRead` interface, to ensure, that invalid UTF-8 will produce errors

src/reader/mod.rs
```diff
@@ -776,7 +778,7 @@ impl<R> Reader<R> {
     /// Creates a `Reader` that reads from a given reader.
     pub fn from_reader(reader: R) -> Self {
         Self {
-            reader,
+            reader: reader,
```
Mingun, line 781: Unnecessary

```suggestion
            reader,
```

src/reader/mod.rs
Mingun: I think, in the choice new constructors vs opt-in feature, constructors is better. However, it seems that they are unnecessary? Users can construct `Utf8BytesReader` and pass it to existing constructors, right? So probably just documenting that approach on `Reader` documentation will be enough.

Also, because `Utf8BytesReader` itself not guarded by `encoding` feature and it the future, as I understand, you plan to add transparent decoding into it (when `encoding` feature is enabled), then we may expect, that this constructor will return the reader, which will always return UTF-8 encoded events, even when source XML is not in UTF-8 encoding. So I think, `from_reader_validating` not the right name to the thing that it will do (of course, with the current feature guard its name reflects its purpose). Something like `from_reader_transcoding` would be beter? Not the better name, although, that is why I suggest to just not add those constructors at all.

src/encoding.rs
Mingun: Maybe extract this commit into a separate PR? It seems pretty isolated

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+}
+
+/// A struct for transparently decoding / validating bytes as UTF-8.
+#[derive(Debug)]
+pub struct Utf8BytesReader<R> {
+    #[cfg(feature = "encoding")]
+    reader: io::BufReader<R>,
+    #[cfg(not(feature = "encoding"))]
+    reader: io::BufReader<Utf8ValidatingReader<io::BufReader<R>>>,
+}
```
Mingun, line 381: If you need `BufReader`, I suggest to require `R: BufRead` instead of `R: Read` to avoid double bufferisation when user will have `BufReader` and you will create another `BufReader`.
```suggestion
pub struct Utf8BytesReader<R: BufRead> {
    #[cfg(feature = "encoding")]
    reader: R,
    #[cfg(not(feature = "encoding"))]
    reader: io::BufReader<Utf8ValidatingReader<R>>,
}
```

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+            // multi-byte character. We consumed those bytes into `self.pending` so the BufRead
+            // could refill. Now we combine them with fresh data to complete the character.
+            if self.pending_len > 0 {
+                let available = self.inner.fill_buf()?;
+                if available.is_empty() {
+                    self.pending_len = 0;
+                    return Err(io::Error::new(
+                        io::ErrorKind::InvalidData,
+                        EncodingError::Utf8(Utf8ValidationError::IncompleteSequence),
+                    ));
```
Mingun, line 570: Why store `EncodingError` instead of `Utf8ValidationError`? It seems that you always use `EncodingError::Utf8` variant for `io::ErrorKind::InvalidData`

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+
+        // On the very first read, peek at the stream to detect encoding via BOM
+        // or XML declaration patterns. UTF-16 is rejected; UTF-8 BOM is stripped.
+        if !self.encoding_checked {
+            self.encoding_checked = true;
+
+            let available = self.inner.fill_buf()?;
+            // detect_encoding uses starts_with, so patterns longer than the
+            // available data simply won't match — no length guard needed.
+            if let Some(detected) = detect_encoding(available) {
```
Mingun, line 535: Isn't `encoding_checked` set too early? If the first returned buffer will not give us enough bytes to detect encoding we will not detect it at all. That is possible even for `BufReader` with large buffer, because `fill_buf` may continuously return small buffer until it will not be consumed. That was the reason of #939

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+
+    fn consume(&mut self, amt: usize) {
+        self.reader.consume(amt)
+    }
+}
+
+/// Returns the expected total number of bytes in a UTF-8 character given its first byte
+/// (2, 3, or 4). Used to determine how many continuation bytes are needed to complete a
+/// pending incomplete sequence.
+fn utf8_char_width(first_byte: u8) -> usize {
```
Mingun, line 427: May be `const`?
```suggestion
const fn utf8_char_width(first_byte: u8) -> usize {
```

Also, it seems, that using compare will faster (spied at [utf8_width](https://docs.rs/utf8-width/latest/utf8_width/)):
https://godbolt.org/z/oxsjY8nja

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+}
+
+/// Finds the largest byte index <= `index` that falls on a UTF-8 character boundary.
+/// Used when the caller's output buffer is smaller than the available valid data —
+/// we must avoid consuming a partial multi-byte character from the BufRead, because
+/// orphaned continuation bytes at the start of the next `fill_buf()` would be
+/// misreported as invalid UTF-8.
+///
+/// The caller must ensure that `bytes[..index]` contains valid UTF-8 data.
+fn floor_char_boundary(bytes: &[u8], index: usize) -> usize {
```
Mingun, line 448: If it possible to make it `const`, add `const`.

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+                    // partial character from the BufRead — otherwise the next fill_buf() would
+                    // start with orphaned continuation bytes, causing a false validation error.
+                    let len = floor_char_boundary(available, buf.len());
+                    if len == 0 {
+                        return Ok(0);
+                    }
+                    buf[..len].copy_from_slice(&available[..len]);
+                    self.inner.consume(len);
+                    return Ok(len);
+                }
```
Mingun, line 642: You validate some input multiple times here. Imagine, that `available` is 8Kb, but `buf` only 4 bytes. You will validate 8192 bytes, but consume only 1..=4. Then validate 8188..=8191 bytes (which already was checked) and consume only 1..=4 and so on.

Limit `available` by `buf.len()` and validate only that part. That may be slower because you will validate by chunks that depends by size of output buffer. Maybe better approach would be
- maintain the length of already validated data
- return them without checking when requested not more than that length
- when new data coming, validate only them

src/encoding.rs
```diff
@@ -246,21 +309,1080 @@ pub fn decode_into(
...
+/// let n = reader.read(&mut buf).unwrap();
+/// assert_eq!(&buf[..n], data);
+/// ```
+#[derive(Debug)]
+pub struct Utf8ValidatingReader<R> {
+    inner: R,
+    /// Small buffer for incomplete UTF-8 sequences at BufRead boundaries.
+    /// At most 3 bytes (the start of a 2, 3, or 4-byte sequence).
+    pending: [u8; 3],
+    pending_len: u8,
```
Mingun, line 489: It seems that if you pull this code into a separate auxiliary struct, the algorithm will become clearer. The same approach is used in `encoding_rs_rw`

Mingun (changes requested):
> Architecturally, do you have any objections to the direction of this work? The general approach, how it's layered, etc.

Architecturally, I suggest that the parser inside still work with `[u8]`, since it does not require UTF-8 guarantees to work. We only need guarantees of XML-compatible encoding[^1], and these are all single-byte encodings that `encoding_rs` supports. On the other hand, users usually work with UTF-8 text and they would like to have an API that operates on `str` / `String` rather than `[u8]` / `Vec<u8>`. I think this is quite achievable, but we will have to introduce duplicate structures for event types or somehow parametrize existing ones.

We can achieve correctness now[^2] by substituting `R` in `Reader<R>`, which will transcode the readable data to UTF-8 on the fly, or, as in the case of this PR, check that the data is correctly encoded in UTF-8 (which is also a kind of zero-cost transcoding).

In this sense, I agree with the `Utf8ValidatingReader` approach implemented here. To further develop and provide the `str`-API, I assume that one should move towards implementing the new methods in
```rust
impl<R> Reader<Utf8ValidatingReader<R>> {
}
```
which will call the existing `[u8]`-based methods and convert the result to `str` (using `unsafe` methods, since we will check that the input was in UTF-8).

> Or the (long-term) goal of taking all decoding, encoding detection and/or validation, and BOM detection and/or stripping (and maybe EOL normalization) OUT of the parser itself and into a pre-processor, provided doing so either improves performance or has minimal cost?

This is probably closer to my vision, but more research is needed here. For example, we want to report offsets in units of the original byte source, and not in units of the source after normalization/decoding, which will be hidden from the original user.

> I still think the better approach is to just push all preprocessing down underneath the parsing code, and take advantage of the simplifications that makes possible. And the ability to be able to parse UTF-16.

Yes, #948 just allows you to do it. If the first event is an XML declaration (and the XML standard says that the XML declaration should be the very first event, even empty lines in front are not allowed. This, by the way, allows you to detect UTF-16-like encoding without using BOM), then we can extract the encoding from it and replace the reader in `EntityReader` with another one that will implement decoding by large chunks before parsing (since `Box<dyn BufRead>` is stored there, the type will not change).

> edit: probably this path could also be implemented with `encoding_rs`, it would just require carrying that dependency always. I am flexible on the details

In theory, we could use `encoding_rs_io` or `encoding_rs_rw` if we make PR there, which will make their dependence on `encoding_rs` optional, and the only supported operation in the absence of `encoding_rs` will check the input to UTF-8 using standard library methods.

----

I choose **Request changes** variant, because I feel that at least that can be changed:
- use `BufReader::with_capacity` instead of `ChunkedReader`
- use `BufRead` in `Utf8BytesReader`
- `detect_encoding` could be extracted to separate PR
- I think, that new constructors are redundant now. We always can add them later if required

[^1]: That is, those in which XML markup characters have the same codes as in ASCII / UTF-8
[^2]: Of course, currently we should know the encoding in advance. If we want to change it after reading the XML declaration or BOM, then we will need to change the `Reader` from which we read. I believe it should be easier to do after the introduction of a new XML reader in #948

dralley:
Alright, looking at this again now (spent some time on another project)

src/encoding.rs
dralley: Done

src/encoding.rs
dralley: I don't think so, BufReader has various tricks to bypass the buffer entirely if the buffer is too small

Thus reads are not necessarily capped at "capacity" bytes

https://doc.rust-lang.org/src/std/io/buffered/bufreader.rs.html#339-342

src/encoding.rs
dralley: `Utf8ValidatingReader` internally buffers until it can return a valid sequence.  So one single call to `{Utf8ValidatingReader}.read()` ought to successfully return the 2 bytes, but that maps to two `read()` calls on the inner `ChunkedReader`.  And then I suppose a 2nd `read()` of the `Utf8ValidatingReader` would map to a single `read()` call on the internal `ChunkedReader`.  So 2 external calls = 3 internal calls.

There's no way to read the intermediate results from `ChunkedReader` in the test.  But we could store a `call_count` to assert on the number of internal `read()` calls, potentially.

src/errors.rs
dralley: Sure, but on the other hand it's just 1.79, probably 90% of the (actively maintained) ecosystem has a newer MSRV than that already.

IMO the benefit is marginal when the pull request is one click away from the commit, and the commit itself is only one commit away from the commit that is "the reason"

I could just note that in the commit message.

src/reader/mod.rs
dralley: It's not strictly necessary, no.  They do make it easier to use, which maybe matters if we want users to try it out before we commit to the approach, but TBH think about this again now I think I'm happy to just not have the constructors and make the tests a little more verbose for now.

dralley:
@Mingun If you think it would be worthwhile, I'm open to splitting this PR up this PR slightly and only providing `DecodingReader` for now.  It's much simpler than the `Utf8ValidatingReader` implementation and approximately as fast, and would at least ship something useful for the next release.  Obviously it's still an opt-in feature.

src/encoding.rs
dralley: I had tried to be clear that the specific details of the implementation were not ready for review, just the overall concept.  I wanted agreement on general direction before I put in the additional hours needed to polish - which it definitely needs.

But, yes that's true.

src/encoding.rs
Mingun: I just mean to unroll loop manually. That should be possible, because we expect that on some iteration `n` would be 0 and the number of iterations should be constant.

Mingun:
Yes, if you think, that PR may be split into isolated parts, I think, it would be worthwhile. You still may keep this PR, just rebase it over the other smaller PRs as required.

src/encoding.rs
dralley: I don't think it ends up being simpler tbh.  Minimal difference, more code.

dralley:
I'm closing this, I think the immediate path forwards looks slightly different.  What I want is:

If the "decoding" feature is enabled:

* `DecodingReader` is built in-to `Reader` and validates or transcodes input automatically
* Therefore, the parser can use `std::str::from_utf8_unchecked()` at XML element boundaries, eliminating all decoding overhead from small strings - the decoding loop therefore remains a tight loop with good data/cache locality and maximum SIMD benefits from operating over a large buffer
* The user need not do anything manually to handle alternate document encodings, they are decoded transparently

If the "decoding" feature is not enabled:

* `Reader` uses `std::str::from_utf8()` (or similar) at XML element boundaries, doing the validation lazily and allowing the consumption of non-UTF-8 data embedded within the XML using `BinaryStream` / `stream()`, which would bypass the validation.
* The user still does not need to do anything manually to decode the document, but the portions of the input processed as XML must be UTF-8

encoding_rs can handle the streaming UTF-8 validation case, and the non-decoding path needs to be capable of handling non-UTF-8 data for users of `BinaryStream`.  The value proposition of a 3rd path (streaming validation without encoding_rs) is slim.

PR #951, opened 2026-04-24
Deserialize custom entities inside attributes

mmirate:
As shown by the new tests, the result is that deserialization will process custom entities in attributes as well as in text content.

This seems to require some breaking changes to the public API.

### DIFF
--- a/src/de/attributes.rs
+++ b/src/de/attributes.rs
@@ -6,6 +6,7 @@ use serde::de::{DeserializeSeed, Deserializer, Error, MapAccess, Visitor};
 use serde::forward_to_deserialize_any;
 
 use crate::de::key::QNameDeserializer;
+use crate::de::{EntityResolver, PredefinedEntityResolver};
 use crate::de::SimpleTypeDeserializer;
 use crate::errors::serialize::DeError;
 use crate::events::attributes::Attributes;
@@ -25,6 +26,7 @@ impl<'i> Attributes<'i> {
     /// ```
     /// # use pretty_assertions::assert_eq;
     /// use quick_xml::events::BytesStart;
+    /// use quick_xml::de::PredefinedEntityResolver;
     /// use quick_xml::XmlVersion;
     /// use serde::Deserialize;
     ///
@@ -47,7 +49,7 @@ impl<'i> Attributes<'i> {
     ///     3
     /// );
     /// // Strip nothing from the field names
-    /// let de = tag.attributes().clone().into_map_access(XmlVersion::V1_0, "");
+    /// let de = tag.attributes().clone().into_map_access(XmlVersion::V1_0, "", &PredefinedEntityResolver);
     /// assert_eq!(
     ///     MyData::deserialize(de).unwrap(),
     ///     MyData {
@@ -57,7 +59,7 @@ impl<'i> Attributes<'i> {
     /// );
     ///
     /// // Strip "@" from the field name
-    /// let de = tag.attributes().into_map_access(XmlVersion::V1_0, "@");
+    /// let de = tag.attributes().into_map_access(XmlVersion::V1_0, "@", &PredefinedEntityResolver);
     /// assert_eq!(
     ///     MyDataPrefixed::deserialize(de).unwrap(),
     ///     MyDataPrefixed {
@@ -67,17 +69,19 @@ impl<'i> Attributes<'i> {
     /// );
     /// ```
     #[inline]
-    pub const fn into_map_access(
+    pub const fn into_map_access<E: EntityResolver>(
         self,
         version: XmlVersion,
         prefix: &'static str,
-    ) -> AttributesDeserializer<'i> {
+        entity_resolver: &'i E,
+    ) -> AttributesDeserializer<'i, E> {
         AttributesDeserializer {
             iter: self,
             value: None,
             prefix,
             key_buf: String::new(),
             version,
+            entity_resolver,
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/de/simple_type.rs
```diff
@@ -505,26 +505,38 @@ pub struct SimpleTypeDeserializer<'de, 'a> {
...
     /// Creates a deserializer from an XML text node, that possible borrowed from input.
     ///
     /// It is assumed that `text` does not have entities.
     ///
     /// This constructor used internally to deserialize from text nodes.
-    pub fn from_text_content(value: Text<'de>) -> Self {
-        Self::from_text(value.text)
+    pub fn from_text_content(value: Text<'de>, entity_resolver: &'a E) -> Self {
+        Self::from_text(value.text, entity_resolver)
     }
```
Mingun: `SimpleTypeDeserializer` created from text content cannot contain entities, as noted in the docs, so it is redundant to specify entity resolver here. `Text` contains already merged text and expanded entities:
https://github.com/tafia/quick-xml/blob/6238d8a16179d68718b43882e544ebd20982c79c/src/de/mod.rs#L2382-L2434

Please add the following test:
```rust
#[test]
fn need_to_give_reasonable_name() {
    let value: String = from_str("<root>&amp;lt;</root>").unwrap();
    assert_eq!(value, "&lt;");
}
```

src/de/simple_type.rs
```diff
@@ -505,26 +505,38 @@ pub struct SimpleTypeDeserializer<'de, 'a> {
     /// Not used for deserializing raw byte buffers
     decoder: Decoder,
     version: XmlVersion,
+
+    /// Used to resolve unknown entities that would otherwise cause the parser
+    /// to return an [`EscapeError::UnrecognizedEntity`] error.
+    ///
+    /// [`EscapeError::UnrecognizedEntity`]: crate::escape::EscapeError::UnrecognizedEntity
+    entity_resolver: &'a E,
```
Mingun: `entity_resolver` should be stored by value.

tests/serde-de.rs
```diff
@@ -1894,6 +1922,63 @@ mod resolve {
...
     }
+
+    #[test]
+    fn resolve_custom_entity_in_attr() {
+        let resolver = TestEntityResolver {
+            capture_called: false,
+        };
+        let mut de = Deserializer::with_resolver(
+            br#"
+            <!DOCTYPE dict[ <!ENTITY unc "unclassified"> ]>
```
Mingun: To keep as least one [validity constraint](https://www.w3.org/TR/xml11/#vc-roottype) of XML, use the name as the root tag:

```suggestion
            <!DOCTYPE root[ <!ENTITY unc "unclassified"> ]>
```

tests/serde-de.rs
```diff
@@ -1894,6 +1922,63 @@ mod resolve {
...
+    fn resolve_custom_entity_in_attr_but_reject_infinite_recursion() {
+        use quick_xml::errors::Error;
+        use quick_xml::escape::EscapeError;
+
+        let resolver = TestHostileEntityResolver {
+            capture_called: false,
+        };
+        let mut de = Deserializer::with_resolver(
+            br#"
+            <!DOCTYPE dict[ <!ENTITY unc "unclassified"> ]>
```
Mingun: And here
```suggestion
            <!DOCTYPE root[ <!ENTITY unc "unclassified"> ]>
```

src/de/attributes.rs
```diff
@@ -67,17 +69,19 @@ impl<'i> Attributes<'i> {
...
+    ) -> AttributesDeserializer<'i, E> {
         AttributesDeserializer {
             iter: self,
             value: None,
             prefix,
             key_buf: String::new(),
             version,
+            entity_resolver,
         }
     }
```
Mingun: Probably
```rust
impl AttributesDeserializer {
  fn with_resolver<E: EntityResolver>(self, resolver) -> AttributesDeserializer<E> {
    // replace entity_resolver with new one
  }
}
```
would be more practical, because I think, in most cases `PredefinedEntityResolver` will be used. So if you need use specific resolver, you may write a chain:
```rust
attributes.into_map_access("@").with_resolver(my_resolver)
```

Mingun (changes requested):
It seems to me that current implementation will apply entity resolving twice for the text case:
```rust
#[test]
fn need_to_give_reasonable_name() {
    let value: String = from_str("<root>&amp;lt;</root>").unwrap();
    assert_eq!(value, "&lt;");
}
```
That test should be added.

Probably it would be better to create separate deserializer for attribute values rather that reuse `SimpleTypeDeserializer`. It is hard to say without trying.

src/de/simple_type.rs
mmirate: This would require EntityResolver to have Clone as a supertrait - is this desirable?

src/de/simple_type.rs
mmirate: The abovementioned test will pass.

This test will also pass:

```rust
#[test]
fn need_to_give_reasonable_name() {
    let value: BTreeMap<String, String> = from_str(r#"<root attr="&amp;lt;" />"#).unwrap();
    assert_eq!(value, BTreeMap::from_iter([
        (String::from("@attr"), String::from("&lt;"))
    ]));
}
```

The tests where instead of `&amp;` there is a custom entity resolver that supports a custom analogue of `&amp;` are more longwinded. These tests pass for `&amp;lt;` in element text but produce an UnterminatedEntity error for `&amp;lt;` as attribute content. I'm not sure whether that is a bug or an inevitable consequence of the recursive nature of "attribute normalization".

PR #952, opened 2026-04-27
Remove feature gating from detect_encoding()

dralley:
Have it return a generic enum so that it can be used with or without the encoding feature.

It was suggested that this be split off from #947

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -4,18 +4,16 @@ use std::borrow::Cow;
 use std::str::Utf8Error;
 
 #[cfg(feature = "encoding")]
-use encoding_rs::{DecoderResult, Encoding, UTF_16BE, UTF_16LE, UTF_8};
+use encoding_rs;
 
 /// Unicode "byte order mark" (\u{FEFF}) encoded as UTF-8.
 /// See <https://unicode.org/faq/utf_bom.html#bom1>
 pub(crate) const UTF8_BOM: &[u8] = &[0xEF, 0xBB, 0xBF];
 /// Unicode "byte order mark" (\u{FEFF}) encoded as UTF-16 with little-endian byte order.
 /// See <https://unicode.org/faq/utf_bom.html#bom1>
-#[cfg(feature = "encoding")]
 pub(crate) const UTF16_LE_BOM: &[u8] = &[0xFF, 0xFE];
 /// Unicode "byte order mark" (\u{FEFF}) encoded as UTF-16 with big-endian byte order.
 /// See <https://unicode.org/faq/utf_bom.html#bom1>
-#[cfg(feature = "encoding")]
 pub(crate) const UTF16_BE_BOM: &[u8] = &[0xFE, 0xFF];
 
 /// An error when decoding or encoding
@@ -30,7 +28,7 @@ pub enum EncodingError {
     Utf8(Utf8Error),
     /// Input did not adhere to the given encoding
     #[cfg(feature = "encoding")]
-    Other(&'static Encoding),
+    Other(&'static encoding_rs::Encoding),
 }
 
 impl From<Utf8Error> for EncodingError {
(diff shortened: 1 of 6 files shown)


src/reader/slice_reader.rs
```diff
@@ -4,10 +4,12 @@
 
 use std::io;
 
+#[cfg(feature = "encoding")]
+use crate::encoding::DetectedEncoding;
 #[cfg(feature = "encoding")]
 use crate::reader::EncodingRef;
 #[cfg(feature = "encoding")]
-use encoding_rs::{Encoding, UTF_8};
+use encoding_rs;
```
dralley, line 12: I wanted it to be 100% clear where types were coming from, as some of the constants in `mod encoding` are very similar to other types or constants from `encoding_rs`, and especially with arguments and return values it's best that it be clear they are foreign types.

src/reader/async_tokio.rs
```diff
@@ -7,6 +7,7 @@ use std::task::{Context, Poll};
 
 use tokio::io::{self, AsyncBufRead, AsyncBufReadExt, AsyncRead, ReadBuf};
 
+use crate::encoding;
```
dralley, line 10: Used by the macros.

src/reader/mod.rs
```diff
@@ -267,7 +269,7 @@ macro_rules! read_event_impl {
                     #[cfg(feature = "encoding")]
                     if let Some(encoding) = $reader.detect_encoding() $(.$await)? ? {
                         if $self.state.encoding.can_be_refined() {
-                            $self.state.encoding = crate::reader::EncodingRef::BomDetected(encoding);
+                            $self.state.encoding = crate::reader::EncodingRef::BomDetected(encoding.encoding());
```
dralley, line 272: As mentioned on the previous PR, `detect_encoding()` returns / returned `Some()` even in cases where no BOM was found, thus this is perhaps misnamed.

However, the initial XML characters function *like* a BOM, and the documentation mentions it, so I see an argument for leaving this be.

dralley:
After this PR, can we go ahead and cut a release that includes the attribute value normalization?

Mingun:
> After this PR, can we go ahead and cut a release that includes the attribute value normalization?

Agree, I also want to release new version soon. I will review this and try to fix the latest issues and then release new version. If the latest issues will hard to fix quickly, then I'll release new version without their fixes.

src/encoding.rs
```diff
@@ -246,21 +245,63 @@ pub fn decode_into(
...
+    pub fn encoding(&self) -> &'static encoding_rs::Encoding {
+        match self {
+            DetectedEncoding::AsciiCompatible | DetectedEncoding::Utf8Bom => encoding_rs::UTF_8,
+            DetectedEncoding::Utf16Le | DetectedEncoding::Utf16LeBom => encoding_rs::UTF_16LE,
+            DetectedEncoding::Utf16Be | DetectedEncoding::Utf16BeBom => encoding_rs::UTF_16BE,
+        }
+    }
+
+    /// Length of the BOM, which may need to be stripped from the input
+    pub fn bom_len(&self) -> usize {
```
Mingun: It seems, that both may be const

src/encoding.rs
```diff
@@ -246,21 +245,63 @@ pub fn decode_into(
...
+/// Possible scenarios for start-of-xml detection of encoding
+///
+/// See the documentation of [`detect_encoding`]
+pub enum DetectedEncoding {
+    /// Matches UTF-8 or some other ascii-compatible encoding
+    AsciiCompatible,
+    /// We saw a UTF-8 BOM
+    Utf8Bom,
+    /// Matches UTF-16-LE or some other UTF-16 compatible encoding (e.g. ISO-10646-UCS-2)
+    Utf16Le,
```
Mingun: Maybe 
```suggestion
    Utf16LeLike,
```

src/encoding.rs
```diff
@@ -246,21 +245,63 @@ pub fn decode_into(
...
+    /// We saw a UTF-16 BOM in little-endian orientation
+    Utf16LeBom,
+    /// Matches UTF-16-BE or some other UTF-16 compatible encoding (e.g. ISO-10646-UCS-2)
+    Utf16Be,
```
Mingun: Maybe
```suggestion
    Utf16BeLike,
```

src/reader/slice_reader.rs
Mingun: Agree

Mingun (approved):
It seems two small improvements could be made, but other is good

dralley:
re-pushed

PR #955, opened 2026-05-03
Improve XmlVersion enumeration: explicit and implicit 1.0 variants

Mingun:
This PR changes XmlVersion from
```rust
pub enum XmlVersion {
  V1_0,
  V1_1,
}
```
to
```rust
pub enum XmlVersion {
  Implicit1_0,
  Explicit1_0,
  Explicit1_1,
}
```
The motivation behind this change is simplification with processing versions. XML specification explicitly allows at most one XML declaration (the `<?xml version=...?>` part). At the same time the `version` information in the declaration is mandatory. This leads us to a simple cycle that allows, using one variable, to track the uniqueness of the declaration and obtain values   according to the version
```rust
let mut reader = Reader::from_str(...);
let mut version = XmlVersion::Implicit1_0;
loop {
  match reader.read_event() {
    // Return error in case of duplicated declaration
    Event::XmlDecl(e) if version != XmlVersion::Implicit1_0 => return Err(...),
    // Return error in case of missing "version"
    Event::XmlDecl(e) => { version = e.xml_version()?; }

    Event::Text(e) => {
      let text = e.xml_content(version)?;
      // ...
    }
    ...
  }
}
```
Without thus changes you may use `Option<XmlVersion>`, but then you will forced to use `e.xml_content(version.unwrap_or(XmlVersion::V1_0))`, which will add check for `None` for every usage of `version` in the loop. Usually XML documents a well-formed, so this is just wasting of CPU cycles.

`XmlVersion` is not a part of any release yet, so it can be changed freely.

### DIFF
--- a/benches/macrobenches.rs
+++ b/benches/macrobenches.rs
@@ -50,7 +50,7 @@ static INPUTS: &[(&str, &str)] = &[
 
 fn parse_document_from_str(doc: &str) -> XmlResult<()> {
     let mut r = Reader::from_str(doc);
-    let mut version = XmlVersion::V1_0;
+    let mut version = XmlVersion::Implicit1_0;
     loop {
         match black_box(r.read_event()?) {
             Event::Decl(e) => {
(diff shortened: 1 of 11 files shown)


dralley (commented):
Needs a changelog entry, yes?  `XmlVersion` is public

edit: never mind, didn't see the note about it not being released yet.

dralley (approved):

PR #956, opened 2026-05-04
Add DecodingReader in the encoding module

dralley:


dralley:
Adding more tests

Mingun:
I think, we'll merge this after 0.40 (because it bumps MSRV)? From my side, I do not plan to include anything else to the release -- other potential issues requires a lot of work, and we do not want to postpone the release, right? If so, then I'll start releasing.

dralley:
If you really need to exclude it, that's fine, however this PR is very non-invasive.  The only impact on core functionality is a bit of error handling scaffolding and `normalize*` functions being re-enabled under `#[cfg(feature = "encoding")]`.

Yes, it bumps MSRV, but to a version released 2 years ago.  IMO it would be nice to include.

But, if it must, it can wait.

src/events/attributes.rs
dralley: Should this note be removed or updated in favor of the `DecodingReader` approach?  I considered it, but thought I should just leave it be for now.

src/encoding.rs
```diff
@@ -410,6 +423,32 @@ impl<R> DecodingReader<R> {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the prefix buffer has already been drained (i.e. the main
+    /// decode path has consumed raw bytes from the inner reader). At that
+    /// point, switching encodings would silently produce wrong output.
+    ///
+    /// This is safe to call any time before the prefix is exhausted - in
+    /// practice, immediately after parsing the XML declaration.
+    pub fn set_encoding(&mut self, encoding: &'static encoding_rs::Encoding) {
```
dralley, line 470: Another way to do this would be to make the `prefix` buffer "peekable", let the caller access the raw bytes of it and call `detect_encoding` on it before actually advancing the state of the reader.

That would probably simplify `DecodingReader` a fair bit.

The downside there, and why I didn't take that approach (though I am open to it) is that `XmlSource::detect_encoding()` would need to re-implement some subset of the handling of the "Decl" event.  There are maybe ways you could perhaps simplify it, like just scanning for the bytes "encoding", but avoiding touching the parser entirely seemed like the better call for now.

Also, it would require some coupling, since currently `Reader` only speaks `BufRead`.  We will want the coupling eventually, but not right now.

dralley: Still, it might be the better idea long-term, as it avoids any potential issues with resetting the decoder at improper boundaries.

dralley:
@Mingun This is ready for review

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+                detected: false,
+            })),
+            inner_eof: false,
+        }
+    }
+
+    /// Returns a reference to the underlying reader
+    pub fn get_ref(&self) -> &R {
+        &self.inner
+    }
```
Mingun, line 442: It seems that you can add `const` to some methods

Mingun (commented):
Just want to note, that I'm working on review. Right now only one minor note.

If new rebase would be needed, then fix some commit comments, that mentions `Utf8BytesReader` (which is now `DecodingReader` as I think)

examples/read_utf16.rs
```diff
@@ -0,0 +1,46 @@
...
+    loop {
+        match reader.read_event_into(&mut buf) {
+            Ok(Event::Start(ref e)) => {
+                let name = e.name();
+                let name = std::str::from_utf8(name.as_ref()).unwrap();
+                println!("start: <{}>", name);
+
+                for attr in e.attributes().flatten() {
+                    let key = std::str::from_utf8(attr.key.as_ref()).unwrap();
+                    let val = attr.normalized_value(XmlVersion::Implicit1_0)?;
```
Mingun: I think, we may demonstrate here the correct way to deal with XML version, as shown in #955. Yes, I know, this example is not focused on correct version handling, but there is no examples of that yet, so it will be better to show it somewhere

src/errors.rs
```diff
@@ -221,7 +221,12 @@ impl From<IoError> for Error {
     /// Creates a new `Error::Io` from the given error
     #[inline]
     fn from(error: IoError) -> Error {
-        Self::Io(Arc::new(error))
+        match error.kind() {
+            IoErrorKind::InvalidData => Self::Encoding(error.downcast::<EncodingError>().expect(
+                "Got an IoError::InvalidData, but it wasn't downcastable to EncodingError?",
+            )),
+            _ => Self::Io(Arc::new(error)),
+        }
```
Mingun, line 230: The underlying reader wrapped in `DecodingReader` also may use `InvalidData` for its own purposes, so we need to to pass that error unchanged

```suggestion
        match error.kind() {
            IoErrorKind::InvalidData => match error.downcast::<EncodingError>() {
                Ok(err) => Self::Encoding(err),
                Err(err) =>  Self::Io(Arc::new(err)),
            }
            _ => Self::Io(Arc::new(error)),
        }
```

tests/documents/encoding/utf8.xml
```diff
@@ -1,3 +1,120 @@
...
+        <author>Corets, Eva</author>
+        <title>Oberon's Legacy</title>
+        <genre>Fantasy</genre>
+        <price>5.95</price>
+        <publish_date>2001-03-10</publish_date>
+        <description>In post-apocalypse England, the mysterious
+      agent known only as Oberon helps to create a new life
+      for the inhabitants of London. Sequel to Maeve
+      Ascendant.</description>
+    </book>
```
Mingun: Actually, this file in ASCII. If we really want to test encoding, it would be better to include actual examples of 2-, 3- and 4-byte characters.

For example, 2-byte characters (Russian):
```suggestion
    <book id="bk104">
        <author>Corets, Eva</author>
        <title>Наследие Оберона</title>
        <genre>Фантастика</genre>
        <price>5.95</price>
        <publish_date>2001-03-10</publish_date>
        <description>В постапокалипсической Англии
      загадочный агент, известный как Оберон, помогает жителям
      Лондона обустроить новую жизнь. Продолжение Maeve
      Ascendant.</description>
    </book>
```

benches/encoding.rs
```diff
@@ -0,0 +1,98 @@
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
```
Mingun, line 1: Use `pretty_assertions`
```suggestion
use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
use pretty_assertions::assert_eq;
```

benches/encoding.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    for buf_size in [64, 1024, 8192] {
+        group.bench_with_input(
+            BenchmarkId::new("DecodingReader_utf16le", buf_size),
+            &buf_size,
+            |b, &buf_size| {
+                b.iter(|| {
+                    let mut reader = DecodingReader::new(&utf16_sample[..]);
+                    let mut buf = vec![0u8; buf_size];
+                    drain_reader(&mut reader, &mut buf);
```
Mingun: Why you do not check the size? Because UTF-16 decoded into UTF-8 we should expect the same size, as for UTF-8 tests (and I checked that)
```suggestion
                    let n = drain_reader(&mut reader, &mut buf);
                    assert_eq!(n, SAMPLE.len());
```

tests/encodings.rs
```diff
@@ -1,11 +1,141 @@
...
 use encoding_rs::{UTF_16BE, UTF_16LE, UTF_8, WINDOWS_1251};
-use pretty_assertions::assert_eq;
-use quick_xml::events::{BytesEnd, BytesStart, BytesText, Event::*};
+#[cfg(feature = "encoding")]
+use quick_xml::events::Event::*;
 use quick_xml::reader::Reader;
 
-mod decode {
+#[cfg(feature = "encoding")]
+mod xml_decoding_reader {
```
Mingun, line 12: It would be better to change this file in several commits, because current diff hard to read (and as consequence hard to understand what's changed). The first change could be everything except adding new tests, and the second adds `xml_decoding_reader` tests.

tests/encodings.rs
```diff
@@ -1,11 +1,141 @@
...
+            0x3C, 0x00, // '<'
+            0x61, 0x00, // 'a'
+            0x3E, 0x00, // '>'
+            0x00, 0xDC, // unpaired low surrogate
+        ]);
+        assert!(
+            matches!(result, Error::Encoding(_)),
+            "Expected EncodingError::Other, got: {:?}",
+            result,
+        );
```
Mingun, line 57: You say, that expecting concrete encoding error, but you only check that some encoding error was produced

tests/encodings.rs
```diff
@@ -1,11 +1,141 @@
...
+            0x00, 0x3C, // '<'
+            0x00, 0x61, // 'a'
+            0x00, 0x3E, // '>'
+            0xDC, 0x00, // unpaired low surrogate
+        ]);
+        assert!(
+            matches!(result, Error::Encoding(_)),
+            "Expected EncodingError::Other, got: {:?}",
+            result,
+        );
```
Mingun, line 74: Same here

tests/encodings.rs
```diff
@@ -1,11 +1,141 @@
...
+    }
+
+    // UTF-8 / 16 "happy paths" already well-tested in encoding.rs
+
+    #[test]
+    fn test_koi8_r_encoding() {
+        let src = include_bytes!("documents/opennews_all.rss").as_ref();
```
Mingun: Extract into const? It is used 2 times

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+        let mut out = vec![0xFE, 0xFF]; // UTF-16 BE BOM
+        for code_unit in s.encode_utf16() {
+            out.extend_from_slice(&code_unit.to_be_bytes());
+        }
+        out
+    }
+
+    /// Encode a string as UTF-16 LE bytes without BOM
+    fn utf16le_no_bom(s: &str) -> Vec<u8> {
+        let mut out = Vec::new();
```
Mingun, line 736: Nit: you may preallocate
```suggestion
        let mut out = Vec::with_capacity(s.len() * 2);
```

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+    /// Read all bytes from a reader into a String
+    fn read_all(reader: &mut DecodingReader<impl BufRead>) -> io::Result<String> {
+        let mut result = Vec::new();
+        reader.read_to_end(&mut result)?;
+        Ok(String::from_utf8(result).expect("DecodingReader should produce valid UTF-8"))
+    }
+
+    /// Simple edge cases and degenerate inputs
+    mod edge_cases {
+        use super::*;
```
Mingun, line 761: Use `pretty_assertions` for nicer error messages
```suggestion
        use super::*;
        use pretty_assertions::assert_eq;
```

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+                result.extend_from_slice(&buf[..n]);
+            }
+            assert_eq!(String::from_utf8(result).unwrap(), "Hello, 世界!");
+        }
+    }
+
+    /// Tests that exercise the BufRead contract (fill_buf + consume) directly,
+    /// as opposed to the Read-based helpers used elsewhere.
+    mod bufread_interface {
+        use super::*;
```
Mingun, line 839:
```suggestion
        use super::*;
        use pretty_assertions::assert_eq;
```

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+
+            let buf = reader.fill_buf().unwrap();
+            assert!(!buf.is_empty());
+            assert_eq!(buf[0], b',');
+        }
+
+        /// fill_buf/consume loop over UTF-16 LE input; verifies that
+        /// transcoded output reassembles to the original string.
+        #[test]
+        fn utf16_fill_buf() {
```
Mingun: It seems that this test superseded by the next test just be test nature. I would remove it.

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+        fn consume_overflow_panics_in_debug() {
+            let data = b"Hi";
+            let mut reader = DecodingReader::new(&data[..]);
+            let _ = reader.fill_buf().unwrap();
+            reader.consume(100);
+        }
+    }
+
+    mod accessors {
+        use super::*;
```
Mingun, line 920:
```suggestion
        use super::*;
        use pretty_assertions::assert_eq;
```

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+
+    // TODO: These tests emulate the updating of the internal decoder after reading the XML decl.
+    // Since `Reader` currently only speaks the `BufRead` trait, we can't test that directly.
+    // Eventually once `Reader` knows about the underlying `DecodingReader` we should test
+    // that directly.
+
+    /// Tests for encoding() and set_encoding(): detection, switching,
+    /// same-encoding no-op safety, and mid-stream override behavior.
+    mod encoding_switching {
+        use super::*;
```
Mingun, line 966:
```suggestion
        use super::*;
        use pretty_assertions::assert_eq;
```

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+    /// - UTF-8 passthrough (ASCII and multibyte) with and without BOM
+    /// - UTF-16 LE/BE decoding with and without BOM
+    /// - BOM-less UTF-16 detection via `<?xml` byte pattern
+    /// - UTF-16 surrogate pairs (astral plane characters)
+    /// - Chunked input at misaligned boundaries (odd chunk sizes vs 2-byte code units)
+    /// - One-byte-at-a-time delivery for all encodings
+    /// - Inputs larger than the 8192-byte internal output buffer
+    /// - Empty and single-character inputs (prefix-only decode path)
+    mod matrix_decoding_tests {
+        use super::*;
```
Mingun, line 1071:
```suggestion
        use super::*;
        use pretty_assertions::assert_eq;
```

src/encoding.rs
```diff
@@ -305,3 +307,927 @@ impl DetectedEncoding {
...
+// bytes decoded with a potentially wrong initial encoding.
+#[cfg(feature = "encoding")]
+const PREFIX_CAP: usize = 64;
+
+#[cfg(feature = "encoding")]
+struct Prefix {
+    buf: [u8; PREFIX_CAP],
+    len: usize,
+    detected: bool,
+}
```
Mingun, line 322: I would rewrite PR so that this structure is from the beginning. `PREFIX_CAP` in the introducing commint could be 4, and struct in embedded directly, and in the second pack it into `Option<Box>`. That will reduce unnecessary noise in diffs

src/events/attributes.rs
Mingun: The note about lib usage still essential, if lib does not control the `R` type. It users may not use `DecodingReader`. But the warning should be reworded, of course.

Mingun (changes requested):

tests/documents/encoding/utf8.xml
dralley: I added a new document in the latest commit.  If you're satisfied, I will update each of these documents with the new content

src/encoding.rs
dralley: Sure, but it's just a test helper, I'd rather just leave it.

dralley:
That should resolve all of the feedback apart from doing any substantial structural changes to the commits.  I can look at that tomorrow, though if you are fine with the end result, it might make sense to just merge those commits.

src/encoding.rs
Mingun: I think, that more tight integration may not be necessary, because you already can do the following thing (it makes sense to add this to read_utf16 example)
```rust
    Ok(Event::Decl(e)) => {
        if let Some(encoding) = e.encoder() {
            reader.get_mut().set_encoding(encoding);
        }
    }
}
```
It seems to me that this the enough level of integration. In any case, the `Reader` are still a low-level primitive, it does not obligated to solve all your problems.

tests/encodings.rs
```diff
@@ -1,11 +1,141 @@
+#[cfg(feature = "encoding")]
```
Mingun: The whole test marked as required an `encoding` feature (in `Cargo.toml`), it is not required to add those feature gates.

tests/encodings.rs
```diff
@@ -219,66 +324,52 @@ mod detect {
...
-/// BOM indicates UTF-16LE, but XML - windows-1251
-#[test]
-fn bom_overridden_by_declaration() {
-    let mut reader = Reader::from_reader(b"\xFF\xFE<?xml encoding='windows-1251'?>".as_ref());
-    let mut buf = Vec::new();
+    /// Checks that encoding is detected by BOM and changed after XML declaration
+    /// BOM indicates UTF-16LE, but XML declares windows-1251
+    #[cfg(feature = "encoding")]
+    #[test]
+    fn bom_overridden_by_declaration() {
```
Mingun: It seems to me that moving those tests into `detect` module makes little sense:
- detect_encoding -- it is about free function `detect_encoding`, and not tests of reader (you may move it from `decode` module, because it is also makes little sense to have it in the dedicated module)
- bom_overridden_by_declaration
- only_one_declaration_changes_encoding
- str_always_has_utf8

If you wish, we may rename `detect` -> `detect_by_declaration`

Mingun (commented):
Thanks! After second look it seems to me that some changes in `tests/encodings.rs` are not necessary. Also, could you rebase over latest master? Automatic merge in current state will place some lines in `Changelog.md` into wrong place and it is better to keep history linear as much as possible.

dralley:
> I think, that more tight integration may not be necessary, because you already can do the following thing (it makes sense to add this to read_utf16 example)

> It seems to me that this the enough level of integration. In any case, the Reader are still a low-level primitive, it does not obligated to solve all your problems.

As a user, you can do that from the parsing loop yes, but I would still prefer that it not be necessary.  

The direction I would like to go in is to make `Reader` able to assume UTF-8 input, and then be able to use `&str` for `Event`, `Attribute`, `QName` etc.. rip out `Decoder` and `decode_*` from the API surface and use `DecodingReader` exclusively, and then potentially use a similar pattern to handle line ending normalization and perhaps line/column tracking.

Though the `Utf8ValidatingReader` is perhaps lower priority than I initially thought, because I hadn't noticed `BinaryStream`, and if we still want to support people dumping random binary blobs into XML files as we currently do, the only realistic way to support that is to just `str::from_utf8()` individual `Event`s

Which is "fine", just a larger (though still insignificant) performance penalty relative to doing so in larger blocks.

tests/documents/encoding/utf8.xml
dralley: @Mingun Would you like the `utf8.xml`, `utf16le.xml`, `utf16le-bom.xml` (etc) updated with the contents of the new file?

src/encoding.rs
dralley: Did you still want this changed?

Mingun: Not necessary

tests/documents/encoding/utf8.xml
Mingun: Yes, I think, that makes sense

tests/encodings.rs
dralley: On the other branch I removed the feature flag from the file at some point, to enable (some of these)  tests to run both when `encoding` is enabled and also when not enabled. (That is also why I put them in a separate module).  

However some of those changes don't necessarily make sense with this commit series, so I can update it.

tests/encodings.rs
dralley: See the other note.

dralley:
I removed the mentioned tests from the modules and removed the feature flags as per feedback.  Documents are updated, and latest commit makes a few additional test changes.

Mingun (approved):
Thanks! I made some minor improvements:
1. Some commits contains unnecessary changes, some of them was reverted later. I rewrite the history so this changes are never existed. This is better for blame purposes. That changes are
   - Accidentally delete an empty row in the `Changelog.md`
   - Formatting of test in `tests/encodings.rs`
2. Changes in `tests/encodings.rs` was split over 3 commits to simplify future blames
3. Add some forgotten and accidentally removed imports of `pretty_assertions::assert_eq`
4. I also slightly rearranged commits, so when you say "Needed for: Error::downcast() used in the next commit" it is actually in next commit :grin:
5. Direct links to `DecodingReader` produced broken link warnings from rustdoc even when run with `--all-features` (using the command from cheat sheet in `Changelog.md`), because that struct available only when encoding feature is enabled, but it is used in places that are not feature-gated. Replaced by usual HTML links
6. Change documentation of `encoding` feature to mention `DecodingReader` and add example in `DecodingReader` about usage
7. `detect_test!` tests are still useful, because they test that we can detect encoding even when working without `DecodingReader`, so I returned them. UTF-16 tests was broken and making XMLs more substantial revealed that, so they also changed to only detect encoding after reading the first event.

Mingun:
So, if there are no more very small changes that someone want to merge at the last minute, I will make a 0.40 release in the late afternoon (in about 6-8 hours).

Issue #950, opened 2026-03-16
Stability regression in v0.39.2: `attempt to subtract with overflow`


alex-security-guy:
Good day :)

I am a security researcher with [X41](https://x41-dsec.de/), and our team has recently performed a security review of [Routinator](https://github.com/NLnetLabs/routinator), which uses your project.
During this review we were performing some fuzz testing, and found what seems to be a stability regression from version `v.038.3` to `v0.39.2`.
The old version will respond to parsing the attached input with an `Error occurred Xml(Syntax(UnclosedDoctype))`, whereas the new version will panic with the aforementioned `attempt to subtract with overflow`.

The Routinator team was kind enough to extract only the `quick-xml` code, a minimal `main` of which you can find below:
```rust
use quick_xml::events::Event;
use quick_xml::reader::Reader;
use std::fs::File;
use std::io::BufReader;

fn main() {
    let f = File::open("crashing_notification.xml").unwrap();
    let reader = BufReader::new(f);
    let mut reader = Reader::from_reader(reader);
    let mut buf = Vec::new();
    loop {
        match reader.read_event_into(&mut buf) {
            Ok(Event::Eof) => break,
            _ => (),
        }
        buf.clear();
    }
}
```

And the crashing input you can find here: [crashing_notification.xml](https://github.com/user-attachments/files/26015825/crashing_notification.xml)

Hopefully this helps, I hope you have a nice day 😊

Mingun:
Minimal reproducible example:
```rust
/// Regression test for https://github.com/tafia/quick-xml/issues/950
#[test]
fn issue950() {
    let reader = BufReader::with_capacity(8, "<!d[<>23456789".as_bytes());
    //                                chunks: ________--------
    let mut reader = Reader::from_reader(reader);
    let mut buf = Vec::new();
    reader.read_event_into(&mut buf).unwrap();
}
```
Key point is that DTD skip code tries to read at least 10 characters before it would attempt to skip unknown markup, but that markup (`<>` here) is ended inside that 10-byte chunk:
- `<!d[<>12` -- the first chunk from `BufReader`, `<>12` is skipped as `UndecidedMarkup(3)`. Actually, unknown markup already read, but we wait at least 10 bytes to check if it is `<!NOTATION` (the longest possible valid markup in DTD)
[...]

PR #958, opened 2026-05-06
Fix slice-index panic in DtdParser on chunked unknown markup (#957)

### LINKED ISSUE #957: DtdParser panics with slice OOB on chunked DTD input that never matches a keyword (regression after #954)
**Affected version:** 0.39.3 (the same code is on `master`, line numbers below match both)

**Summary**

`Reader::read_event_into` aborts with

```
thread 'main' panicked at .../quick-xml-0.39.3/src/parser/dtd.rs:195:26:
range end index 11 out of range for slice of length 9
```

when an input has a `<!`-shaped token inside a DTD internal subset that never matches one of `<!--`, `<![CDATA[`, `<!ELEMENT`, `<!ATTLIST`, `<!ENTITY`, `<!NOTATION` *and* the underlying reader feeds the DTD parser in chunks. With `BufReader` (or any chunked I/O — e.g. fuzzers feeding bytes piece by piece) this is reachable from arbitrary user-supplied input, which makes it a DoS surface for any consumer reading untrusted XML.

This survives the fix from #954 / PR #954: that fix addressed the `attempt to subtract with overflow` panic. The state-machine bug it left behind is that `skipped` can grow past `bytes.len() == 9`, and the *next* re-entry into the `UndecidedMarkup` arm panics on a slice index instead.

**Minimal reproducer**

```rust
use quick_xml::Reader;
use quick_xml::events::Event;
use std::io::BufReader;

fn main() {
    let payload = b"<!DOCTYPE r[<aa<bb<cc<dd>";
    let reader = BufReader::with_capacity(4, &payload[..]);
    let mut reader = Reader::from_reader(reader);
    let mut buf = Vec::new();
    loop {
            Ok(Event::Eof) => break,
            Ok(_) => {}
            Err(_) => break,
        }
        buf.clear();
    }
}
```

`Cargo.toml`:

```toml
[dependencies]
quick-xml = "0.39.3"
```

Output:

```
thread 'main' panicked at .../quick-xml-0.39.3/src/parser/dtd.rs:195:26:
range end index 11 out of range for slice of length 9
```

The 4-byte `BufReader` cap is what forces `read_event_into` to feed the DTD parser across enough chunks for `skipped` to grow past 9. The same input fed in a single chunk parses fine.

**Root cause**

In `src/parser/dtd.rs`:

```rust
Self::UndecidedMarkup(skipped) => {
    let mut bytes = [0u8; 9];
    bytes[..skipped].copy_from_slice(&buf[buf.len() - skipped..]);   // panics when skipped > 9
    let end = bytes.len().min(skipped + cur.len());
    bytes[skipped..end].copy_from_slice(&cur[..end - skipped]);
    if let Some(skip) = self.switch(&bytes[..end]) {
        cur = &cur[skip - skipped..];
        continue;
    }
    *self = Self::UndecidedMarkup(skipped + cur.len());              // unbounded accumulation
    break;
}
```

When `self.switch(&bytes[..end])` returns `None` (no keyword match), the state is updated to `UndecidedMarkup(skipped + cur.len())` — `skipped` is *not* clamped to `bytes.len()` (= 9). On the next chunk, the very first line `bytes[..skipped].copy_from_slice(...)` panics on the slice index. The comment a few lines up (*"Buffer is long enough to store the longest possible keyword `!NOTATION`"*) implies `skipped` should never exceed 9, but the state-update doesn't enforce that.

**Why this matters**

Any application that uses `quick_xml::Reader` on untrusted XML can be panicked by ~25 bytes of crafted input. We hit this in our project via a coverage-guided fuzzer feeding `BufReader<File>`; downstream we worked around it by pre-rejecting `<!`-shaped tokens that aren't `<!--` or `<![CDATA[` before the reader sees them, but the state-machine invariant should be enforced in the library so consumers don't have to guess.

dobermai:
Fixes #957.

**Bug**

In `src/parser/dtd.rs`, the `Self::UndecidedMarkup(skipped)` arm of `DtdParser::feed` stages the prefix of an unresolved markup token into a fixed `[u8; 9]` work buffer (long enough for the longest recognised keyword, `!NOTATION`). When the buffered window did not match any keyword *and* did not contain a `>`, the state was updated to `UndecidedMarkup(skipped + cur.len())` and `feed` returned to wait for more bytes.

With chunked input that keeps hitting `<!`-shaped tokens that never match one of `<!--`, `<![CDATA[`, `<!ELEMENT`, `<!ATTLIST`, `<!ENTITY`, `<!NOTATION`, `skipped` could grow past `bytes.len() == 9`. The next re-entry into the arm panicked on the very first line, `bytes[..skipped].copy_from_slice(&buf[buf.len() - skipped..])`, with

```
range end index N out of range for slice of length 9
```

This survived [#954]'s fix for #950 (different panic, same state machine).

**Fix**

Once the 9-byte window is full (`end == bytes.len()`) and `switch()` returned `None`, the markup is definitively not one of the recognised keywords. There is no benefit to waiting for more data; the parser can drop the examined window and skip until the closing `>`. The `InElementDecl` state already does exactly that (no quote awareness needed — we only reach this point if we *failed* to match the keywords that promote into `InQuoteSensitive`), so the new branch reuses it rather than introducing a new state.

```rust
if end == bytes.len() {
    cur = &cur[end - skipped..];
    *self = Self::InElementDecl;
    continue;
}
```

The shape of the existing comments — *\"Buffer is long enough to store the longest possible keyword `!NOTATION`\"* — already implies `skipped` was meant to be bounded by `bytes.len()`; this just enforces it.

**Test**

Added `issue957` in `tests/issues.rs` in the same style as `issue950`. A 25-byte malformed DTD plus `BufReader::with_capacity(4)` is enough to force the failure mode (each `<aa`, `<bb`, … straddles a chunk boundary, so `UndecidedMarkup` is re-entered repeatedly):

```rust
let reader = BufReader::with_capacity(4, "<!DOCTYPE r[<aa<bb<cc<dd>".as_bytes());
//                                chunks: ____----____----____----_
```

Same disposition as `issue950` — we don't assert on the returned event since proper DTD validity reporting is a future improvement; we just pin that the parser doesn't panic.

**Verification**

- New regression test passes on the fix; reverting the fix reproduces the panic.
- `cargo test` (full suite, all features default) stays green: doc tests + every `tests/*` binary, no failures.

**Why this matters**

Any consumer reading untrusted XML through `Reader::read_event_into` (with a `BufReader` or any chunked `BufRead`) can be panicked by ~25 bytes of crafted input. We hit it via a coverage-guided fuzzer feeding `BufReader<File>` over arbitrary bytes; the workaround downstream is to pre-reject `<!`-shaped tokens that aren't comments / CDATA before the reader sees them, but the state-machine invariant should be enforced in the library.

[#954]: https://github.com/tafia/quick-xml/pull/954

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -37,6 +37,10 @@
 
 - [#938]: Use correct rules for EOL normalization in `Deserializer` when parse XML 1.0 documents.
   Previously XML 1.1. rules was applied.
+- [#957]: Fix slice-index panic when reading malformed DTD whose unknown markup
+  is split across `BufReader` chunks. As with [#950], the returned
+  `Event::DocType` may contain the malformed DTD; this fix only ensures that
+  the parser does not panic.
 
 ### Misc Changes
 
@@ -51,6 +55,7 @@
 [#914]: https://github.com/tafia/quick-xml/pull/914
 [#938]: https://github.com/tafia/quick-xml/pull/938
 [#944]: https://github.com/tafia/quick-xml/pull/944
+[#957]: https://github.com/tafia/quick-xml/issues/957
 
 
 ## 0.39.3 -- 2026-05-04
--- a/src/parser/dtd.rs
+++ b/src/parser/dtd.rs
@@ -205,6 +205,17 @@ impl DtdParser {
                         cur = &cur[skip - skipped..];
                         continue;
                     }
+                    // No keyword matched. If we have a full 9-byte window the
+                    // markup is definitively not one of `<!--`, `<![CDATA[`,
+                    // `<!ELEMENT`, `<!ATTLIST`, `<!ENTITY`, `<!NOTATION`, so
+                    // fall back to skipping until the closing `>` instead of
+                    // accumulating `skipped` past `bytes.len()` (which would
+                    // panic on the slice-copy above on a later iteration).
+                    if end == bytes.len() {
+                        cur = &cur[end - skipped..];
+                        *self = Self::InElementDecl;
+                        continue;
+                    }
                     *self = Self::UndecidedMarkup(skipped + cur.len());
                     break;
                 }
--- a/tests/issues.rs
+++ b/tests/issues.rs
@@ -666,3 +666,23 @@ fn issue950() {
     // but just ensure that we do not panic
     let _ = reader.read_event_into(&mut buf);
 }
+
+/// Regression test for https://github.com/tafia/quick-xml/issues/957
+#[test]
+fn issue957() {
+    // Each `<aa`, `<bb`, ... is unknown markup that does not match any of
+    // `<!--`, `<![CDATA[`, `<!ELEMENT`, `<!ATTLIST`, `<!ENTITY`, `<!NOTATION`
+    // and is split across `BufReader` chunks, so the DTD parser re-enters
+    // the `UndecidedMarkup` arm multiple times in a row. Before the fix,
+    // each re-entry grew `skipped` by `cur.len()` until the slice copy
+    // `bytes[..skipped]` against the 9-byte work buffer panicked with
+    // `range end index N out of range for slice of length 9`.
+    let reader = BufReader::with_capacity(4, "<!DOCTYPE r[<aa<bb<cc<dd>".as_bytes());
+    //                                chunks: ____----____----____----_
+    let mut reader = Reader::from_reader(reader);
+    let mut buf = Vec::new();
+    // Same disposition as `issue950`: malformed DTD must not panic; we do
+    // not assert on the returned event since DTD validity reporting is a
+    // future improvement.
+    let _ = reader.read_event_into(&mut buf);
+}


Mingun:
Thanks! Yes, that is elegant solution.

It is bad that this was not catched by our fuzzing tests. I haven't good understanding how they work. Improving the testset would be good. You say, that this bug was catched by fuzzer. Could you share the details of how it works for you, or maybe you even want to do PR to improve our fuzzing procedure?

dobermai:
Happy to help, and good question.

Nothing fancy on our side — it's just `cargo-fuzz`/libFuzzer, same as yours. The one thing that's different is that we feed the parser via `BufReader<File>` rather than `Cursor<&[u8]>`. `<Cursor as BufRead>::fill_buf` always hands back the whole remaining input in one go, so a `Cursor`-backed harness never gets to exercise parser states that span `fill_buf` calls. That's the whole reason #950 and #957 slipped through — both regression tests reproduce only with `BufReader::with_capacity(small_N, ..)`.

I went ahead and opened #959 with a small `fuzz_chunked_reader` target. It just takes the first input byte as a `BufReader` capacity and feeds the rest as XML — same input shape as `fuzz_target_1`, so corpora are interchangeable. I sanity-checked it by reverting the #957 fix locally and seeding the corpus with the regression-test input prefixed with the capacity byte; libFuzzer reproduced the panic on the seed within one run, and with the fix back in place 5000 runs stay clean. All the details are in the PR.

Hope it's useful!

PR #959, opened 2026-05-06
Add fuzz_chunked_reader target for chunk-boundary parser bugs

dobermai:
Follow-up to #958 / #957, and a response to your request in https://github.com/tafia/quick-xml/pull/958#issuecomment-3520760117 about improving the fuzzing setup.

## Why

`fuzz_target_1` drives `Reader::from_reader` via `Cursor::new(data)`. `<Cursor as BufRead>::fill_buf` always returns the entire remaining input in a single window, so the harness never exercises parser states that span `fill_buf` calls.

That gap is exactly what hid #950 and #957. Both regression tests in `tests/issues.rs` reproduce by wrapping the input in `BufReader::with_capacity(N, ..)` for a small N — the parser has to resume across chunk boundaries, and state that's only valid within a single `fill_buf` window can grow past its bounds. Neither bug is reachable from a `Cursor`-backed harness, regardless of fuzz budget.

## What

Adds `fuzz_chunked_reader`. The first input byte is the `BufReader` capacity (clamped to `1..=255`); the rest is fed as XML. Same input shape as `fuzz_target_1`, so corpora can be shared between the two.

The `1..=255` range is deliberate — capacities `>= data.len()` degenerate to the `Cursor` case already covered by `fuzz_target_1`, so this is the most efficient use of the fuzz budget for the chunk-boundary regime.

## Verification

I reverted 9fd1d58 (the #957 fix) on a local branch, seeded `fuzz/corpus/fuzz_chunked_reader/` with `tests/issues.rs::issue957`'s input prefixed with `0x04` (the capacity byte), and ran `cargo fuzz run fuzz_chunked_reader corpus/fuzz_chunked_reader`. libFuzzer reproduced the `slice_index_fail` in `DtdParser::feed` on the seed within one run — same panic site as the original bug.

With the fix back in place, 5000 runs over the seeded corpus are clean.

## Notes

- No corpus is committed — `fuzz/.gitignore` already excludes `corpus/`.
- No new dependencies; reuses existing `arbitrary` / `libfuzzer-sys`.
- `fuzz/README.md` updated to describe the three targets and how they differ.
- Doesn't touch `fuzz_target_1` or `structured_roundtrip` — keeping each target single-purpose.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -31,3 +31,9 @@ name = "structured_roundtrip"
 path = "fuzz_targets/structured_roundtrip.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "fuzz_chunked_reader"
+path = "fuzz_targets/fuzz_chunked_reader.rs"
+test = false
+doc = false
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -5,3 +5,19 @@ cargo fuzz run -O -j4 fuzz_target_1
 ```
 
 See also: https://github.com/rust-fuzz/cargo-fuzz
+
+## Targets
+
+- `fuzz_target_1` — drives `Reader::from_reader` over `Cursor<&[u8]>` and
+  `Reader::from_str`. Broad coverage of event-decoding and writer
+  round-tripping. Fast per-execution.
+- `structured_roundtrip` — uses `arbitrary` to build a writer program,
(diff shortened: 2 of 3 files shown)


dobermai:
Unfortunately (or fortunately?) the new fuzz runs seems to have surfaced another issue. I'll give it a shot to fix. Opened #960 to track

Mingun:
Thanks!

PR #963, opened 2026-05-11
Improve ergonomics - rebuild Reader around native utf-8 string types

dralley:
- [x] Add UTF-8 validation in Reader internals for parsing events
- [x] Change name / namespace types from &[u8] to &str
- [x] Change event types from `Cow<[u8]>` to `Cow<str>`, remove Decoder
- [x] Change attribute types from `Cow<[u8]>` to `Cow<str>`
- [x] Remove Decoder & methods from public API

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -92,6 +92,10 @@
   `BytesCData`, `BytesRef`) and `Attributes`. The `decoder()` method is no longer available
   on these types. Decode methods on events now always assume UTF-8 input.
   `Error::missed_end()` no longer takes a `Decoder` parameter.
+- [#963]: Event types (`BytesStart`, `BytesEnd`, `BytesText`, `BytesCData`, `BytesPI`,
+  `BytesRef`) now store `Cow<str>` internally instead of `Cow<[u8]>`. `into_inner()` on
+  `BytesText`, `BytesCData`, `BytesPI`, and `BytesRef` now returns `Cow<str>`.
+  `BytesStart::set_name()` now takes `&str` instead of `&[u8]`.
 
 [#963]: https://github.com/tafia/quick-xml/pull/963
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -98,6 +98,9 @@
   `BytesRef`) now store `Cow<str>` internally instead of `Cow<[u8]>`. `into_inner()` on
   `BytesText`, `BytesCData`, `BytesPI`, and `BytesRef` now returns `Cow<str>`.
   `BytesStart::set_name()` now takes `&str` instead of `&[u8]`.
+- [#963]: All event types and the `Event` enum now implement `Deref<Target = str>`
+  instead of `Deref<Target = [u8]>`. Explicit `AsRef<str>` impls are provided to
+  avoid ambiguity.
 
 [#963]: https://github.com/tafia/quick-xml/pull/963
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -84,6 +84,10 @@
   Non-UTF-8 input passed to `Reader::from_reader()` without `DecodingReader` will now
   produce `Error::Encoding` instead of silently passing through invalid bytes.
   Use `DecodingReader` to transcode non-UTF-8 sources.
+- [#963]: Name types (`QName`, `LocalName`, `Prefix`, `Namespace`, `PrefixDeclaration`)
+  now wrap `&str` instead of `&[u8]`. `into_inner()` returns `&str`, and `AsRef<str>`
+  is implemented (`AsRef<[u8]>` has been removed). `ResolveResult::Unknown` now contains `String`
+  instead of `Vec<u8>`, and `NamespaceError` variants contain `String` instead of `Vec<u8>`.
 
 [#963]: https://github.com/tafia/quick-xml/pull/963
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -78,6 +78,14 @@
 
 ### Misc Changes
 
+### Breaking Changes
+
+- [#963]: Reader now validates that input is valid UTF-8 when constructing events.
+  Non-UTF-8 input passed to `Reader::from_reader()` without `DecodingReader` will now
+  produce `Error::Encoding` instead of silently passing through invalid bytes.
+  Use `DecodingReader` to transcode non-UTF-8 sources.
+
+[#963]: https://github.com/tafia/quick-xml/pull/963
 
 ## 0.40.0 -- 2026-05-11
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,76 +14,12 @@
 
 ## Unreleased
 
-### New Features
+This is a large release. The primary change is an ergonomic improvement across the entire API -
+the quick_xml now makes use of `&str` and `String` types where possible instead of
+`&[u8]` and `Vec<u8>`. This requires significant refactoring of downstream code,
+but should result in a net simplification as well as potential performance improvements.
 
-### Bug Fixes
-
-- [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
-  event) now returns the new `NamespaceError::TooDeeplyNested` when a document
-  nests elements deeper than `u16::MAX`, instead of overflowing the internal
-  `u16` depth counter. Previously the unguarded `nesting_level += 1` panicked
-  under `overflow-checks` builds and silently wrapped in release, corrupting
-  namespace-scope bookkeeping on deeply nested untrusted input.
-
-### Misc Changes
-
-- [#983]: Adopted an AI use and contribution policy for new upstream contributions.
-
-[#977]: https://github.com/tafia/quick-xml/issues/977
-[#983]: https://github.com/tafia/quick-xml/issues/983
-
-## 0.41.0 -- 2026-06-29
-
-### New Features
-
-- [#970]: Add `NsReader::resolver_mut()` and
-  `NamespaceResolver::{max_declarations_per_element, set_max_declarations_per_element}`.
-
-### Bug Fixes
-
-- [#969]: `Attributes` (and anything that iterates `BytesStart::attributes()`
-  with the default `with_checks(true)`) no longer takes O(N²) time on a start
-  tag with a large number of attributes. Small tags keep the previous linear
-  scan; larger ones switch to a 64-bit hash pre-filter, so the whole tag is
-  O(N). The exact `AttrError::Duplicated(new, prev)` positions are unchanged.
-- [#970]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
-  event) now rejects a start tag that declares more than
-  `DEFAULT_MAX_DECLARATIONS_PER_ELEMENT` (256) `xmlns` / `xmlns:*` namespace
-  bindings, returning the new `NamespaceError::TooManyDeclarations`. Previously
-  `push` allocated one `NamespaceBinding` per declaration with no upper bound,
-  before the event was returned to the caller, so an `NsReader` consumer could
-  not bound its memory exposure on untrusted input. The limit is configurable
-  via `NamespaceResolver::set_max_declarations_per_element` (use `usize::MAX`
-  to disable).
-
-[#969]: https://github.com/tafia/quick-xml/issues/969
-[#970]: https://github.com/tafia/quick-xml/issues/970
-
-
-## 0.40.1 -- 2026-05-15
-
-### Bug Fixes
-
-- [#964]: Fix `unreachable!()` panic in the serde deserializer when a DOCTYPE
-  declaration appears between two text runs inside an element (e.g.
-  `<a>x<!DOCTYPE y>z</a>`). The DOCTYPE used to break `drain_text`'s
-  consecutive-text merge, so two `DeEvent::Text` events reached
-  `read_text` and tripped its "Cannot be two consequent Text events"
-  invariant. DOCTYPE is now treated as transparent during text drain —
-  it still goes through the entity resolver, but the surrounding text
-  is merged into one run. Discovered via libFuzzer on a real-world
-  SAML deserializer harness.
-
-[#964]: https://github.com/tafia/quick-xml/pull/964
-
-### Misc Changes
-
-- [#963]: MSRV bumped to 1.86 (April 2025)
-- [#963]: Deprecated `Attribute` methods that take a `Decoder` parameter, since
-  attribute values are now always valid UTF-8: `decoded_and_normalized_value()`,
-  `decoded_and_normalized_value_with()`, `decode_and_unescape_value()`, and
-  `decode_and_unescape_value_with()`. Use `normalized_value()` and
-  `normalized_value_with()` instead.
+The MSRV has been raised to 1.86.
 
 ### Breaking Changes
 
@@ -119,13 +55,77 @@
   serde trait. Removed all methods from `Decoder` (the struct is kept only for
   backward compatibility with deprecated `Attribute` methods).
 
+### Bug Fixes
+
+- [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
+  event) now returns the new `NamespaceError::TooDeeplyNested` when a document
+  nests elements deeper than `u16::MAX`, instead of overflowing the internal
+  `u16` depth counter. Previously the unguarded `nesting_level += 1` panicked
+  under `overflow-checks` builds and silently wrapped in release, corrupting
+  namespace-scope bookkeeping on deeply nested untrusted input.
+
+### Misc Changes
+
+- [#983]: Adopted an AI use and contribution policy for new upstream contributions.
+- [#963]: MSRV bumped to 1.86 (April 2025)
+- [#963]: Deprecated `Attribute` methods that take a `Decoder` parameter, since
+  attribute values are now always valid UTF-8: `decoded_and_normalized_value()`,
+  `decoded_and_normalized_value_with()`, `decode_and_unescape_value()`, and
+  `decode_and_unescape_value_with()`. Use `normalized_value()` and
+  `normalized_value_with()` instead.
+
 [#963]: https://github.com/tafia/quick-xml/pull/963
+[#977]: https://github.com/tafia/quick-xml/issues/977
+[#983]: https://github.com/tafia/quick-xml/issues/983
+
+## 0.41.0 -- 2026-06-29
+
+### New Features
+
+- [#970]: Add `NsReader::resolver_mut()` and
+  `NamespaceResolver::{max_declarations_per_element, set_max_declarations_per_element}`.
+
+### Bug Fixes
+
+- [#969]: `Attributes` (and anything that iterates `BytesStart::attributes()`
+  with the default `with_checks(true)`) no longer takes O(N²) time on a start
+  tag with a large number of attributes. Small tags keep the previous linear
+  scan; larger ones switch to a 64-bit hash pre-filter, so the whole tag is
+  O(N). The exact `AttrError::Duplicated(new, prev)` positions are unchanged.
+- [#970]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
+  event) now rejects a start tag that declares more than
+  `DEFAULT_MAX_DECLARATIONS_PER_ELEMENT` (256) `xmlns` / `xmlns:*` namespace
+  bindings, returning the new `NamespaceError::TooManyDeclarations`. Previously
+  `push` allocated one `NamespaceBinding` per declaration with no upper bound,
+  before the event was returned to the caller, so an `NsReader` consumer could
+  not bound its memory exposure on untrusted input. The limit is configurable
+  via `NamespaceResolver::set_max_declarations_per_element` (use `usize::MAX`
+  to disable).
+
+[#969]: https://github.com/tafia/quick-xml/issues/969
+[#970]: https://github.com/tafia/quick-xml/issues/970
+
+## 0.40.1 -- 2026-05-15
+
+### Bug Fixes
+
+- [#964]: Fix `unreachable!()` panic in the serde deserializer when a DOCTYPE
+  declaration appears between two text runs inside an element (e.g.
+  `<a>x<!DOCTYPE y>z</a>`). The DOCTYPE used to break `drain_text`'s
+  consecutive-text merge, so two `DeEvent::Text` events reached
+  `read_text` and tripped its "Cannot be two consequent Text events"
+  invariant. DOCTYPE is now treated as transparent during text drain —
+  it still goes through the entity resolver, but the surrounding text
+  is merged into one run. Discovered via libFuzzer on a real-world
+  SAML deserializer harness.
+
+[#964]: https://github.com/tafia/quick-xml/pull/964
 
 ## 0.40.0 -- 2026-05-11
 
 MSRV bumped to 1.79.
 
-Now `quick-xml` supports the UTF-16 encoded documents. See the new `DecodingReader` type.
+Now `quick-xml` supports UTF-16 encoded documents. See the new `DecodingReader` type.
 
 ### New Features
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -79,6 +79,11 @@
 ### Misc Changes
 
 - [#963]: MSRV bumped to 1.86 (April 2025)
+- [#963]: Deprecated `Attribute` methods that take a `Decoder` parameter, since
+  attribute values are now always valid UTF-8: `decoded_and_normalized_value()`,
+  `decoded_and_normalized_value_with()`, `decode_and_unescape_value()`, and
+  `decode_and_unescape_value_with()`. Use `normalized_value()` and
+  `normalized_value_with()` instead.
 
 ### Breaking Changes
 
@@ -105,6 +110,10 @@
   Content is already available as `&str` via `Deref`. The `xml10_content()`,
   `xml11_content()`, `xml_content()`, and `html_content()` methods now return
   `Cow<str>` directly instead of `Result<Cow<str>, EncodingError>`.
+- [#963]: `Attribute::value` is now `Cow<'a, str>` instead of `Cow<'a, [u8]>`.
+  The `From<(&[u8], &[u8])>` impl has been removed.
+- [#963]: `BytesDecl::version()`, `encoding()`, and `standalone()` now return
+  `Cow<'_, str>` instead of `Cow<'_, [u8]>`.
 
 [#963]: https://github.com/tafia/quick-xml/pull/963
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -88,6 +88,10 @@
   now wrap `&str` instead of `&[u8]`. `into_inner()` returns `&str`, and `AsRef<str>`
   is implemented (`AsRef<[u8]>` has been removed). `ResolveResult::Unknown` now contains `String`
   instead of `Vec<u8>`, and `NamespaceError` variants contain `String` instead of `Vec<u8>`.
+- [#963]: Removed the `decoder: Decoder` field from event types (`BytesStart`, `BytesText`,
+  `BytesCData`, `BytesRef`) and `Attributes`. The `decoder()` method is no longer available
+  on these types. Decode methods on events now always assume UTF-8 input.
+  `Error::missed_end()` no longer takes a `Decoder` parameter.
 
 [#963]: https://github.com/tafia/quick-xml/pull/963
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -114,6 +114,10 @@
   The `From<(&[u8], &[u8])>` impl has been removed.
 - [#963]: `BytesDecl::version()`, `encoding()`, and `standalone()` now return
   `Cow<'_, str>` instead of `Cow<'_, [u8]>`.
+- [#963]: Removed `Reader::decoder()` method. Use `Reader::encoding()` instead
+  (available with the `encoding` feature). Removed `decoder()` from the `XmlRead`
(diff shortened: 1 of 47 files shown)


src/de/key.rs
```diff
@@ -110,14 +107,16 @@ impl<'i, 'd> QNameDeserializer<'i, 'd> {
...
-            Cow::Borrowed(b) => match decode_name(QName(&b[..start.name_len]), start.decoder())? {
-                Cow::Borrowed(borrowed) => CowRef::Input(borrowed),
-                Cow::Owned(owned) => CowRef::Owned(owned),
-            },
-            Cow::Owned(ref o) => match decode_name(QName(&o[..start.name_len]), start.decoder())? {
-                Cow::Borrowed(borrowed) => CowRef::Slice(borrowed),
-                Cow::Owned(owned) => CowRef::Owned(owned),
-            },
+            Cow::Borrowed(b) => {
+                let name_str = std::str::from_utf8(&b[..start.name_len])
```
dralley: There will be a handful of these temporary `from_utf8()` calls, but they should be able to be removed by by subsequent commits as additional types are switched over.

dralley:
@Mingun Would you be satisfied if `BinaryStream` / raw byte buffers were only supported when the rest of the document apart from those buffers is UTF-8?

edit: well, that's what I implemented.

Sidenote: maybe `decoded_and_normalized_value()` (etc.) ought to be marked deprecated in 0.40.1, to point people in the direction of using `DecodingReader`

dralley:
These 3 particular commits are ready for review, with the caveat that there will be (probably) 6-8 additional commits coming.

src/utils.rs
```diff
@@ -341,51 +334,36 @@ pub const fn name_len(mut bytes: &[u8]) -> usize {
...
-pub const fn trim_xml_start(mut bytes: &[u8]) -> &[u8] {
-    // Note: A pattern matching based approach (instead of indexing) allows
-    // making the function const.
-    while let [first, rest @ ..] = bytes {
-        if is_whitespace(*first) {
-            bytes = rest;
-        } else {
-            break;
-        }
+pub fn trim_xml_start(s: &str) -> &str {
```
dralley: @Mingun Keeping this as `const` would, I think, require bumping the MSRV to 1.86 (for `const fn split_at`) unless we use `unsafe` here.

dralley: That would look like 

```Rust
pub const fn trim_xml_start(s: &str) -> &str {
    let mut bytes = s.as_bytes();
    while let [first, rest @ ..] = bytes {
        if is_whitespace(*first) {
            bytes = rest;
        } else {
            break;
        }
    }
    // SAFETY: trimming ASCII whitespace from a valid UTF-8 string always yields valid UTF-8
    unsafe { core::str::from_utf8_unchecked(bytes) }
}
```

I went ahead with the safe `split_at` implementation however

src/events/attributes.rs
```diff
@@ -15,27 +14,23 @@ use std::{borrow::Cow, ops::Range};
...
     pub key: QName<'a>,
     /// The raw value of the attribute.
-    pub value: Cow<'a, [u8]>,
+    pub value: Cow<'a, str>,
 }
 
 impl<'a> Attribute<'a> {
     /// Returns the attribute value normalized as per [the XML specification] (or [for 1.0]).
     ///
-    /// The document **must** be UTF-8 encoded, or pre-processed using [`DecodingReader`].
```
dralley, line 39: This is still true, but it doesn't really need to be said here, since the `Reader` itself now has this requirement.

dralley:
Remaining design questions, not all of which actually need to be dealt with in this PR:

* Should `Reader` and / or `Deserializer` wrap a `DecodingReader` automatically if the `encoding` feature is enabled?
    * If we do, should we use `from_utf8_unchecked()` (since input is pre-validated)?
    * Should some form of built-in stream decoding or stream validation be the "default", with a special feature to disable it for access to `BinaryStream` / `Reader::stream()`?
    * Should `slice_reader` be converted to `&str`-only, with a reader over `&[u8]` using the `buffered_reader` path?
* Should `BytesStart`, `BytesEnd`, `BytesText` (etc.) be renamed since they are no longer raw bytes but rather guaranteed UTF-8?
    *  e.g. `XmlStart`, `XmlEnd`, `XmlText` or just `Start`, `End`, `Text` (which would maybe be ambiguous given `Event::Text`, `Event::Start` etc.)
* Would introducing a `Utf8ValidatingReader` be worthwhile, or is the `encoding_rs` dependency not that big a deal?
* Should a similar built-in wrapper of the inner `Reader` be used to track position within the file globally (line numbers, etc.) and perform EOL normalization?

dralley:
Also, I can improve the commit messages and Changelog entries if needed.  The initial are pretty... concise.

Mingun:
First, I would prefer to keep the ability to parse non-utf8 encoded documents without recoding. XML itself can be parsed without knowing the exact encoding, it is enough if it is XML-compatible (which is all legacy 1-byte encodings that we support). So, is it possible to create a separate reader and event which will be always UTF-8 encoded and keep the current ones for advanced usage? It is fine to promote the new UTF-8-based reader as default, but keep the ability to work with non-UTF-8 input without recoding.

Here is the same situation as for regexp -- although it is defined in terms of strings, nothing prevents it from running on top of any byte arrays. The author of regexp engine even created a [`bstr`](https://docs.rs/bstr/latest/bstr/) crate to add useful string-based methods to byte arrays.

dralley:
> First, I would prefer to keep the ability to parse non-utf8 encoded documents without recoding. 

IMO, it is not worth the ergonomic and maintenance costs.  If you look at all the major XML parsing libraries like `libxml2`, `expat`, `encoding/xml` (Go), and Jackson (Java) etc, they all do internal transcoding and throw errors if they encounter something that can't be decoded (or or replaced) - with no escape hatch.  I suspect if this was a significant use case we would likely not be the only ones catering to it.

e.g. 

libxml2 parses & handles UTF-8 only, performs a streaming decode of other encodings
https://dev.w3.org/XInclude-Test-Suite/libxml2-2.4.24/libxml2-2.4.24/doc/encoding.html

expat selects either UTF-8 or UTF-16 as an internal encoding at compile time, decodes to that, returns whichever type of string was selected
https://libexpat.github.io/doc/expat-internals-encodings/ 

encoding/xml is the same as libxml2 - utf-8 only
https://pkg.go.dev/encoding/xml (search `CharsetReader`)

Decoding is very very fast relative to XML parsing - it varies depending on encoding and the precise makeup of the document of course, but generally between 15 and 90 Gbps, whereas XML parsing is currently in the ballpark of 0.5 Gbps and often slower, so I don't really think that's a reason to avoid it either.

I would maybe accept the argument that it's a huge API change and it might be warranted to support both for some time to allow a migration, but even then it would likely be easier to just maintain an older branch for a longer period of time. 

Duplicating the reader would, I think, be way way more work than it's worth.

dralley:
Also, the reason the XML libraries work that way, apart from overall simplicity, is that the XML standard effectively requires working that way.  The standard actually said that all XML processors should be able to handle
 either UTF-16 or UTF-8, and have mandatory fatal decoding errors in many situations, the easiest way to satisfy the requirements is to just do what everyone does, which is decode the document up front, and build a parser against one canonical encoding.

I'm not a complete stickler for compliance, and we do provide a handful of features catering to noncompliant XML and XML-derived document formats (which is fine), but in this case I really don't see a good reason to go out of our way to break with it.  It's just more complexity for a use case of (IMO) very questionable value.

https://www.w3.org/TR/xml/

Section 2.2

> The mechanism for encoding character code points into bit patterns may vary from entity to entity. **All XML processors MUST accept **the UTF-8 **and UTF-16** encodings of Unicode [[Unicode]](https://www.w3.org/TR/xml/#Unicode); the mechanisms for signaling which of the two is in use, or for bringing other encodings into play, are discussed later, in [4.3.3 Character Encoding in Entities](https://www.w3.org/TR/xml/#charencoding).

Section 4.3.3 - Character Encoding in Entities

>Each external parsed entity in an XML document may use a different encoding for its characters. All XML processors MUST be able to read entities in both the UTF-8 and UTF-16 encodings. 

...

> It is a [fatal error](https://www.w3.org/TR/xml/#dt-fatal) when an XML processor encounters an entity with an encoding that it is unable to process. **It is a [fatal error](https://www.w3.org/TR/xml/#dt-fatal) if an XML entity is determined (via default, encoding declaration, or higher-level protocol) to be in a certain encoding but contains byte sequences that are not legal in that encoding. Specifically, it is a fatal error if an entity encoded in UTF-8 contains any ill-formed code unit sequences, as defined in section 3.9 of Unicode [[Unicode]](https://www.w3.org/TR/xml/#Unicode).** Unless an encoding is determined by a higher-level protocol, it is also a [fatal error](https://www.w3.org/TR/xml/#dt-fatal) if an XML entity contains no encoding declaration and its content is not legal UTF-8 or UTF-16.

dralley:
@Mingun Thoughts?  

I understand your view, but duplicating the readers would end up being a lot of work I think, and I don't see a reason why we should have a different architecture than every other major XML parsing library.  Even trying to put myself in the position of someone that needs to parse legacy encodings, I can't see why I would want to have to deal with those encodings in Rust.

dralley:
Are you having any issues with Github access?

Mingun:
> Are you having any issues with Github access?

No, just not every day enough time to read everything that appeared and answer.

> duplicating the readers would end up being a lot of work I think

It seems to me that it is difficult to judge this without trying. It doesn't seem like that to me. The core reader will work with bytes, only event and and attribute types will need to be duplicated. It's not a lot of code.

For editing purposes we may want to be able to resave part of document, keeping the original encoding. We also may want to report byte offsets into the original source from which the document is parsed, which may be hard if we will reencode document always. I think, we need to think about this problems and how to handle them (or not handle at all). I would prefer not to pessimize ahead of time.

I need some time (perhaps of several months, because sometimes you need to take a break from the project in order to then take on it with renewed vigor) to review this and think about / experiment with possible alternative implementations. I will try to finish review for this PR as soon as possible.

> I can't see why I would want to have to deal with those encodings in Rust.

For example, if we want to save memory. UTF-8 could be more that 2x times more memory than using legacy encodings. This can be critical if you need to keep large files in memory for some reason.

dralley:
Take your time on the review.  However, re: your points -

Counterarguments:

> It seems to me that it is difficult to judge this without trying. It doesn't seem like that to me. The core reader will work with bytes, only event and and attribute types will need to be duplicated. It's not a lot of code.

The amount of code that works with those types is pretty significant though.  They'd still be different types ultimately, so even if the meat of the implementation is shared, you still need either duplicate definitions of all the user-facing methods of those types including duplicate doc comments that need to be kept in-sync; or else making a much larger portion of the codebase into macros than is already the case (which I wouldn't want to do).   Tests are also an issue, because they can only test one API at a time, unless again everything becomes a macro, or you just pick one variant to test thoroughly and only trivially exercise the other.

Either way, I think that ends up being a significant maintenance cost and challenge.  That's just my impression, maybe I'm wrong.

>   For example, if we want to save memory. UTF-8 could be more that 2x times more memory than using legacy encodings. This can be critical if you need to keep large files in memory for some reason.

quick-xml is already a streaming parser, which already puts it ahead of most parser options.  Memory use would only be a major concern if the user needed to accumulate large portions of the document in memory anyway.  And I would have to imagine it unlikely that a document in a legacy encoding would be so large in the first place, because such documents are usually somewhat constrained by the resources available during the time of its vintage.  There's not a lot of new documents of those encodings being written since the 2000s.

Even if that was not the case - due to the streaming nature and the fact that no other XML library I can find permits this (much less streaming ones, specifically), quick-xml would *still* be one of the best options.

> For editing purposes we may want to be able to resave part of document, keeping the original encoding.  We also may want to report byte offsets into the original source from which the document is parsed, which may be hard if we will reencode document always. I think, we need to think about this problems and how to handle them (or not handle at all). I would prefer not to pessimize ahead of time.

I take this to mean "parse to position, obtain byte position, acquire a handle and seek + overwrite within the file".  I guess that's possible, but has limited utility (it could only be used to overwrite exactly the same # of bytes) and comes at higher risk of corrupting the file.   (On top of possible mistakes, it's also more susceptible to TOCTOU - maybe not so much in the "exploit" sense but in the correctness sense).

However it's also not that difficult or expensive to reverse-engineer the input buffer positions from a little bit of state that we already need to keep track of.  We know the position of the start of the input buffer, we know the parser position in the output stream, it's just an exercise of counting characters on both sides,  which can be optimized in various ways to minimize re-counting.

e.g. with single byte input encodings, the character count between the start of the output buffer and the current parser position *is* the byte offset (which you can add to the position of the beginning of the input buffer).  The infrastructure for this is mostly identical to what needs to be added for proper reporting of line + column as well - something we can do much more reliably with a uniform input encoding.

Anyway, I'd like to not pessimize the common cases in favor of uncommon ones - I can't say that _nobody_ would use those things, but I strongly suspect that more users would appreciate improved spec compliance and more convenient APIs.

dralley:
A few more points:

>For example, if we want to save memory. UTF-8 could be more that 2x times more memory than using legacy encodings. This can be critical if you need to keep large files in memory for some reason.

No legacy encoding actually achieves more than about 45% savings over UTF-8 for any language, in practice.  Or to put it another way UTF-8 is never more than about 1.9x any legacy encoding (and usually less) -- not "more than 2x"

That's because languages like Russian still share basic punctuation marks with ASCII - so spaces, newlines,  and other basic punctuation (.,;:!) are still single bytes - which waters down the ratio significantly.  And CJK / SE Asian languages where 3-byte characters are predominant in UTF-8 never had pure single-byte legacy encodings in the first place, so the alternative encodings are all multibyte.  Which also means that regardless of the ratio, quick-xml can't parse them without decoding first anyway.

The entirety of "War and Peace" by Leo Tolstoy in the original Russian is ~7mb in UTF-8, and about 3.7mb in a single-byte encoding like Windows-1251 or KOI8-R. So that's a practical example both of the <2 compression ratio and also that you'd need to be dealing with enormous quantities of text at once in order for this to be much of a problem. 

And large XML documents are usually not large because of text, they are large because they contain large quantities of structured data, which is not going to be helped by different encodings.  The solution to that is parsing, interning, filtering and other various techniques rather than encoding tricks.  I'm personally very familiar with this, because I use this library for dealing with RPM repository metadata, which is regularly multi-gigabyte files, for which string interning is the most effective solution.

Even if a user does, for some reason, need to collect such large quantities of text while streaming the document, LZ4 or Zstd compression is likely to give you an equivalent or better ratio, and would likely be just as easy or possibly easier for a user to deal with than a legacy encoding from Rust.  It also keep the pain contained to whichever fractions of the document they're working with that are actually large, which is likely specific elements rather than the whole thing.

src/de/mod.rs
```diff
@@ -2776,16 +2768,16 @@ where
         self.skip_event(event)?;
         // Skip all subtree, if we skip a start event
         if let Some(DeEvent::Start(e)) = self.write.back() {
-            let end = e.name().as_ref().to_owned();
+            let end = e.name().as_ref().as_bytes().to_owned();
```
Mingun, line 2792: Why convet to `Vec<u8>`? May be converted to `String`

src/de/mod.rs
```diff
@@ -2909,19 +2903,23 @@ where
...
             // We can get End event in case of `<tag></tag>` or `<tag/>` input
             // Return empty text in that case
             // The matching tag name is guaranteed by the reader
             // Reached by {...}::xs_list::empty
             DeEvent::End(_) => Ok("".into()),
             // Reached by trivial::{...}::{field_nested, field_tag_before, nested, tag_before}
-            DeEvent::Start(s) => Err(DeError::UnexpectedStart(s.name().as_ref().to_owned())),
+            DeEvent::Start(s) => Err(DeError::UnexpectedStart(
+                s.name().as_ref().as_bytes().to_owned(),
+            )),
```
Mingun: `UnexpectedStart` now may store `String`

src/de/resolver.rs
```diff
@@ -39,7 +39,7 @@ use crate::events::BytesText;
 ///     type Error = FromUtf8Error;
 ///
 ///     fn capture(&mut self, doctype: BytesText) -> Result<(), Self::Error> {
-///         for cap in self.re.captures_iter(&doctype) {
+///         for cap in self.re.captures_iter(doctype.as_ref().as_bytes()) {
```
Mingun: We may use strings now

src/events/attributes.rs
```diff
@@ -877,54 +782,50 @@ impl<T> Attr<T> {
...
+            Attr::DoubleQ(_, value) => *value,
+            Attr::SingleQ(_, value) => *value,
+            Attr::Empty(_) => "",
+            Attr::Unquoted(_, value) => *value,
         }
     }
 }
 
-impl<T: AsRef<[u8]>> Debug for Attr<T> {
+impl<T: Debug> Debug for Attr<T> {
```
Mingun, line 804: Now `Debug` can be derived. Manual implementation is used to use `Bytes` wrapper

src/writer.rs
```diff
@@ -262,7 +262,7 @@ impl<W: Write> Writer<W> {
         let mut next_should_line_break = true;
         let result = match event.into() {
             Event::Start(e) => {
-                let result = self.write_wrapped(b"<", &e, b">");
+                let result = self.write_wrapped(b"<", e.as_bytes(), b">");
```
Mingun: Shouldn't those functions be changed to accept `&str`?

Changelog.md
```diff
@@ -20,6 +20,48 @@
 
 ### Misc Changes
 
+- [#963]: MSRV bumped to 1.86 (April 2025)
```
Mingun: I think, this should be mentioned right under "Unreleased" section, because it is important change.

Changelog.md
```diff
@@ -20,6 +20,48 @@
...
+- [#963]: Deprecated `Attribute` methods that take a `Decoder` parameter, since
+  attribute values are now always valid UTF-8: `decoded_and_normalized_value()`,
+  `decoded_and_normalized_value_with()`, `decode_and_unescape_value()`, and
+  `decode_and_unescape_value_with()`. Use `normalized_value()` and
+  `normalized_value_with()` instead.
+
+### Breaking Changes
```
Mingun, line 24: And this

src/name.rs
```diff
@@ -251,36 +247,36 @@ impl<'a> From<QName<'a>> for LocalName<'a> {
...
+        self.0 == "xml"
     }
 
     /// Checks if this prefix is a special prefix `xmlns`.
+    // TODO: restore const once PartialEq for str is usable in const context (const_trait_impl)
     #[inline(always)]
-    pub const fn is_xmlns(&self) -> bool {
-        matches!(self.0, b"xmlns")
+    pub fn is_xmlns(&self) -> bool {
+        self.0 == "xmlns"
```
Mingun: You may use [`.as_bytes()`](https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes) (const since 1.39) to keep constness

src/name.rs
```diff
@@ -450,35 +438,30 @@ struct NamespaceBinding {
...
         } else {
-            // We use split_at to get [start + prefix_len..start + prefix_len + value_len]
-            // in a constant way
-            let (_, ns) = buffer.split_at(self.start + self.prefix_len);
-            let (ns, _) = ns.split_at(self.value_len);
-            ResolveResult::Bound(Namespace(ns))
+            let start = self.start + self.prefix_len;
+            ResolveResult::Bound(Namespace(&buffer[start..start + self.value_len]))
         }
     }
```
Mingun: [`str::split_at`](https://doc.rust-lang.org/std/primitive.str.html#method.split_at) const since 1.86, so you only need to change type here

src/name.rs
```diff
@@ -502,7 +485,7 @@ pub struct NamespaceResolver {
 impl Debug for NamespaceResolver {
```
Mingun: Now `Debug` can be derived

src/writer/async_tokio.rs
```diff
@@ -13,7 +13,7 @@ impl<W: AsyncWrite + Unpin> Writer<W> {
         let mut next_should_line_break = true;
         let result = match event.into() {
             Event::Start(e) => {
-                let result = self.write_wrapped_async(b"<", &e, b">").await;
+                let result = self.write_wrapped_async(b"<", e.as_bytes(), b">").await;
```
Mingun: Shouldn't those functions be changed to accept `&str`?

src/reader/state.rs
```diff
@@ -290,62 +296,62 @@ impl ReaderState {
...
+            Err(Error::Syntax(PiParser(false).eof_error(buf_bytes)))
         }
     }
 
     /// Converts content of a tag to a `Start` or an `Empty` event
     ///
     /// # Parameters
     /// - `content`: Content of a tag between `<` and `>`
-    pub fn emit_start<'b>(&mut self, content: &'b [u8]) -> Event<'b> {
+    pub fn emit_start<'b>(&mut self, buf: &'b str) -> Result<Event<'b>> {
```
Mingun, line 314: `Result` was required as intermediate step, but in the final it can gone, right?

src/reader/state.rs
Mingun: Now can be derived

dralley:
Plus, the constructors for `BytesText`, `BytesStart` and so forth are already based around `Into<Cow<'a, &str>>` and have been for a while.  You could not easily write XML directly from code using any encoding other than `utf-8` as it stands.

dralley:
@Mingun Can you give me some idea of when you plan to review this, I will be away from my laptop for a week and a half starting this weekend.

src/events/mod.rs
```diff
@@ -285,19 +260,19 @@ impl<'a> BytesStart<'a> {
...
+        Attributes::wrap(&self.buf, self.name_len, true)
     }
 
     /// Gets the undecoded raw string with the attributes of this tag as a `&[u8]`,
     /// including the whitespace after the tag name if there is any.
     #[inline]
     pub fn attributes_raw(&self) -> &[u8] {
-        &self.buf[self.name_len..]
+        self.buf[self.name_len..].as_bytes()
     }
```
Mingun: It seems you forgot to change type here

src/events/mod.rs
Mingun: Do we want to use `AsRef<str>` here? Actually, I would remove generic or put implementation to non-generic method to avoid code bloat due to  monomorphization.

src/events/mod.rs
```diff
@@ -1321,11 +1247,11 @@ impl<'a> BytesDecl<'a> {
     ///
     /// // <?xml version='1.1'?>
     /// let decl = BytesDecl::from_start(BytesStart::from_content(" version='1.1'", 0));
-    /// assert_eq!(decl.version().unwrap(), b"1.1".as_ref());
+    /// assert_eq!(decl.version().unwrap().as_ref(), "1.1");
```
Mingun: It seems, `.as_ref()` here is not needed?

src/events/mod.rs
```diff
@@ -1321,11 +1247,11 @@ impl<'a> BytesDecl<'a> {
...
     ///
     /// // <?xml version='1.0' version='1.1'?>
     /// let decl = BytesDecl::from_start(BytesStart::from_content(" version='1.0' version='1.1'", 0));
-    /// assert_eq!(decl.version().unwrap(), b"1.0".as_ref());
+    /// assert_eq!(decl.version().unwrap().as_ref(), "1.0");
```
Mingun: And here

src/reader/mod.rs
```diff
@@ -1007,35 +1008,35 @@ enum ReadTextResult<'r, B> {
...
     /// `&` was not consumed.
-    UpToRef(&'r [u8]),
+    UpToRef(&'r str),
     /// Contains text block up to EOF, neither start of markup (`<` character)
     /// or start of reference (`&` character) was found.
-    UpToEof(&'r [u8]),
+    UpToEof(&'r str),
     /// IO error occurred.
-    Err(io::Error),
+    Err(Error),
```
Mingun, line 1019: Now that's contain not only IO error, right? Then need to update docstring

src/reader/mod.rs
```diff
@@ -1007,35 +1008,35 @@ enum ReadTextResult<'r, B> {
...
     /// Result includes start `&`.
-    UpToRef(&'r [u8]),
+    UpToRef(&'r str),
     /// Contains text block up to start of markup (`<` character).
     /// Result includes start `&`.
-    UpToMarkup(&'r [u8]),
+    UpToMarkup(&'r str),
     /// IO error occurred.
-    Err(io::Error),
+    Err(Error),
```
Mingun, line 1039: And here

src/de/simple_type.rs
```diff
@@ -1220,85 +1190,4 @@ mod tests {
         simple_only!(utf8, identifier: Id = "Field" => Id::Field);
         simple_only!(utf8, ignored_any: Any = "any data" => Any(IgnoredAny));
     }
-
-    #[cfg(feature = "encoding")]
-    mod utf16 {
```
Mingun, line 1225: Is it not possible to keep those tests?

Mingun (commented):
I started review on per-commit basis about month ago, but finished it now using the final diff, to we may move forward. So maybe some last questions may be dumb.

I still think, that changing the low level from `[u8]` to `str` is not a better idea. It would be better at middle-level, where we will also expand general entity references (#948).

Mingun:
Also just want to inform you, that I plan to release 0.41.0 now with the latest security fixes.

src/de/simple_type.rs
dralley: It's probably possible but I don't understand what value it would provide

The actual mechanism of this library's support for UTF-16 involves processing it before it ever reaches the parser / Deserializer code in the first place. There's not really a use case where a user could Deserialize XML markup directly from UTF-16 without being pre-decoded - and to the extent that any of  these tests work, it's only because they don't involve any  actual XML markup?  All the nontrivial cases here already expect failure.

There's a UTF-8 copy of all of these tests cases a few lines up, which covers everything a user could hit in practice.  IMO it would make sense to drop or rename the module though.

src/writer.rs
dralley: For the inner functions which are writing to output, it doesn't make much difference.  Could be either way.

src/reader/state.rs
dralley: Per the current state of the branch, yes it can be removed.

re:

> I still think, that changing the low level from [u8] to str is not a better idea. It would be better at middle-level, where we will also expand general entity references (https://github.com/tafia/quick-xml/pull/948).

Where did you have in mind for "mid-level"?  Because if it's here, then the `Result` would stay.

src/writer.rs
dralley: But I will change it anyways.  It does clean things up slightly.

dralley:
@Mingun Updated

I will wait to rebase / resolve the merge conflicts until you're done reviewing.

src/reader/state.rs
Mingun: Low level -- current `Reader`
Middle level -- #948 (performs automatic expand of general entity references)
High level -- serde interface

dralley:
> Low level -- current Reader
> Middle level -- https://github.com/tafia/quick-xml/pull/948 (performs automatic expand of general entity references)
> High level -- serde interface

I'm not sure I agree, but I'm open to hearing the argument in favor.  I can think of some positives and some negatives.

Is this a blocking issue or something that can be worked on later?

dralley:
I rebased anyway, but all the changes are in new commits, everything after and including [22c4a94](https://github.com/tafia/quick-xml/pull/963/commits/22c4a94eabade9bd3a633fdb03ab88e61f43af9e) "Make DeError::UnexpectedStart carry String"

dralley:
@Mingun Is this still considered blocked? And on what, specifically?

dralley:
@Mingun ^^?

I understand your broader concern around the composition of features, but that need not block this PR specifically as it doesn't change anything in that respect.  And anyway, there are potential solutions which we can discuss on https://github.com/tafia/quick-xml/issues/158

dralley:
I merge this in 3 days without further comment

Mingun:
Unfortunately, I have not yet been able to find time to think about alternative solutions. I don't know when it will be. So that your enthusiasm does not fade, I will not obstruct merging this PR, but only a request - if #982 is completed in the near future, merge it first and make release with it, and then merge this PR. For other open PRs, I have questions that I will ask in the near future, so I do not mention them here.

dralley:
Would you object to adjusting the release process such that we create official branches for each new release that can accept backports directly, rather than encumbering work on `master` branch?

The process at my day job uses GHA to automate the process like so:

* Tagging a PR for backport, or merging a tagged PR, automatically creates a backport PR against the latest release branch

(two CI jobs, "create new release branch" and "publish new release"

* "Create new release branch"
    * Forks `master` to a new branch named w/ the current version number
    * Automatically creates a PR against the `master` branch to bump the release number
* "Publish new release (accepts branch target as argument)"
    * Publishes a patch release from the target release branch
    * Updates the changelog locally
    * Automatically creates a PR against master branch w/ the changelog updates for the new minor release

The release process for new minor releases then consists of running the two jobs in sequence, and for new patch releases consists of running only the second job (as no new branch is needed)

Mingun:
In fact, we do this (at least, I 😄 ), it's just that this process is not automated. It was required only 2 times (for [0.23](https://github.com/tafia/quick-xml/tree/v0.23) and [0.39](https://github.com/tafia/quick-xml/tree/v0.39)), and I didn't want to tinker with automation for this. In addition, while we are in the 0.x release cycle, there are unlikely to be many patch changes for this automation to be required. Almost everything important requires a change of minor version, so it just naturally continues to grow from master.

dralley:
Well, I will see if #982 can be resolved this weekend, if not I will fork off a branch for it to land on later.

dralley:
Fork a branch it is.

PR #964, opened 2026-05-13
Fix unreachable!() panic when DOCTYPE appears between text runs in element content

williamareynolds:
## Summary

  The serde deserializer panics with `internal error: entered unreachable code` (`src/de/mod.rs` `read_text`'s "Cannot be two consequent Text events" branch) when a DOCTYPE declaration appears between text runs inside element content.
  Minimal repro:

  ```rust
  use serde::Deserialize;

  #[derive(Deserialize)]
  struct Wrapper { #[serde(rename = "$text")] text: String }

  // Panics on master:
  //   thread '...' panicked at quick-xml/src/de/mod.rs:2910:21:
  //   internal error: entered unreachable code
  let _: Wrapper = quick_xml::de::from_str(r#"<a>x<!DOCTYPE y>z</a>"#).unwrap();
  ```

I found this while fuzzing a new library I'm working on. I'm not an expert on quick-xml so I hope I got this all right.

  ## Root cause

  `Deserializer::drain_text` is the merge step that guarantees `read_text` (and friends) see at most one `DeEvent::Text` per element. It loops while `current_event_is_last_text` returns `false`, draining `Text` / `CData` / `GeneralRef`
  payload events into a single result.

  `DocType` is **not** in `current_event_is_last_text`'s match arm, so the loop exits when a DocType appears. The outer `Deserializer::next()` then:

  1. Returns the accumulated `DeEvent::Text` for the *first* text run.
  2. On the next call, processes the `DocType` (captures it via the entity resolver) and `continue`s.
  3. Returns the *second* text run as a separate `DeEvent::Text`.

  `read_text`, expecting a single text run followed by `End`, hits the consecutive-`Text` `unreachable!()`.

  ## Fix

  Treat `DocType` as transparent during the text drain:

  - `current_event_is_last_text` now also returns `false` when the lookahead is `DocType`, so `drain_text` keeps draining instead of breaking at a DOCTYPE boundary.
  - `drain_text`'s match arm gains a `PayloadEvent::DocType(e)` case that forwards the event to `entity_resolver.capture()` (identical to the existing `Deserializer::next()` path) and continues draining.

  DTD entities defined inside the document body remain available for subsequent `&entity;` resolution — the entity resolver sees DOCTYPE events in exactly the same order as before. The only observable change is that the surrounding text
  runs merge into one `DeEvent::Text`, which matches what `read_text` and the rest of the serde deserializer have always assumed.

  ## Tests

  Adds a new `doctype_in_element_text` module in `tests/serde-de.rs` with three cases:

  | Case | Input |
  | --- | --- |
  | `single_doctype_between_text` | `<a>x<!DOCTYPE y>z</a>` |
  | `multiple_doctypes_between_text` | `<a>x<!DOCTYPE y><!DOCTYPE z>w</a>` (matches the fuzzer-discovered shape) |
  | `leading_doctype_then_text` | `<a><!DOCTYPE y>x</a>` |

  All three previously panicked, now pass. Full `cargo test --all-features` stays green (**1,711 passing**, no regressions).

  ## MSRV / fmt / minimal-versions

  - `cargo fmt --check`: clean.
  - `cargo check` on `rust-toolchain 1.79.0`: clean (no new language or stdlib features used).
  - No new dependencies; the fix is local to `src/de/mod.rs` plus the new test cases.

### DIFF
--- a/tests/serde-de.rs
+++ b/tests/serde-de.rs
@@ -1896,6 +1896,84 @@ mod resolve {
     }
 }
 
+/// A DOCTYPE declaration inside an element's text content used to split
+/// the surrounding character data into two consecutive `DeEvent::Text`
+/// events, tripping an `unreachable!()` in `read_text` ("Cannot be two
+/// consequent Text events"). The reader emits Text/DocType/Text for an
+/// input like `<a>x<!DOCTYPE y>z</a>`, and `drain_text` did not drain
+/// across DocType events, so the two text runs reached the
+/// deserializer un-merged. Discovered via libFuzzer on a real-world
+/// SAML response harness.
+///
+/// The expected behavior is: DOCTYPE inside an element is captured by
+/// the entity resolver (it has the same role as a prolog-level DOCTYPE
+/// for entity definitions) and the surrounding text is treated as one
+/// continuous run.
+mod doctype_in_element_text {
+    use super::*;
+    use pretty_assertions::assert_eq;
+
+    #[derive(Debug, Deserialize, PartialEq)]
+    struct Wrapper {
+        #[serde(rename = "$text")]
+        text: String,
+    }
+
+    /// Minimal regression repro: a single DOCTYPE event between two
+    /// adjacent text runs inside an element should not panic.
+    #[test]
+    fn single_doctype_between_text() {
+        let xml = r#"<a>x<!DOCTYPE y>z</a>"#;
+        let got: Wrapper = from_str(xml).unwrap();
+        assert_eq!(got, Wrapper { text: "xz".into() });
+    }
+
+    /// Multiple DOCTYPE events stacked between text runs (matching the
+    /// shape of the fuzzer-discovered input that originally surfaced
+    /// this panic on real SAML traffic).
+    #[test]
+    fn multiple_doctypes_between_text() {
+        let xml = r#"<a>x<!DOCTYPE y><!DOCTYPE z>w</a>"#;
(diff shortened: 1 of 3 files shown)


tests/serde-de.rs
```diff
@@ -1896,6 +1896,58 @@ mod resolve {
...
+    struct Wrapper {
+        #[serde(rename = "$text")]
+        text: String,
+    }
+
+    /// Minimal regression repro: a single DOCTYPE event between two
+    /// adjacent text runs inside an element should not panic.
+    #[test]
+    fn single_doctype_between_text() {
+        let xml = r#"<a>x<!DOCTYPE y>z</a>"#;
```
Mingun, line 1926: I recommend also check how the trailing spaces will be handled and is that matches expectations
```suggestion
        let xml = "<a>x <!DOCTYPE y>\tz</a>";
```

Mingun (changes requested):
Generally approve except needing to check the trailing whitespace handling question.

tests/serde-de.rs
williamareynolds: Will do. I'll have an update in shortly.

williamareynolds: I've added the desired assertions.

https://github.com/tafia/quick-xml/pull/964/changes#diff-78461d026840e91b6b02137004e4a9c6b3dd22b710843079743c9d471b1d3c6bR1950-R1964

Issue #970, opened 2026-06-22
`NamespaceResolver::push` — unbounded per-`xmlns` heap allocation inside `NsReader`, before the event is returned → OOM on untrusted XML


qifan-sailboat:
## Summary

`NsReader::read_resolved_event_into()` buffers a start tag, then calls
`process_event()` (`src/reader/ns_reader.rs:80-90`, v0.40.1; `:166-170` in
v0.39.4), which calls `NamespaceResolver::push(&e)` **before** returning the
event to the caller. `push()` (`src/name.rs:674-688`, v0.40.1) iterates every
attribute on the start tag and, for each `xmlns`/`xmlns:*` binding, calls
`add()` (`name.rs:618-668`, v0.40.1), which
`extend_from_slice`s the prefix into `self.buffer: Vec<u8>` and `push`es a
32-byte `NamespaceBinding` into `self.bindings: Vec<NamespaceBinding>`. Neither
vector is capped. A start tag with N `xmlns:*` attributes therefore drives
~`(prefix_len + 32) × N` bytes of heap allocation (plus `Vec` amortised-growth
slack, ~1.5×) inside quick-xml, *before* the `NsReader` consumer ever sees the
event and has any opportunity to reject it.

A consumer that bounds *input bytes* (the start tag fits in M bytes) still
incurs ~3.2 × M bytes of `NamespaceResolver` heap that it cannot observe or
limit. With several concurrent `NsReader`s on independent inputs the
allocations stack to process-fatal levels.

## quick-xml version and build

- **quick-xml:** 0.39.4 (as bundled by rpki-rs 0.19.3) and **0.40.1 (latest
  release, re-confirmed 2026-06-22)** — `NamespaceResolver::push` is
  byte-for-byte identical at `src/name.rs:674-688`; `process_event` is
  identical at `src/reader/ns_reader.rs:80-90`.
- **Rust:** 1.86.0 stable, `--release`
- **Platform:** Linux x86_64 (Debian 12, glibc 2.36)

## Preconditions and configuration

- **Attack surface:** any code path that uses `NsReader` (or otherwise drives
  `NamespaceResolver::push`) on XML from an untrusted source.
- **Object type carrying the trigger:** a single XML start tag with N
  `xmlns:prefix=""` attributes.
- **Config gate:** none — `NsReader` always calls `push()` before yielding
  `Event::Start`/`Event::Empty`. There is no public knob to cap
  `bindings.len()` or to disable namespace resolution while still using
[...]

PR #971, opened 2026-06-22
Fix O(N²) duplicate-attribute check in Attributes iterator

### LINKED ISSUE #969: `IterState::check_for_duplicates` — O(N²) attribute-key linear scan → CPU DoS on untrusted XML
## Summary

`BytesStart::attributes()` returns an `Attributes` iterator whose default
`IterState` has `check_duplicates = true`. For each attribute yielded,
`IterState::check_for_duplicates` (`src/events/attributes.rs:1067-1081`,
v0.40.1; `:817-832` in v0.39.4)
performs an O(N) linear scan of `self.keys: Vec<Range<usize>>` and then
`push`es the new key, so iterating N attributes on a single start tag costs
O(N²/2) byte comparisons. There is no upper bound on N other than the size of
the start-tag buffer. Any application that calls `.attributes()` (or uses
`NsReader`, which calls it internally) on attacker-supplied XML without
explicitly chaining `.with_checks(false)` is exposed to a CPU-exhaustion DoS.

Because the loop is pure computation with no `.await`/I/O, downstream
per-request I/O timeouts (e.g. `reqwest::RequestBuilder::timeout`, tokio
`tokio::time::timeout` around a read) cannot fire while it runs.

## quick-xml version and build

- **quick-xml:** 0.39.4 (as bundled by rpki-rs 0.19.3) and **0.40.1 (latest
  release, re-confirmed 2026-06-22)** — `check_for_duplicates` is byte-for-byte
  identical, now at `src/events/attributes.rs:1067-1081`.
- **Rust:** 1.86.0 stable, `--release`
- **Platform:** Linux x86_64 (Debian 12, glibc 2.36)

## Preconditions and configuration

- **Attack surface:** any code path that iterates `BytesStart::attributes()`
  (or `BytesStart::try_get_attribute`, or any `NsReader`/`Reader` consumer that
  reaches `Attributes::next`) on XML from an untrusted source.
- **Object type carrying the trigger:** a single XML start tag with N distinct
  attribute names. `xmlns:aXXXXXX=""` is convenient because most consumers
  silently skip namespace bindings, so the attack does not change application
  semantics.
- **Config gate:** none — `check_duplicates = true` is the library default
  (`IterState::new`, `attributes.rs`). The consumer must *opt out* via
  `.with_checks(false)` to avoid the O(N²) path.
- **Default config affected?** yes.

## Attacker's position and abilities

Any party able to supply XML to a quick-xml consumer. Confirmed downstream
impact: **NLnet Labs Routinator 0.15.2** (RPKI relying-party validator,
`rpki-rs 0.19.3`) — any of ~18,000+ delegated RPKI CAs worldwide can deliver a
crafted RRDP `snapshot.xml` over HTTPS and stall the validator's CPU past its
own 600 s `rrdp-timeout`, which is I/O-gated and therefore cannot interrupt the
dup-check (see "Demonstrated downstream impact" below). The same primitive
applies to any SOAP/RSS/SVG/config parser built on quick-xml that does not call
`.with_checks(false)`.

## Impact

| N (attributes) | Start-tag bytes | gzip | Release dup-check (~530 M cmp/s) |
|---|---|---|---|
| 50,000 | ~850 KB | ~35 KB | ~2.4 s |
| 800,000 | ~13.4 MB | ~400 KB | ~604 s |
| 5,900,000 | ~100 MB | ~2.9 MB | ~33,190 s |

No crash, no memory exhaustion (the `keys` Vec is O(N) and small); pure CPU
stall on a single thread. The thread holds whatever locks/resources the caller
held when it entered the iterator.

### CVSS v3.1

`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:L` — **5.3 Medium** (CWE-407
Algorithmic Complexity, CWE-400 Uncontrolled Resource Consumption). Downstream
consumers without an effective wall-clock bound on XML parsing may rate this
A:H / 7.5.

## Steps to reproduce

Minimal standalone reproducer (no external dependencies beyond quick-xml):

```rust
// cargo new qx-f016 && cd qx-f016
// cargo add quick-xml@0.40.1
use std::time::Instant;
use quick_xml::events::Event;

fn main() {
    let n: usize = std::env::args().nth(1).and_then(|s| s.parse().ok()).unwrap_or(50_000);
    let mut tag = String::from(r#"<e uri="x" "#);
    for i in 0..n { tag.push_str(&format!(r#"xmlns:a{:06}="" "#, i)); }
    tag.push_str("/>");

    let mut r = Reader::from_str(&tag);
    let t0 = Instant::now();
    if let Ok(Event::Empty(bs)) = r.read_event() {
        let mut c = 0usize;
        for a in bs.attributes() { a.unwrap(); c += 1; }   // default: check_duplicates = true
        eprintln!("N={} attrs={} elapsed={:?}", n, c, t0.elapsed());
    }
}
```

Measured against **quick-xml 0.40.1**, `cargo build --release`:

```
N=10000  attrs=10001  elapsed_ms=75
N=20000  attrs=20001  elapsed_ms=303
N=40000  attrs=40001  elapsed_ms=1251
N=80000  attrs=80001  elapsed_ms=6109
```

Elapsed time scales as N² (each 2× N → ~4× time: 75→303→1251→6109 ms),
confirming the quadratic. Extrapolated N=800,000 → ~610 s.

A full network-level Docker PoC against Routinator 0.15.2 is available on
request (Go RRDP server + ASAN-instrumented Routinator; `run.sh` reproduces
end-to-end).

### Trigger object

```xml
<e uri="x" xmlns:a000000="" xmlns:a000001="" … xmlns:a049999="" />
```

The attribute *names* must be distinct (else `AttrError::Duplicated` short-
circuits); values are irrelevant.

## Current (bug) behaviour

`Attributes::next()` → `IterState::next()` → `check_for_duplicates()` performs
`self.keys.iter().find(|r| slice[(*r).clone()] == slice[key.clone()])` then
`self.keys.push(key.clone())` for every attribute, including `xmlns:*`.

## Expected (correct) behaviour

Duplicate-attribute detection should be at most O(N log N) (sorted insert) or
amortised O(N) (hash set) over the whole start tag, and/or namespace-binding
attributes should be exempt from the check (XML Namespaces 1.1 already forbids
two bindings of the same prefix on one element, but the *cost* of detecting
that should not be quadratic).

## Root cause

`quick-xml-0.40.1/src/events/attributes.rs:1067-1081`
(`quick-xml-0.39.4/src/events/attributes.rs:817-832` — identical):

```rust
fn check_for_duplicates(
    &mut self,
    slice: &[u8],
    key: Range<usize>,
) -> Result<Range<usize>, AttrError> {
    if self.check_duplicates {
        if let Some(prev) = self.keys.iter()
            .find(|r| slice[(*r).clone()] == slice[key.clone()])
        {
            return Err(AttrError::Duplicated(key.start, prev.start));
        }
        self.keys.push(key.clone());
    }
    Ok(key)
}
```

Called unconditionally from `IterState::next()` (`attributes.rs:1093` and
`:1178` in v0.40.1; `:928` in v0.39.4). No attribute category is exempt; no
size bound on `self.keys`.

## Suggested fix

Any of (in rough order of preference):

1. **Hash the seen keys** — replace `Vec<Range<usize>>` + linear `find` with a
   `HashSet<u64>` of FxHash/aHash of `&slice[key.clone()]` (collision → fall
   back to byte compare). O(N) total, preserves the existing
   `AttrError::Duplicated` semantics.
2. **Exempt namespace bindings** from the dup-check:
   ```rust
   if self.check_duplicates {
       if slice[key.clone()].starts_with(b"xmlns")
           && (slice.get(key.start + 5) == Some(&b':') || key.len() == 5)
       {
           return Ok(key);
       }
       ...
   }
   ```
   This closes the cheapest amplification vector (consumers usually ignore
   `xmlns:*`) while keeping the check for "real" attributes.
3. **Cap `self.keys.len()`** at a configurable limit (e.g. 256) and return
   `AttrError::TooManyAttributes` beyond it.
4. At minimum, **document** in `Attributes`/`with_checks` rustdoc that the
   default dup-check is O(N²) and that callers handling untrusted input should
   chain `.with_checks(false)`.

## Demonstrated downstream impact (Routinator 0.15.2)

`rpki-rs 0.19.3` `src/xml/decode.rs:124` iterates
`self.start.attributes()` without `.with_checks(false)`; `xmlns:*` is filtered
at `decode.rs:126` only after the dup-check has run. Routinator's
`rrdp-timeout` (default 600 s) is set via `reqwest::RequestBuilder::timeout`,
which is checked only at async-I/O await points and cannot fire inside the
dup-check loop. Empirical (ASAN build, N=50,000, `rrdp-timeout=5s`):
wall_clock=11.05 s ≫ 5 s; release projection N=800,000 → ~614 s > 600 s. A
parallel report has been sent to NLnet Labs (`security@nlnetlabs.nl`)
recommending the consumer-side `.with_checks(false)` mitigation in rpki-rs.

## Coordination

- **Shared with anyone else:** NLnet Labs (`security@nlnetlabs.nl`), as
  Routinator/rpki-rs maintainers and the demonstrated downstream consumer.
- **Publication plan:** no plan to publish independently; will follow the
  quick-xml maintainers' timeline. If a fix lands, a RustSec advisory
  (https://github.com/rustsec/advisory-db) is suggested so other downstreams
  are notified via `cargo audit`.

## Acknowledgements

Credits: Qifan Zhang (Palo Alto Networks).

qifan-sailboat:
Fixes #969.

`IterState::check_for_duplicates` did a linear `Vec` scan of all already-seen attribute name ranges for every new attribute, so a start tag with N distinct attribute names cost O(N²/2) byte comparisons. On untrusted XML this is a CPU-exhaustion DoS — see #969 for measurements (N=80,000 ≈ 6.1 s release; N=800,000 ≈ 10 min) and the demonstrated downstream impact on NLnet Labs Routinator.

This adds a `HashSet<u64>` of `DefaultHasher` hashes of the key bytes as an O(1) pre-filter:

* a fresh hash means the key cannot be a duplicate → push and return `Ok` (the no-duplicate path is now amortised O(1) per attribute / O(N) per start tag);
* on a hash hit (a real duplicate, or an astronomically rare 64-bit collision) fall back to the existing linear scan to recover the exact previous position for `AttrError::Duplicated(new, prev)` — error semantics are unchanged.

The set is lazily allocated (`Option<HashSet<u64>>`) so `IterState::new`, `Attributes::new` and `Attributes::html` stay `const fn`.

| N (distinct attrs) | before | after |
|---|---|---|
| 10,000 | 75 ms | 0 ms |
| 20,000 | 303 ms | 1 ms |
| 40,000 | 1,251 ms | 2 ms |
| 80,000 | 6,109 ms | 4 ms |

A timing regression test is included in `events::attributes::xml::duplicated::with_check`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,6 +21,11 @@
 
 ### Bug Fixes
 
+- [#969]: `Attributes` (and anything that iterates `BytesStart::attributes()`
+  with the default `with_checks(true)`) no longer takes O(N²) time on a start
+  tag with a large number of attributes. Small tags keep the previous linear
+  scan; larger ones switch to a 64-bit hash pre-filter, so the whole tag is
+  O(N). The exact `AttrError::Duplicated(new, prev)` positions are unchanged.
 - [#970]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
   event) now rejects a start tag that declares more than
   `DEFAULT_MAX_DECLARATIONS_PER_ELEMENT` (256) `xmlns` / `xmlns:*` namespace
@@ -31,9 +36,10 @@
   via `NamespaceResolver::set_max_declarations_per_element` (use `usize::MAX`
   to disable).
 
-[#970]: https://github.com/tafia/quick-xml/issues/970
-
 ### Misc Changes
+
+[#969]: https://github.com/tafia/quick-xml/issues/969
+[#970]: https://github.com/tafia/quick-xml/issues/970
 
 
 ## 0.40.1 -- 2026-05-15
--- a/src/events/attributes.rs
+++ b/src/events/attributes.rs
@@ -9,7 +9,9 @@ use crate::name::{LocalName, Namespace, NamespaceResolver, QName};
 use crate::utils::{is_whitespace, Bytes};
 use crate::XmlVersion;
 
+use std::collections::HashSet;
 use std::fmt::{self, Debug, Display, Formatter};
+use std::hash::{BuildHasherDefault, DefaultHasher, Hasher};
 use std::iter::FusedIterator;
 use std::{borrow::Cow, ops::Range};
 
@@ -964,6 +966,53 @@ enum State {
     SkipEqValue(usize),
 }
 
+/// Number of attributes a start tag may have before the duplicate-name check
+/// switches from a direct linear scan of the previously seen names to a hash
+/// pre-filter (see [`IterState::check_for_duplicates`]).
+///
+/// Real-world start tags carry only a handful of attributes -- the busiest
+/// element in our benchmark corpus (`tests/documents/players.xml`) has 22 --
+/// where the scan is faster than hashing and needs no allocation. Larger tags
+/// are where the scan became the O(N²) CPU-DoS of [#969], so above this count we
+/// pay for a hash set to keep the whole tag O(N). The value sits just above the
+/// measured linear-vs-hash crossover.
+///
+/// [#969]: https://github.com/tafia/quick-xml/issues/969
+const SMALL_ATTRIBUTE_COUNT: usize = 32;
+
+/// A no-op [`Hasher`] for the `key_hashes` set, whose values are already 64-bit
+/// hashes of attribute names; re-hashing them with the default SipHash would be
+/// wasted work. Only `write_u64` is ever exercised (via `u64`'s `Hash` impl).
+#[derive(Default)]
+struct IdentityHasher(u64);
+
+impl Hasher for IdentityHasher {
+    #[inline]
+    fn finish(&self) -> u64 {
+        self.0
+    }
+
+    #[inline]
+    fn write(&mut self, _: &[u8]) {
+        // The set only ever stores `u64` keys, which route through `write_u64`.
+        unreachable!("IdentityHasher only supports u64 keys")
+    }
+
+    #[inline]
+    fn write_u64(&mut self, n: u64) {
+        self.0 = n;
+    }
+}
+
+/// Hashes a single attribute name. A fresh [`DefaultHasher`] per name keeps each
+/// hash independent (so it is also DoS-resistant on untrusted input).
+#[inline]
+fn hash_name(name: &[u8]) -> u64 {
+    let mut hasher = DefaultHasher::new();
+    hasher.write(name);
+    hasher.finish()
+}
+
 /// External iterator over spans of attribute key and value
 #[derive(Clone, Debug)]
 pub(crate) struct IterState {
@@ -979,6 +1028,13 @@ pub(crate) struct IterState {
     /// names. We store a ranges instead of slices to able to report a previous
     /// attribute position
     keys: Vec<Range<usize>>,
+    /// 64-bit hashes of the byte content of `keys`, used as an O(1) pre-filter
+    /// once a start tag declares more than `SMALL_ATTRIBUTE_COUNT` attributes, so
+    /// the duplicate check stays O(N) over the whole tag instead of O(N²). The
+    /// values are already hashes, so the set stores them with `IdentityHasher`
+    /// instead of re-hashing. Allocated only when the threshold is crossed, so
+    /// small tags (and [`IterState::new`]) stay allocation-free and `const`.
+    key_hashes: Option<HashSet<u64, BuildHasherDefault<IdentityHasher>>>,
 }
 
 impl IterState {
@@ -988,6 +1044,7 @@ impl IterState {
             html,
             check_duplicates: true,
             keys: Vec::new(),
+            key_hashes: None,
         }
     }
 
@@ -1063,6 +1120,17 @@ impl IterState {
         }
     }
 
+    /// Checks that the attribute name `key` (a range into `slice`) was not seen
+    /// earlier in the same start tag, recording it for subsequent checks.
+    ///
+    /// Small tags use a direct linear scan of [`Self::keys`]: for a handful of
+    /// attributes that beats hashing and needs no allocation, which is the
+    /// overwhelmingly common case. Once a tag declares more than
+    /// `SMALL_ATTRIBUTE_COUNT` attributes -- where the scan would become the
+    /// O(N²) CPU-DoS of [#969] -- it switches to a hash pre-filter that keeps the
+    /// whole tag O(N).
+    ///
+    /// [#969]: https://github.com/tafia/quick-xml/issues/969
     #[inline]
     fn check_for_duplicates(
         &mut self,
@@ -1070,6 +1138,9 @@ impl IterState {
         key: Range<usize>,
     ) -> Result<Range<usize>, AttrError> {
         if self.check_duplicates {
+            if self.keys.len() >= SMALL_ATTRIBUTE_COUNT {
+                return self.check_for_duplicates_hashed(slice, key);
+            }
             if let Some(prev) = self
                 .keys
                 .iter()
@@ -1079,6 +1150,44 @@ impl IterState {
             }
             self.keys.push(key.clone());
         }
+        Ok(key)
+    }
+
+    /// Cold path of [`Self::check_for_duplicates`] for start tags with many
+    /// attributes: a [`HashSet`] of 64-bit name hashes acts as an O(1) pre-filter
+    /// so iterating N attributes is O(N) rather than O(N²).
+    #[cold]
+    fn check_for_duplicates_hashed(
+        &mut self,
+        slice: &[u8],
+        key: Range<usize>,
+    ) -> Result<Range<usize>, AttrError> {
+        let keys = &self.keys;
+        let key_hashes = self.key_hashes.get_or_insert_with(|| {
+            // First time over the threshold: seed the set with the names already
+            // collected during the linear phase so the pre-filter knows them.
+            let mut set = HashSet::with_capacity_and_hasher(
+                keys.len() * 2,
+                BuildHasherDefault::<IdentityHasher>::default(),
+            );
+            for r in keys {
+                set.insert(hash_name(&slice[r.clone()]));
+            }
+            set
+        });
+        // A fresh hash proves the name is new. On a hit (a real duplicate, or the
+        // astronomically rare 64-bit collision) fall back to the linear scan to
+        // recover the exact previous position for `AttrError::Duplicated`.
+        if !key_hashes.insert(hash_name(&slice[key.clone()])) {
+            if let Some(prev) = self
+                .keys
+                .iter()
+                .find(|r| slice[(*r).clone()] == slice[key.clone()])
+            {
+                return Err(AttrError::Duplicated(key.start, prev.start));
+            }
+        }
+        self.keys.push(key.clone());
         Ok(key)
     }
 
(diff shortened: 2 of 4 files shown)


dralley:
And the distinct attrs column, I assume that represents N attributes within one element, as opposed to N attributes interspersed across many elements?

qifan-sailboat:
Yes. It means N distinct attributes on a single start tag.

src/events/attributes.rs
```diff
@@ -1070,12 +1078,22 @@ impl IterState {
         key: Range<usize>,
     ) -> Result<Range<usize>, AttrError> {
         if self.check_duplicates {
-            if let Some(prev) = self
-                .keys
-                .iter()
-                .find(|r| slice[(*r).clone()] == slice[key.clone()])
-            {
-                return Err(AttrError::Duplicated(key.start, prev.start));
+            let mut h = DefaultHasher::new();
```
dralley: SipHash (the Rust default) is fairly expensive in computational terms, though I suppose if the goal is DoS-resistance is the most "safe" option.

Nonetheless would you mind running some benchmarks (e.g. https://github.com/tafia/quick-xml/blob/master/benches/microbenches.rs#L177) to compare how the existing implementation compares to this implementation w/ SipHash and aHash for normal inputs?  Or at least the former two.

I would just like an idea of what impact this will have for the standard case.

dralley:
Please run `cargo fmt`

src/events/attributes.rs
```diff
@@ -1988,6 +2006,35 @@ mod xml {
                 assert_eq!(iter.next(), None);
                 assert_eq!(iter.next(), None);
             }
+
+            /// Regression test for <https://github.com/tafia/quick-xml/issues/969>:
+            /// the duplicate-attribute check used a linear `Vec` scan per attribute,
+            /// so a start tag with many distinct attribute names cost O(N²) byte
+            /// comparisons. With the hash pre-filter the same input is O(N).
+            #[test]
+            fn many_distinct_attributes_is_linear() {
```
dralley: I don't love the idea of having something like this in a unit test.  I suppose if it's reliable enough, it could be OK.  On the other hand maybe it's best to just have a criterion microbenchmark, those tend to get run for any important implementation changes.

@Mingun , thoughts

Mingun: Yes, I agree, this should be placed in `./benches` directory. I suggest to place it into `./benches/issue971.rs`. Benchmarks runs as tests on CI, so we will get the same results as for unit test:
https://github.com/tafia/quick-xml/blob/9aaea9281d346ec0249c679639a15eef8f9cbb18/.github/workflows/rust.yml#L64-L65

src/events/attributes.rs
```diff
@@ -979,6 +981,11 @@ pub(crate) struct IterState {
     /// names. We store a ranges instead of slices to able to report a previous
     /// attribute position
     keys: Vec<Range<usize>>,
+    /// 64-bit hashes of the byte content of `keys`, used as an O(1) pre-filter so
+    /// the duplicate check is amortised O(N) over the whole start tag instead of
+    /// O(N²). Lazily allocated on first use so that [`IterState::new`] can stay
+    /// `const`.
+    key_hashes: Option<HashSet<u64>>,
```
Mingun: It seems, we should use `NoHasher` here because we already put the hash to the set.

Changelog.md
```diff
@@ -18,6 +18,14 @@
 
 ### Bug Fixes
 
+- [#969]: `Attributes` (and anything that iterates `BytesStart::attributes()`
+  with the default `with_checks(true)`) no longer takes O(N²) time on a start
+  tag with N attributes. The duplicate-name check now uses a 64-bit hash
+  pre-filter so the no-duplicate path is amortised O(1) per attribute; the
+  exact `AttrError::Duplicated(new, prev)` positions are unchanged.
+
+[#969]: https://github.com/tafia/quick-xml/issues/969
```
Mingun: We put links at the end of version section (here you put it at the end of **Bug Fixes** section). Could you please move it below (keep the 2 blank lines before the next `##` section)

Mingun (changes requested):
Please move the test into the benchmark directory and could you investigate the possibility to address other comments. You may force-push the changes, because that PR is small enough (and anyway, we prefer to have clean history).

Changelog.md
dralley: @qifan-sailboat Please rebase your PR and fix both this note and the link that was added in the other PR

dralley:
@qifan-sailboat Do you plan to pick this back up?

qifan-sailboat:
Thanks @dralley and @Mingun for the reviews. Rebased onto `master` (so this now sits on top of #970) and force-pushed a single clean commit. Summary of the changes plus the benchmarks @dralley asked for.

## Design: keep the linear scan for small tags, hash only for large ones

@dralley's point about SipHash being expensive is the right one, so I leaned into it: the O(N²) only bites for tags with *many* attributes, and real start tags carry a handful — where the linear scan is faster than any hashing and needs no allocation. So this revision keeps the existing linear scan for tags with up to `SMALL_ATTRIBUTE_COUNT` (= 32) attributes and only switches to a hash pre-filter above that. On the switch it seeds the set from the names already collected, and on a hit it still falls back to the linear scan to recover the exact previous position, so `AttrError::Duplicated(new, prev)` is unchanged.

## Benchmarks

Cost of the duplicate check for a single start tag of N distinct attributes (`--release`, standalone harness so the numbers isolate the check itself):

| N (attrs / tag) | existing (linear, O(N²)) | SipHash → default set | SipHash → **NoHasher** set | aHash | **this PR (threshold)** |
|--:|--:|--:|--:|--:|--:|
| 64   | 3.5 µs | 2.2 µs | 1.4 µs | 1.4 µs | **1.7 µs** |
| 256  | 54 µs  | 8.7 µs | 4.8 µs | 4.8 µs | **4.9 µs** |
| 1024 | 904 µs | 34 µs  | 18 µs  | 18 µs  | **18 µs**  |

The linear column is cleanly quadratic (~16× per 4× of N); by 8192 attributes it is into the tens-to-hundreds of milliseconds, while every hash variant stays ~linear (~0.15 ms).

For the **standard case** I used the existing `attributes` micro-benchmark (`players.xml`, ≤ 22 attributes per element, so it never leaves the linear path):

| `attributes/with_checks = true` | `master` | this PR |
|--|--:|--:|
| | 32.8 µs | 32.9 µs |

That delta is within run-to-run noise — the `with_checks(false)` path, which this change never touches, drifts by ±1–3.5 % between runs on this machine, i.e. more than the `true` delta. So there's no measurable standard-case impact.

Takeaways:
- **Standard case:** at parity with `master`, because small tags run the identical linear scan. The earlier all-SipHash revision added a ~2–4× constant factor (plus a per-tag allocation) here — the regression you were worried about.
- **NoHasher (@Mingun):** done. The `HashSet<u64>` keys on an identity hasher now, since the values are already hashes. ~2× faster than the default `HashSet<u64>` on the hash path (e.g. 146 µs vs 271 µs at 8192 attrs).
- **aHash vs SipHash:** within ~2 % of each other once the set uses NoHasher (142 vs 146 µs at 8192), so aHash buys nothing for short attribute names — I kept SipHash and added **no new dependency**. With the threshold the hash never runs on normal documents anyway.
- **Pathological tags:** O(N) instead of O(N²).

## Other review comments
- **Move the test to a benchmark (@Mingun / @dralley):** the timing assertion is gone from the unit tests; there is now `benches/issue971.rs` (criterion, exercised by `cargo test --benches` on CI). I kept one small *correctness* unit test (`duplicate_past_hash_threshold`) covering a duplicate that straddles the linear→hash switch, since that path wasn't otherwise exercised.
- **Changelog (@Mingun):** moved the `[#969]` link reference (and the `[#970]` one added by the other PR) to the end of the version section, after `### Misc Changes`.
- **`cargo fmt` + rebase (@dralley):** applied and rebased onto current `master`.

Mingun (approved):
Thanks, this is wonderful!

dralley (approved):

PR #974, opened 2026-06-30
Fix incorrect decreasing of nesting level in NamespaceResolver

### LINKED ISSUE #953: `overlapped-lists` feature causes parser errors to `xsi:nil` fields after array fields
We have a rather complex XML parser written with `quick-xml`  that started failing some tests while trying to update from `quick-xml` 0.39.0 to 0.39.1. The old version did successful parse the document, the new version returned an error. 

I minimized both the data and the code down to the following example:

```rust
pub mod bml {
    use super::*;

    #[derive(Debug, Clone, serde::Deserialize)]
    pub struct BoreholeType {
        #[serde(rename = "boreholeSegment")]
        pub borehole_segment: Vec<bml::BoreholeSegmentPropertyType>,
        #[serde(rename = "drillingProcess")]
        pub drilling_process: Option<Vec<DrillingProcess>>,
    }

    #[derive(Debug, Clone, serde::Deserialize)]
    pub struct BoreholeSegmentType {}

    #[derive(Debug, Clone, serde::Deserialize)]
    pub struct DrillingProcess {
        #[serde(rename = "DrillingProcess")]
        pub drilling_process: bml::DrillingProcessType,
    }

    #[derive(Debug, Clone, serde::Deserialize)]
    pub struct DrillingProcessType {
        #[serde(rename = "to")]
        pub to: gml::LengthType,
    }

    #[derive(Debug, Clone, serde::Deserialize)]
    pub struct BoreholeSegmentPropertyType {
        #[serde(rename = "BoreholeSegment")]
        pub borehole_segment: bml::BoreholeSegmentType,
    }
}

pub mod gml {

    #[derive(Debug, Clone, serde::Deserialize)]
    pub struct LengthType {
        #[serde(rename = "$value")]
        pub value: Option<f64>,
    }
}

fn main() {
    let input = r#"
<?xml version='1.0' encoding='UTF-8'?>
    <bml:Borehole xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/deprecatedTypes.xsd http://www.infogeo.de/boreholeml/3.0 https://www.infogeo.de/boreholeml/3.0/BoreholeML.xsd" xmlns:gml="http://www.opengis.net/gml/3.2" xmlns:bml="http://www.infogeo.de/boreholeml/3.0">
      <bml:boreholeSegment>
        <bml:BoreholeSegment>
          <bml:from>0.0</bml:from>
        </bml:BoreholeSegment>
      </bml:boreholeSegment>
      <bml:drillingProcess xsi:nil="true"/>
    </bml:Borehole>
"#;
    let res = quick_xml::de::from_str::<bml::BoreholeType>(&input);
    assert!(res.is_ok(), "Error: {}", res.unwrap_err());
}

```

With that I run `git bisect` against all revisions between the 0.39.0 and the 0.39.1 tag. This reported 210d3e1d460a4c63e13d455b0b15623b77fa669c as the first failing commit. 

For me this looks like something around the parsing of `xsi:nil="true"` must have changed there as the parser now tries to parse inner fields of a null value.

weiznich:
This commit fixes an incorrect decreasing of the nesting level in the NamespaceResolver through the ns_reader. We could end up there in a state where we request a future decreasing of the nesting level without pushing an actual increase. This could cause namespaces to be unregistered too early.

Fix #953

### DIFF
--- a/src/reader/ns_reader.rs
+++ b/src/reader/ns_reader.rs
@@ -10,7 +10,7 @@ use std::ops::Deref;
 use std::path::Path;
 
 use crate::errors::Result;
-use crate::events::{BytesText, Event};
+use crate::events::{BytesStart, BytesText, Event};
 use crate::name::{NamespaceResolver, QName, ResolveResult};
 use crate::reader::{Config, Reader, Span, XmlSource};
 
@@ -91,6 +91,10 @@ impl<R> NsReader<R> {
                 Ok(Event::Empty(e))
             }
             Ok(Event::End(e)) => {
+                // push a fake element to the `ns_resolver` as the pending pop will remove it again
+                // if we don't push this element we don't increase the level, but the subsequent pending pop will
+                // decrease it anyway.
+                self.ns_resolver.push(&BytesStart::new(""))?;
                 // notify next `read_event_impl()` invocation that it needs to pop this
                 // namespace scope
                 self.pending_pop = true;
--- a/tests/serde-de-xsi.rs
+++ b/tests/serde-de-xsi.rs
@@ -702,3 +702,48 @@ mod as_field {
         }
     }
 }
+
+/// Regression test for https://github.com/tafia/quick-xml/issues/953.
+mod issue953 {
+    use super::*;
+
+    #[derive(Debug, Deserialize)]
+    struct BoreholeType {
+        #[serde(rename = "boreholeSegment")]
+        borehole_segment: Vec<()>,
+
+        #[serde(rename = "drillingProcess")]
+        drilling_process: Option<DrillingProcess>,
+    }
+
+    #[derive(Debug, Deserialize)]
+    struct DrillingProcess {
+        #[serde(rename = "DrillingProcess")]
+        drilling_process: (),
+    }
+
+    // success before 210d3e1d460a4c63e13d455b0b15623b77fa669c
+    // failed since 210d3e1d460a4c63e13d455b0b15623b77fa669c
+    #[test]
+    fn open_close() {
+        let input = r#"
+        <Borehole xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+          <boreholeSegment> </boreholeSegment>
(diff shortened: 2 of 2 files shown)


src/reader/ns_reader.rs
```diff
@@ -91,6 +91,10 @@ impl<R> NsReader<R> {
                 Ok(Event::Empty(e))
             }
             Ok(Event::End(e)) => {
+                // push a fake element to the `ns_resolver` as the pending pop will remove it again
+                // if we don't push this element we don't increase the level, but the subsequent pending pop will
+                // decrease it anyway.
+                self.ns_resolver.push(&BytesStart::new(""))?;
```
Mingun, line 97: This will only mask the issue in some cases. [As described](https://github.com/tafia/quick-xml/issues/953#issuecomment-4366419578), the root cause lies in the [event rewind code](https://github.com/tafia/quick-xml/blob/4deda08abeffdc188c269360229cf47e12a77a9f/src/de/mod.rs#L2781-L2859), that does not save/restore snapshots of `NamespaceResolver` state. The correct fix should implement such snapshoting.

weiznich: I run out of time looking again at this this week. I cannot promise that I find time to have a in depth look anytime soon due to other priorities. It still would be helpful to have some more details about the fix you are thinking of, as this would make it easier to schedule some fixed amount of time for the fix instead of require a vague amount of time.

Mingun: Now I do not have any clear vision of what exactly the fix should look like. Definitely, we need to save the `NamespaceResolver` snapshot when a snapshot of events is made, but at the same time we don't want to save unnecessary (for example, if the XML nesting is very deep and there are dozens of namespace definitions at each level), then we don't want to duplicate all this bunch of information just to keep the last level.

PR #979, opened 2026-07-19
Bound NamespaceResolver nesting depth (fixes #977)

### LINKED ISSUE #977: Unbounded u16 depth counter in NamespaceResolver::push → panic / namespace-scope corruption on deeply nested XML
`NamespaceResolver::push` (`src/name.rs`) increments its depth counter with an unguarded `self.nesting_level += 1;` on a `u16` field (`nesting_level`, declared near `src/name.rs:514`). `push` is called once per `Start`/`Empty` event by `NsReader::process_event` (`src/reader/ns_reader.rs`), and there is no depth cap anywhere on the `read_event` path — so a document with **65 536 nested elements** overflows the counter. Note the asymmetry: the matching `pop()` already saturates (`self.set_level(self.nesting_level.saturating_sub(1));`, `src/name.rs:763`), but the increment does not. Reachable from the public `NsReader::read_resolved_event*` on untrusted XML.

**Impact — two facets, depending on build profile:**

- **`overflow-checks` builds** (all debug builds; release profiles that opt in): the 65 536-th `push` **panics** `attempt to add with overflow` → remote, pre-auth **DoS** from a ~490 KB document.
- **Default `--release`** (`overflow-checks` off): the `u16` silently **wraps** 65535 → 0, after which `set_level`'s scope-truncation (`rposition(|n| n.level <= level)`) runs on non-monotonic levels and **corrupts namespace resolution** — bindings from closed scopes leak, in-scope bindings vanish. For any consumer that trusts resolved namespaces (XML-DSig, SOAP action routing, element-by-namespace access decisions) this is a security-relevant correctness bug, not just a crash.

This is **live on `master`** (`src/name.rs` still `self.nesting_level += 1;` with `pop` `saturating_sub`), and is **distinct from the closed #970 / #972**, which capped *declarations per element* (`max_declarations_per_element` / `TooManyDeclarations`) — this is the orthogonal *depth* counter.

## How to reproduce

`cargo new repro && cd repro`, add `quick-xml = "0.41"` to `Cargo.toml`, drop this into `src/main.rs`:

```rust
use quick_xml::events::Event;

fn main() {
    let depth = 70_000;
    let xml = "<a>".repeat(depth) + &"</a>".repeat(depth); // ~490 KB
    let mut n = 0u64;
    loop {
            Ok((_, Event::Eof)) => break,
            Ok(_) => n += 1,
            Err(e) => { eprintln!("Err after {n} events: {e:?}"); break; }
        }
    }
    eprintln!("done, {n} events");
}
```

- `RUSTFLAGS="-Coverflow-checks=on" cargo run --release` → `thread 'main' panicked at src/name.rs:...: attempt to add with overflow`.
- Default `cargo run --release` → no panic, but the resolver is corrupted. Nesting one `xmlns:p` outside and a different `xmlns:p` inside a nested `<x>`, then resolving the same prefix after popping back out:

  | depth | inner `<p:e>` (expect `BBB`) | outer `<p:f>` (expect `AAA`) |
  |---|---|---|
  | 3 (control) | `BBB` ✓ | `AAA` ✓ |
  | 65534 | `BBB` | **`BBB`** ✗ (closed-scope binding leaked) |
  | ≥65535 | `BBB` | **`Unknown("p")`** ✗ (in-scope binding vanished) |

## Suggested fix

A `saturating_add` (mirroring `pop`) removes the panic but **not** the misresolution — at saturation all levels ≥ `u16::MAX` collapse to one and `set_level` still over-truncates (verified). The correct fix is a **depth cap that returns a clean error** at the `u16` boundary, in the same spirit as the existing `TooManyDeclarations` guard. I have a patch (a `NamespaceError::TooDeeplyNested` variant + `checked_add`, all existing tests pass incl. `--features serialize`) and am happy to open a PR.

scadastrangelove:
Fixes #977.

`NamespaceResolver::push` incremented its `u16` `nesting_level` with an unguarded `+= 1`. A document nested past `u16::MAX` overflowed it — a panic under `overflow-checks` builds, or a silent wrap that corrupts namespace-scope truncation in a stock release build (bindings from closed scopes leak; in-scope bindings vanish). The sibling `pop` already saturates (`saturating_sub`); only the increment was unguarded. Reachable from the public `NsReader::read_resolved_event*` on ~490 KB of deeply nested XML. Still present on `master`; orthogonal to #970/#972 (which capped *declarations per element* — this is the *depth* counter).

### Fix

Return a new `NamespaceError::TooDeeplyNested` at the `u16` boundary, matching the existing per-element `TooManyDeclarations` guard:

```rust
self.nesting_level = self.nesting_level
    .checked_add(1)
    .ok_or(NamespaceError::TooDeeplyNested(u16::MAX as usize))?;
```

A `saturating_add` (mirroring `pop`) was considered but is **insufficient** — it removes the panic but not the scope corruption: at saturation all deep levels collapse to `u16::MAX` and `set_level`'s `rposition(|n| n.level <= level)` truncation still over-truncates, so the namespace misresolution persists (verified). The clean error rejects the pathological document instead.

### Validation

- `cargo test` and `cargo test --features serialize`: all green (incl. the existing `push_rejects_too_many_declarations`).
- Added `push_rejects_pathological_nesting_depth` next to the #970 test — `u16::MAX` successful pushes, then a clean `Err(TooDeeplyNested)`.
- The 490 KB PoC from #977 now returns `Err(TooDeeplyNested)` instead of panicking, and the namespace-misresolution demonstration no longer occurs.

*Alternative considered:* widen `nesting_level` to `u32`/`usize` (overflow becomes practically unreachable). I chose the explicit cap + error to mirror `max_declarations_per_element` and keep the depth bound intentional on untrusted input — happy to switch if you'd prefer no new variant.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,13 @@
 
 ### Bug Fixes
 
+- [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
+  event) now returns the new `NamespaceError::TooDeeplyNested` when a document
+  nests elements deeper than `u16::MAX`, instead of overflowing the internal
+  `u16` depth counter. Previously the unguarded `nesting_level += 1` panicked
+  under `overflow-checks` builds and silently wrapped in release, corrupting
+  namespace-scope bookkeeping on deeply nested untrusted input.
+
 ### Misc Changes
 
 
@@ -47,6 +54,7 @@
 
 [#969]: https://github.com/tafia/quick-xml/issues/969
 [#970]: https://github.com/tafia/quick-xml/issues/970
+[#977]: https://github.com/tafia/quick-xml/issues/977
 
 
 ## 0.40.1 -- 2026-05-15
--- a/Changelog.md
+++ b/Changelog.md
@@ -27,6 +27,8 @@
 
 ### Misc Changes
 
+[#977]: https://github.com/tafia/quick-xml/issues/977
+
 
 ## 0.41.0 -- 2026-06-29
 
@@ -54,7 +56,6 @@
 
 [#969]: https://github.com/tafia/quick-xml/issues/969
 [#970]: https://github.com/tafia/quick-xml/issues/970
-[#977]: https://github.com/tafia/quick-xml/issues/977
 
 
 ## 0.40.1 -- 2026-05-15
--- a/src/name.rs
+++ b/src/name.rs
@@ -47,6 +47,10 @@ pub enum NamespaceError {
     /// This bounds the heap allocated by [`NamespaceResolver::push`] (and hence
     /// by [`NsReader`](crate::reader::NsReader)) on untrusted input.
     TooManyDeclarations(usize),
+    /// The document nested elements more deeply than the namespace resolver's
+    /// depth counter (a `u16`) can track. This bounds stack / scope-bookkeeping
+    /// work on untrusted input. Contains the depth limit that was exceeded.
+    TooDeeplyNested(usize),
 }
 
 impl fmt::Display for NamespaceError {
@@ -84,6 +88,9 @@ impl fmt::Display for NamespaceError {
                      raise the limit with NamespaceResolver::set_max_declarations_per_element",
                     limit,
                 )
+            }
+            Self::TooDeeplyNested(limit) => {
+                write!(f, "document nests elements deeper than the supported limit of {}", limit)
             }
         }
     }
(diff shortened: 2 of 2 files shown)


Changelog.md
```diff
@@ -47,6 +54,7 @@
 
 [#969]: https://github.com/tafia/quick-xml/issues/969
 [#970]: https://github.com/tafia/quick-xml/issues/970
+[#977]: https://github.com/tafia/quick-xml/issues/977
```
Mingun: This should be on line 30, in the **Unreleased** section

Mingun (approved):
Only one small pit, and need to fix formatting.

dralley:
@scadastrangelove Please run `cargo fmt`, otherwise looks good

scadastrangelove:
Done — pushed `cargo fmt`.

dralley (approved):

PR #981, opened 2026-07-19
Fix O(depth²) namespace prefix resolution in NamespaceResolver

### LINKED ISSUE #980: NamespaceResolver::resolve_prefix linearly scans the entire binding stack per lookup — O(depth²) CPU DoS on deeply nested XML (sibling of RUSTSEC-2026-0194)
# ISSUE — quick-xml: `NamespaceResolver::resolve_prefix` is O(depth) per lookup → O(depth²) CPU DoS on deeply nested XML (sibling of RUSTSEC-2026-0194)

**Title:** `NamespaceResolver::resolve_prefix` linearly scans the entire binding stack per lookup — quadratic CPU cost on deeply nested documents

### Summary

RUSTSEC-2026-0194 (#969) fixed an O(N²) linear scan in `check_for_duplicates` (comparing each attribute
name against every prior name on the same tag) by adding a 64-bit hash pre-filter once a tag has more than
`SMALL_ATTRIBUTE_COUNT` (32) attributes (`events/attributes.rs`, fix commit 07f3db8).

`NamespaceResolver::resolve_prefix` (`src/name.rs:953`) has the identical shape, over a different axis:

```rust
let mut iter = self.bindings.iter().rev();
```
followed by a full linear `.find(...)` over the **entire** in-scope `bindings` stack — once per element
name resolved (line 958, unprefixed name with no default namespace: `iter.find(|n| n.prefix_len == 0)`) or
per prefixed name whose prefix was never bound (line 963: `iter.find(|n| n.prefix(&self.buffer) == prefix)`).
`bindings` grows by one entry per `xmlns[:*]` declaration on every currently-open ancestor element, so its
length is proportional to nesting **depth**. There is no `HashMap<prefix, binding>` index and no cap on
`bindings.len()` or nesting depth (only the u16 `nesting_level` ceiling from #977). A lookup that misses
(the common case for unprefixed names with no default namespace, or a genuinely unbound prefix) is
**O(depth)**; doing that once per event down a depth-D document is **O(D²)**.

### Measured (release build, single core)

Each nesting level declares one `xmlns:a` binding (so `bindings` grows) but element names are unprefixed
with no default namespace declared, so every `resolve_prefix` call misses and scans the full stack:

| N (nesting depth) | events | time | ratio vs previous (÷2N) |
|---|---|---|---|
| 5,000 | 10,000 | 0.019 s | — |
| 10,000 | 20,000 | 0.083 s | ×4.33 |
| 20,000 | 40,000 | 0.360 s | ×4.33 |
| 40,000 | 80,000 | 1.463 s | ×4.06 |

Each doubling of N roughly quadruples the time — textbook O(N²). A ~1.3 MB crafted document (`N=200,000`
would be needed to hit the same order the linked advisory used for #969) drives multi-second single-core
CPU stalls from ordinary-looking nested XML with one namespace declaration per level — no huge tag, no
huge attribute count, just ordinary depth.

### PoC

```rust
use quick_xml::events::Event;

fn main() {
    let n = 40_000;
    let xml = "<e xmlns:a='x'>".repeat(n) + &"</e>".repeat(n);
    loop {
            Ok((_, Event::Eof)) | Err(_) => break,
            Ok(_) => {}
        }
    }
}
```

### Suggested direction

Same shape of fix as #969's: an index (e.g. a `HashMap<prefix-hash, Vec<index>>` maintained alongside
`push`/`pop`, or a hash pre-filter analogous to `check_for_duplicates`'s once `bindings.len()` crosses a
threshold) so a miss doesn't require scanning every in-scope binding. Since `bindings` is a stack
(push/pop tracks scope), the index needs pop-side maintenance too — a bit more involved than #969's
per-tag scan, so I'm not proposing a specific PR; happy to help if useful.

Found with the [rust-in-peace](https://github.com/scadastrangelove/rust-in-peace) security pipeline.

AnayGarodia:
Fixes #980.

## Problem

`NamespaceResolver::resolve_prefix` did a reverse linear scan of the entire
in-scope `bindings` stack on every call:

```rust
let mut iter = self.bindings.iter().rev();
...
(None, true) => match iter.find(|n| n.prefix_len == 0) { ... }
...
(Some(p), _) => match iter.find(|n| n.prefix(&self.buffer) == prefix) { ... }
```

`bindings` grows by one entry per `xmlns[:prefix]` declaration on every
currently open ancestor element, so its length tracks nesting depth. A
lookup that misses -- the common case for an unprefixed name with no
default namespace, or a genuinely undeclared prefix -- costs O(depth); doing
that once per event (`NsReader::read_resolved_event*` resolves every
`Start`/`Empty`/`End`) down a depth-D document costs O(D²). A ~1.3 MB
crafted document with one `xmlns:*` declaration per level (no huge tag, no
huge attribute count -- just ordinary depth) drives multi-second single-core
CPU stalls.

This is the same shape of problem as #969/#971 (`check_for_duplicates`
scanning every previously-seen attribute name), just over a different axis
(nesting depth vs. attribute count on one tag), as the issue calls out.

## Root cause

`NamespaceBinding`s are stored as a flat `Vec` acting as a stack: `push`
(via `add`) always appends, and `pop` (via `set_level`) always truncates
from the back when an element closes. There was no index from prefix to
position, so every lookup had no choice but to scan.

## Fix

Small stacks (up to `SMALL_BINDING_COUNT = 32` bindings) keep the direct
linear scan -- for a handful of in-scope bindings it's faster than hashing
and needs no allocation, which covers the overwhelming majority of real
documents. Once the stack grows past that threshold, `resolve_prefix` is
backed by a `HashMap<Vec<u8>, Vec<usize>>` index (prefix bytes, or an empty
key for the default namespace, to ascending stack positions), so a lookup
is O(1) average instead of O(depth).

Unlike the #971 fix (where the per-tag attribute list is thrown away and
rebuilt from scratch on every start tag), the namespace binding stack
persists and shrinks across the whole document, so the index needs
pop-side maintenance too, as the issue's "suggested direction" section
called out:

- `add()` (called by both the public API and `push()`) records each new
  binding into the index once it exists, seeding it from scratch (scanning
  every binding currently on the stack) the first time the threshold is
  crossed.
- `set_level()` (called by `pop()` and directly) removes exactly the
  entries it's about to truncate from the index before truncating
  `bindings`/`buffer`, using the fact that `bindings` only ever grows by
  appending or shrinks by truncating from the back -- so for every prefix,
  the positions being removed are exactly the trailing entries of that
  prefix's index list.

No new data structure conventions were introduced beyond what #971 already
established (`HashMap` gated behind a small/large threshold); no_std /
zero-alloc behavior for shallow documents is preserved since the index is
`None` (and never allocated) below the threshold.

One behavioral edge case is preserved exactly: an explicit *empty* prefix
(reachable via `QName(b":local").prefix()`, i.e. a malformed name like
`:local`) never resolves, because `NamespaceBinding::prefix` already
represents "no prefix" (the default namespace) the same way
(`prefix_len == 0`) internally, and the original code's
`n.prefix(&buffer) == Some(Prefix(b""))` comparison could never be true.
This is covered by a dedicated regression test.

## Verification

**Benchmark** (`benches/issue980.rs`, `cargo bench --bench issue980`),
resolving one element name per nesting level across a range of depths:

On unpatched `master` (via `git stash` of just `src/name.rs`):

| depth | throughput |
|---|---|
| 64 | ~16.0 Melem/s |
| 256 | ~9.2 Melem/s |
| 1024 | ~3.4 Melem/s |
| 4096 | ~0.91 Melem/s |
| 16384 | ~0.23 Melem/s |

Throughput drops ~70x over a 256x depth increase -- the O(depth²) signature
(each doubling of depth should roughly quarter throughput if the whole
document is O(depth²), matching the issue's own measurements).

With this fix, on the same range:

| depth | throughput |
|---|---|
| 64 | ~10.4 Melem/s |
| 256 | ~11.6 Melem/s |
| 1024 | ~11.8 Melem/s |
| 4096 | ~13.5 Melem/s |
| 16384 | ~13.6 Melem/s |

Throughput stays flat, confirming the fix.

**Tests**: two new regression tests in `src/name.rs` (`resolve_prefix_past_index_threshold`,
covering a hit near the bottom of a >32-deep stack, a hit at the top, a
miss, the default namespace, and -- the key regression check for pop-side
maintenance -- that popping back out of scope correctly un-resolves a
previously-found prefix without panicking or returning stale data; and
`resolve_prefix_with_empty_prefix_never_resolves`, covering the edge case
above both below and above the indexing threshold).

Ran locally and all pass:
- `cargo test --all-features --benches --tests` (1494 lib+doctest + all
  integration test binaries, 0 failures)
- `cargo test --no-default-features`
- `cargo test --features serialize`
- `cargo test --features serialize,encoding`
- `cargo +1.79.0 check` (pinned MSRV toolchain)
- `cargo fmt -- --check`
- `cargo clippy --all-features --all-targets` (no new warnings; the lib's
  only warning at HEAD, an `unnecessary_unwrap` in my first draft of
  `set_level`, is fixed in this PR's version)

## Note

I'm the account submitting this PR (not the issue reporter); I read the
full issue thread, confirmed no other open/merged PR already addresses
this axis of the namespace resolver (only the sibling `push`-side fixes in
#970/#972/#974/#977/#979 exist, none of which touch `resolve_prefix`), and
independently confirmed the O(depth²) behavior with the benchmark above
before writing the fix.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -24,10 +24,14 @@
   `u16` depth counter. Previously the unguarded `nesting_level += 1` panicked
   under `overflow-checks` builds and silently wrapped in release, corrupting
   namespace-scope bookkeeping on deeply nested untrusted input.
+- [#980]: `NamespaceResolver::resolve_prefix` (and hence every `NsReader`
+  `Start`/`Empty`/`End` event resolved through `read_resolved_event*`) no
+  longer takes O(depth) time per lookup on deeply nested documents.
 
 ### Misc Changes
 
 [#977]: https://github.com/tafia/quick-xml/issues/977
+[#980]: https://github.com/tafia/quick-xml/issues/980
 
 
 ## 0.41.0 -- 2026-06-29
--- a/src/name.rs
+++ b/src/name.rs
@@ -7,6 +7,7 @@ use crate::events::attributes::Attribute;
 use crate::events::{BytesStart, Event};
 use crate::utils::{write_byte_string, Bytes};
 use memchr::memchr;
+use std::collections::HashMap;
 use std::fmt::{self, Debug, Formatter};
 use std::iter::FusedIterator;
 
@@ -528,6 +529,10 @@ pub struct NamespaceResolver {
     /// [`NamespaceError::TooManyDeclarations`]. See
     /// [`set_max_declarations_per_element`](Self::set_max_declarations_per_element).
     max_declarations_per_element: usize,
+    /// An index from prefix bytes (empty slice for the default namespace) to
+    /// ascending positions in [`Self::bindings`]. `None` until the stack grows
+    /// past [`SMALL_BINDING_COUNT`], so shallow documents never build it.
+    prefix_index: Option<HashMap<Vec<u8>, Vec<usize>>>,
 }
 
 /// Default limit on the number of `xmlns` / `xmlns:*` declarations
@@ -539,6 +544,11 @@ pub struct NamespaceResolver {
 /// a few kilobytes regardless of input size.
 pub const DEFAULT_MAX_DECLARATIONS_PER_ELEMENT: usize = 256;
 
+/// Number of [`NamespaceBinding`]s that may sit on [`NamespaceResolver::bindings`]
+/// before [`NamespaceResolver::resolve_prefix`] switches from a direct reverse
+/// linear scan to a `HashMap` index (see [`NamespaceResolver::prefix_index`]).
+const SMALL_BINDING_COUNT: usize = 32;
+
 impl Debug for NamespaceResolver {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         f.debug_struct("NamespaceResolver")
@@ -549,6 +559,7 @@ impl Debug for NamespaceResolver {
                 "max_declarations_per_element",
                 &self.max_declarations_per_element,
             )
+            .field("prefix_index", &self.prefix_index)
             .finish()
     }
 }
@@ -600,6 +611,7 @@ impl Default for NamespaceResolver {
             bindings,
             nesting_level: 0,
             max_declarations_per_element: DEFAULT_MAX_DECLARATIONS_PER_ELEMENT,
+            prefix_index: None,
         }
     }
 }
@@ -676,6 +688,7 @@ impl NamespaceResolver {
                     value_len: namespace.0.len(),
                     level,
                 });
+                self.index_binding(&[]);
             }
             PrefixDeclaration::Named(b"xml") => {
                 if namespace != RESERVED_NAMESPACE_XML.1 {
@@ -707,9 +720,46 @@ impl NamespaceResolver {
                     value_len: namespace.0.len(),
                     level,
                 });
+                self.index_binding(prefix);
             }
         }
         Ok(())
+    }
+
+    /// Records the binding just pushed onto the back of [`Self::bindings`] in
+    /// [`Self::prefix_index`], keyed by `key` (the prefix bytes, or an empty
+    /// slice for the default namespace).
+    ///
+    /// Below [`SMALL_BINDING_COUNT`] this is a no-op, because a reverse linear
+    /// scan over a handful of bindings beats hashing and allocates nothing.
+    /// The first time the threshold is crossed the index is built from the
+    /// whole stack at once; after that every push extends it here, and every
+    /// pop truncates it in [`Self::set_level`].
+    fn index_binding(&mut self, key: &[u8]) {
+        let index = self.bindings.len() - 1;
+        if self.prefix_index.is_none() {
+            if self.bindings.len() <= SMALL_BINDING_COUNT {
+                return;
+            }
+            let mut map: HashMap<Vec<u8>, Vec<usize>> = HashMap::new();
+            for (i, n) in self.bindings.iter().enumerate() {
+                let k: &[u8] = match n.prefix(&self.buffer) {
+                    Some(p) => p.into_inner(),
+                    None => &[],
+                };
+                map.entry(k.to_vec()).or_default().push(i);
+            }
+            self.prefix_index = Some(map);
+            return;
+        }
+        // `unwrap` is safe: the `is_none()` branch above always returns.
+        let map = self.prefix_index.as_mut().unwrap();
+        match map.get_mut(key) {
+            Some(positions) => positions.push(index),
+            None => {
+                map.insert(key.to_vec(), vec![index]);
+            }
+        }
     }
 
     /// Begins a new scope and add to it all [namespace bindings] that found in
@@ -824,10 +874,35 @@ impl NamespaceResolver {
             None => {
                 self.buffer.clear();
                 self.bindings.clear();
+                if let Some(map) = &mut self.prefix_index {
+                    map.clear();
+                }
             }
             // drop all namespaces past the last valid namespace
             Some(last_valid_pos) => {
                 if let Some(len) = self.bindings.get(last_valid_pos + 1).map(|n| n.start) {
+                    // Keep `prefix_index` (if built) in sync with the bindings
+                    // we're about to drop. `bindings` only grows by appending
+                    // and shrinks from the back, so the positions going away
+                    // are always the trailing entries of a prefix's index
+                    // list -- one pop per removed binding is enough.
+                    if let Some(map) = &mut self.prefix_index {
+                        for n in &self.bindings[last_valid_pos + 1..] {
+                            let key: &[u8] = match n.prefix(&self.buffer) {
+                                Some(p) => p.into_inner(),
+                                None => &[],
+                            };
+                            let remove_key = if let Some(positions) = map.get_mut(key) {
+                                positions.pop();
+                                positions.is_empty()
+                            } else {
+                                false
+                            };
+                            if remove_key {
+                                map.remove(key);
+                            }
+                        }
+                    }
                     self.buffer.truncate(len);
                     self.bindings.truncate(last_valid_pos + 1);
                 }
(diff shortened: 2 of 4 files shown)


Changelog.md
```diff
@@ -24,10 +24,22 @@
...
+  `Start`/`Empty`/`End` event resolved through `read_resolved_event*`) no
+  longer takes O(depth) time per lookup on deeply nested documents. It
+  previously did a reverse linear scan of the entire in-scope `bindings`
+  stack on every call, so a lookup that misses -- the common case for an
+  unprefixed name with no default namespace, or a genuinely undeclared
+  prefix -- cost O(depth), and a full document cost O(depth²). Stacks of up
+  to `SMALL_BINDING_COUNT` (32) bindings keep the previous linear scan;
+  larger ones are backed by a `HashMap` index from prefix to stack
+  positions, maintained incrementally on both push and pop, so lookups stay
+  O(1) average regardless of nesting depth.
```
dralley: Please make this a bit more concise, it's a bit excessive.  Could even just keep the very first sentence.

src/name.rs
```diff
@@ -1337,6 +1464,169 @@ mod namespaces {
         );
     }
 
+    /// Regression test for <https://github.com/tafia/quick-xml/issues/980>:
+    /// `resolve_prefix` used to always do a reverse linear scan of the whole
+    /// `bindings` stack, making a lookup O(depth). Once the stack grows past
+    /// `SMALL_BINDING_COUNT`, `NamespaceResolver` switches to a `HashMap`
+    /// index (built and maintained by `index_binding` / `set_level`); this
+    /// checks that resolution stays correct once that switch has happened,
```
dralley: Simplify all this language.  The implementation details don't matter much beyond the very highest level - these are tests.  All that needs to be said is that there are two implementations, they are switched based on the number of namespace bindings, and we want to check that both are correct.

src/name.rs
```diff
@@ -528,6 +529,16 @@ pub struct NamespaceResolver {
     /// [`NamespaceError::TooManyDeclarations`]. See
     /// [`set_max_declarations_per_element`](Self::set_max_declarations_per_element).
     max_declarations_per_element: usize,
+    /// An index from prefix bytes (empty slice for the default namespace) to
+    /// ascending positions in [`Self::bindings`], used by
+    /// [`resolve_prefix`](Self::resolve_prefix) to avoid a linear scan of the
+    /// whole stack once it grows past [`SMALL_BINDING_COUNT`]. `None` until
```
dralley: Again, it's not needed to describe the implementation multiple times in different places - here we are just defining the variable, it's enough to say what it's for rather than exactly what will be done with it later.  The details can be provided at the place where it is being used.

src/name.rs
```diff
@@ -539,6 +550,20 @@ pub struct NamespaceResolver {
 /// a few kilobytes regardless of input size.
 pub const DEFAULT_MAX_DECLARATIONS_PER_ELEMENT: usize = 256;
 
+/// Number of [`NamespaceBinding`]s that may sit on [`NamespaceResolver::bindings`]
+/// before [`NamespaceResolver::resolve_prefix`] switches from a direct reverse
+/// linear scan to a `HashMap` index (see [`NamespaceResolver::prefix_index`]).
```
dralley, line 549: Same.  This is enough information.

src/name.rs
```diff
@@ -707,11 +735,51 @@ impl NamespaceResolver {
...
+    /// `key` (the prefix bytes, or an empty slice for the default namespace).
+    ///
+    /// While `bindings.len()` stays at or below [`SMALL_BINDING_COUNT`] this is
+    /// a no-op: [`Self::resolve_prefix`] uses a direct linear scan for those,
+    /// which is faster than hashing and needs no allocation. The first time the
+    /// threshold is crossed, the index is built from scratch (seeded with every
+    /// binding currently on the stack, including the one just pushed); after
+    /// that it is kept up to date on every subsequent push, and truncated in
+    /// lock-step on pop by [`Self::set_level`].
+    fn index_binding(&mut self, key: &[u8]) {
```
dralley, line 738: Detail is fine here so long as it's not duplicative and flows well. Please read through the rest of this yourself and make sure it makes sense as-written, I'll take another look after you've done a pass on everything.  

I assume the code is AI generated as well - that's fine - but please make sure you've done a review yourself and understand exactly what it's doing before submitting if you haven't already.

dralley:
Cut the commit message down a bit as well, there is no need to repeat every detail.

Changelog.md
AnayGarodia: Cut to just the first sentence.

src/name.rs
AnayGarodia: Rewritten to your wording: two implementations, switched on the number of bindings, both need to be correct.

src/name.rs
AnayGarodia: Trimmed to what the field is, dropped the description of how it's maintained. That now lives on `index_binding`.

src/name.rs
AnayGarodia: Dropped the second paragraph, first one kept.

src/name.rs
AnayGarodia: Kept the detail here but removed what it duplicated from the field and const docs, so this is now the one place the threshold behaviour is explained.

AnayGarodia:
@dralley done, force-pushed as e7bac3e. Net effect is -105/+51 on prose, no code changes.

Beyond the five inline comments, I did the full pass you asked for and cut the same kind of duplication everywhere else it had crept in: `last_binding`'s doc, the `set_level` truncation comment, the empty-prefix arm in `resolve_prefix`, the `unwrap` justification, the inline comments in the new test, and the `issue980` bench header (which had drifted longer than `issue971`'s, so I brought it back in line). The commit message is cut down too.

The rule I applied: the threshold behaviour is explained once, on `index_binding`, and everything else just says what it is and points there.

On the AI question: yes, this was AI-assisted. I've read and understand the whole change, and re-verified it locally on the final tree rather than assuming the earlier runs still held:

- `cargo test --all-features --benches --tests` — all suites pass
- `cargo test --no-default-features` and `cargo test --features serialize,encoding` — pass
- `cargo fmt -- --check` clean, `cargo clippy --all-features --all-targets` adds no new warnings (the 8 remaining are pre-existing, none in `name.rs` or the new bench)

Happy to trim further if any of it still reads long.

dralley:
There are a couple of merge conflicts, luckily nothing too bad.  Please clean those up on the next push

src/name.rs
```diff
@@ -707,11 +720,48 @@ impl NamespaceResolver {
...
                     level,
                 });
+                self.index_binding(prefix);
             }
         }
         Ok(())
     }
 
+    /// Records the binding just pushed onto the back of [`Self::bindings`] in
+    /// [`Self::prefix_index`], keyed by `key` (the prefix bytes, or an empty
```
dralley, line 730: How this is called is irrelevant and shouldn't be part of the doc comment. Only what it does is relevant, usages can be found by search.  If you want to document the whole flow of events from one call to the next, then add that to the `NamespaceResolver` struct or `add()` function, where it would make more sense.

dralley:
The more I think about it, the more I feel like this should probably just be handled like entity resolution loops - just yield an error if the threshold is reached.  I don't think most legitimate XML is going to have more than perhaps two dozen namespaces - let's say a few hundred in really truly extreme cases.  An XML document with thousands of namespaces seems very likely to be maliciously constructed.  This is especially true if it's constructed such that those hundreds++ of namespaces are declared on separate elements to avoid the `max_declarations_per_element` checks that were [added recently](https://github.com/tafia/quick-xml/commit/7ca25266e94987210daa864889ab15c9332c8a2a)

That feels potentially more appropriate than just trying to make a potential DoS vector faster.

It would also be more of a tweak of existing behavior, because `max_declarations_per_element` is precisely such a cap - just a bit too narrowly defined.

@Mingun what do you think

edit: see https://github.com/tafia/quick-xml/pull/987

(it's not ready for a full review, just conceptually speaking, do we prefer that over this)

dralley:
Thanks, but I think we will use a different approach to solving this issue

PR #982, opened 2026-07-20
Add recursion-depth limit to serde Deserializer

### LINKED ISSUE #819: Deserialization of enum variant which recursively refers to itself faild with stackoverflow
Found in https://stackoverflow.com/questions/67932584/deserializing-recursive-xml-using-serde-in-rust

Currently recursively defined enums using newtype variants for recursion lead to stack overflow error:
```rust
#[test]
fn recursive() {
    #[serde(rename_all = "camelCase")]
    enum MathNode {
        Apply(Vec<MathNode>),
        Ci(Vec<MathNode>),
        #[serde(rename = "$text")]
        Text(String),
        #[serde(rename = "math")]
        Root(Vec<MathNode>),
    }

    let test = r#"
    <math>
        <apply>
            <ci type="integer">5</ci>
        </apply>
    </math>"#;
    assert_eq!(
        from_str::<MathNode>(test).unwrap(),
        MathNode::Root(vec![MathNode::Apply(vec![MathNode::Ci(vec![
            MathNode::Text("5".into())
        ])])])
    );
}
```
However, if change newtype variant to the struct variant with one `$value` field, everything works:
```rust
#[test]
fn recursive() {
    #[serde(rename_all = "camelCase")]
    enum MathNode {
        Apply {
            #[serde(rename = "$value")]
            value: Vec<MathNode>,
        },
        Ci {
            #[serde(rename = "$value")]
            value: Vec<MathNode>,
        },
        #[serde(rename = "$text")]
        Text(String),
        #[serde(rename = "math")]
        Root {
            #[serde(rename = "$value")]
            value: Vec<MathNode>,
        },
    }

    let test = r#"
    <math>
        <apply>
            <ci type="integer">5</ci>
        </apply>
    </math>"#;
    assert_eq!(
        from_str::<MathNode>(test).unwrap(),
        MathNode::Root {
            value: vec![MathNode::Apply {
                value: vec![MathNode::Ci {
                    value: vec![MathNode::Text("5".into())]
                }],
            }],
        }
    );
}
```
It seems worth to implement newtype variant deserialization like such structs

### LINKED ISSUE #978: serde Deserializer has no recursion-depth limit → stack-overflow DoS on deeply nested untrusted XML (distinct from #819)
**Relationship to #819:** #819 is a *correctness* bug — a **tiny** recursive-**newtype**-enum document overflows even at depth ~3 (the struct-variant `$value` form works fine at the same depth). This report is **different and orthogonal**: the *working* deserialization path (struct variants, or `serde_json::Value` via `deserialize_any`) has **no recursion-depth cap**, so an attacker-controlled **deeply nested** document overflows the native call stack — a **DoS**, not a type-shape bug. A fix for #819 (make newtype recursion terminate) would not address this; a depth cap would.

## Mechanism

`deserialize_struct` (`src/de/mod.rs`) consumes a `Start` and calls `visit_map(ElementMapAccess::new(..))` while its own frame stays live; a nested child re-enters `deserialize_struct` via `ElementMapAccess::next_value_seed` → `MapValueDeserializer::deserialize_struct` → `self.map.de.deserialize_struct`. Each XML nesting level adds several native stack frames. The only bound in `de/` is `event_buffer_size` (the `overlapped-lists` skip buffer) — it does not bound call-stack depth. No recursive *user* type is even required: `deserialize_any` routes self-describing targets (e.g. `serde_json::Value`) through the same path.

The crash is an **uncatchable `abort`** (not a `Result::Err`), so a service that deserializes untrusted XML with the `serialize` feature has a pre-auth crash it cannot handle. Worker threads with smaller stacks fail far earlier than the 8 MB main thread.

## How to reproduce

`cargo new repro && cd repro`, add `quick-xml = { version = "0.41", features = ["serialize"] }` and `serde = { version = "1", features = ["derive"] }`, drop this into `src/main.rs`, then `cargo run --release`:

```rust
use serde::Deserialize;

#[derive(Deserialize)]
struct S {
    v: Vec<S>, // the struct-variant "$value" shape that #819 says WORKS
}

fn main() {
    for depth in [3usize, 1000, 20_000, 60_000] {
        let xml = "<a>".repeat(depth) + &"</a>".repeat(depth);
        let _: Result<S, _> = quick_xml::de::from_str(&xml);
        eprintln!("depth={depth} ({} bytes) -> returned", xml.len());
    }
}
```

Observed (release, default 8 MB main-thread stack):

```
depth=3 (21 bytes) -> returned
depth=1000 (7000 bytes) -> returned
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

The key point vs #819: the **struct-variant / `$value`** shape #819 calls *working* deserializes correctly at shallow depth (3, 1000) but **still aborts at attacker depth** (~20 000, 140 KB) — so this is depth-driven resource exhaustion on the normal path, independent of the newtype-variant correctness bug.

## Suggested direction

A configurable recursion-depth limit in the `Deserializer` that returns a clean `Error` past the cap — mirroring `serde_json`'s `Deserializer::recursion_limit` and quick-xml's own `event_buffer_size` DoS guard.

scadastrangelove:
Fixes #978.

The serde `Deserializer` had no recursion-depth cap on any of its re-entrant paths, so a maliciously deeply nested (but otherwise well-formed) untrusted XML document could crash the whole process via native stack overflow — uncatchable as a `Result::Err`.

## Correction to the record

The original issue (and my own follow-up comment) described three recursion cycles as independently depth-driven. Re-verifying against current `master`, two are: the original `deserialize_struct` → `next_value_seed` cycle, and a `deserialize_seq` cycle via a plain `Vec<Self>` field — both abort only past a real depth threshold (fine at depth 3, aborts by depth ~10,000-20,000). The third (enum `newtype_variant` re-entry) is **not** an independent depth-driven DoS — it's issue #819's existing non-terminating-recursion bug re-exhibited: `variant_seed`/`newtype_variant_seed` never advance the reader position, so it aborts at depth 1 regardless of any cap. This PR does not fix #819 — recursive newtype enums still can't successfully deserialize, they now just fail cleanly (`TooDeeplyNested`) instead of aborting, since the same guard sits at the shared re-entry point.

## Design

`Deserializer` gains `depth`/`max_depth: usize` (default 128, matching `serde_json`) and a public `recursion_limit()` setter. A `DepthGuard` (RAII, Drop-based, panic/`?`-safe) is threaded through the four `ElementMapAccess::new` call sites (covering the struct, seq, and enum struct/tuple-variant paths) and the top-level `deserialize_enum`. Exceeding the limit returns the new `DeError::TooDeeplyNested(usize)` instead of aborting.

## Testing

Before/after via standalone release-mode PoCs (built once against pristine master, once against this fix): both genuine cycles abort (exit 134) pre-fix at attacker-controlled depth and return a clean `Err(TooDeeplyNested(128))` post-fix; ordinary shallow documents (depth ≤ 128) are unaffected. Full existing suite (`cargo test --features serialize,overlapped-lists`, and separately without `overlapped-lists`): all passing, including the serialize round-trip suite. `cargo fmt --check` / `cargo clippy --all-targets` clean.

Discovered by the [rust-in-peace](https://github.com/scadastrangelove/rust-in-peace/) security pipeline.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -2128,7 +2128,7 @@ use std::io::BufRead;
 use std::mem::replace;
 #[cfg(feature = "overlapped-lists")]
 use std::num::NonZeroUsize;
-use std::ops::{Deref, Range};
+use std::ops::{Deref, DerefMut, Range};
 
 /// Data represented by a text node or a CDATA node. XML markup is not expected
 pub(crate) const TEXT_KEY: &str = "$text";
@@ -2574,6 +2574,63 @@ where
 
     /// Buffer to store attribute name as a field name exposed to serde consumers
     key_buf: String,
+
+    /// Current recursion depth of nested structs, sequences-of-structs and
+    /// enum variants. Every level of such nesting adds native call-stack
+    /// frames that are not otherwise bounded (see [`Self::max_depth`] and
+    /// <https://github.com/tafia/quick-xml/issues/978>).
+    depth: usize,
+    /// Maximum allowed value of [`Self::depth`] before [`DeError::TooDeeplyNested`]
+    /// is returned. Configurable via [`Self::recursion_limit`].
+    max_depth: usize,
+}
+
+/// Default value of [`Deserializer::max_depth`], matching `serde_json`'s
+/// default recursion limit.
+const DEFAULT_RECURSION_LIMIT: usize = 128;
+
+/// RAII guard that decrements [`Deserializer::depth`] when dropped. Ensures
+/// the recursion-depth counter stays balanced across early returns (`?`) and
+/// panics unwinding through the guarded recursive call, not just the
+/// successful-return path.
+struct DepthGuard<'a, 'de, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    de: &'a mut Deserializer<'de, R, E>,
+}
+
+impl<'a, 'de, R, E> Drop for DepthGuard<'a, 'de, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    fn drop(&mut self) {
+        self.de.depth -= 1;
+    }
+}
+
+impl<'a, 'de, R, E> Deref for DepthGuard<'a, 'de, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    type Target = Deserializer<'de, R, E>;
+
+    fn deref(&self) -> &Self::Target {
+        self.de
+    }
+}
+
+impl<'a, 'de, R, E> DerefMut for DepthGuard<'a, 'de, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.de
+    }
 }
 
 impl<'de, R, E> Deserializer<'de, R, E>
@@ -2602,7 +2659,26 @@ where
             peek: None,
 
             key_buf: String::new(),
+
+            depth: 0,
+            max_depth: DEFAULT_RECURSION_LIMIT,
         }
+    }
+
+    /// Checks and increments the recursion-depth counter, returning a
+    /// [`DepthGuard`] that decrements it again when dropped. Returns
+    /// [`DeError::TooDeeplyNested`] (without incrementing) if the configured
+    /// [`Self::max_depth`] (see [`Self::recursion_limit`]) would be exceeded.
+    ///
+    /// Must be called once per level of struct/sequence/enum nesting that
+    /// re-enters the deserializer, so that sibling (as opposed to nested)
+    /// elements do not spuriously accumulate depth.
+    fn enter_depth(&mut self) -> Result<DepthGuard<'_, 'de, R, E>, DeError> {
+        if self.depth >= self.max_depth {
+            return Err(DeError::TooDeeplyNested(self.max_depth));
+        }
+        self.depth += 1;
+        Ok(DepthGuard { de: self })
     }
 
     /// Returns `true` if all events was consumed.
@@ -2713,6 +2789,46 @@ where
     #[cfg(feature = "overlapped-lists")]
     pub fn event_buffer_size(&mut self, limit: Option<NonZeroUsize>) -> &mut Self {
         self.limit = limit;
+        self
+    }
+
+    /// Set the maximum recursion depth allowed while deserializing nested
+    /// structs, sequences of structs (`Vec<T>` where `T` is a struct), and
+    /// enum variants.
+    ///
+    /// Each level of such nesting re-enters the deserializer and adds native
+    /// call-stack frames that are not otherwise bounded. Deserializing a
+    /// maliciously deeply nested untrusted XML document can overflow the
+    /// native stack, which aborts the process -- this is a [DoS] that cannot
+    /// be caught as a [`Result::Err`], unlike ordinary parse errors. This
+    /// limit turns that abort into a clean [`DeError::TooDeeplyNested`].
+    ///
+    /// Defaults to 128, matching `serde_json`'s default recursion limit.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use quick_xml::de::Deserializer;
+    /// use quick_xml::DeError;
+    /// use serde::Deserialize;
+    ///
+    /// #[derive(Deserialize)]
+    /// struct Node {
+    ///     #[serde(default, rename = "Node")]
+    ///     node: Vec<Node>,
+    /// }
+    ///
+    /// let xml = "<Node>".repeat(1000) + &"</Node>".repeat(1000);
+    /// let mut de = Deserializer::from_str(&xml);
+    /// de.recursion_limit(100);
+    ///
+    /// let result = Node::deserialize(&mut de);
+    /// assert!(matches!(result, Err(DeError::TooDeeplyNested(100))));
+    /// ```
+    ///
+    /// [DoS]: https://en.wikipedia.org/wiki/Denial-of-service_attack
+    pub fn recursion_limit(&mut self, limit: usize) -> &mut Self {
+        self.max_depth = limit;
         self
     }
 
@@ -3230,7 +3346,7 @@ where
         // When document is pretty-printed there could be whitespaces before the root element
         self.skip_whitespaces()?;
         match self.next()? {
-            DeEvent::Start(e) => visitor.visit_map(ElementMapAccess::new(self, e, fields)),
+            DeEvent::Start(e) => visitor.visit_map(ElementMapAccess::new(self, e, fields)?),
             // SAFETY: The reader is guaranteed that we don't have unmatched tags
             // If we here, then our deserializer has a bug
             DeEvent::End(e) => unreachable!("{:?}", e),
@@ -3305,7 +3421,13 @@ where
         // which represents the enum variant
         // Checked by `top_level::list_of_enum` test in serde-de-seq
         self.skip_whitespaces()?;
-        visitor.visit_enum(var::EnumAccess::new(self))
+        // Bounds recursion depth: a newtype variant wrapping the same enum
+        // type (`enum E { Wrap(Box<E>) }`) re-enters this method through
+        // `VariantAccess::newtype_variant_seed`, adding native call-stack
+        // frames per level of XML nesting with no other bound.
+        // See https://github.com/tafia/quick-xml/issues/978
+        let mut guard = self.enter_depth()?;
+        visitor.visit_enum(var::EnumAccess::new(&mut *guard))
     }
 
     fn deserialize_seq<V>(self, visitor: V) -> Result<V::Value, DeError>
--- a/tests/serde-issues.rs
+++ b/tests/serde-issues.rs
@@ -867,3 +867,156 @@ fn issue928() {
         },
     );
 }
+
+/// Regression tests for https://github.com/tafia/quick-xml/issues/978.
+///
+/// The serde `Deserializer` had no recursion-depth cap on any of its
+/// internal recursive re-entry points, so a sufficiently deeply nested
+/// (but otherwise well-formed) XML document would overflow the native
+/// call stack and abort the whole process -- a DoS that, unlike an
+/// ordinary parse error, cannot be caught as a `Result::Err`.
+///
+/// Variant analysis (see the issue comments) found three independent
+/// re-entry points sharing this root cause:
+/// 1. the struct/map cycle described in the issue body itself -- a
+///    `$value` field holding a `Vec` of the same struct
+///    (`deserialize_struct` -> `ElementMapAccess` -> `next_value_seed`);
+/// 2. `deserialize_seq` for an ordinary (non-`$value`) `Vec<Self>` field
+///    (`MapValueSeqAccess::next_element_seed` -> `ElementDeserializer::deserialize_struct`);
+/// 3. enum `newtype_variant` re-entry
+///    (`VariantAccess::newtype_variant_seed` re-entering `Deserializer::deserialize_enum`).
+///
+/// All three are exercised here with a small `recursion_limit()` so the
+/// test stays fast; the default limit (128, matching `serde_json`) is
+/// checked separately in `default_limit_is_128`.
+mod issue978 {
+    use super::*;
+    use quick_xml::de::Deserializer as XmlDeserializer;
+    use quick_xml::DeError;
+
+    /// Cycle 1: a struct whose `$value` field is a `Vec` of itself -- the
+    /// "struct-variant" shape that #819 says works correctly at shallow
+    /// depth, but which (before this fix) had no depth cap at all.
+    #[test]
+    fn cycle1_struct_value_vec_self() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct S {
+            #[serde(default, rename = "$value")]
+            v: Vec<S>,
+        }
+
+        // Within the limit: deserializes successfully, same as before this fix.
+        let xml = "<a>".repeat(5) + &"</a>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(8);
+        assert!(S::deserialize(&mut de).is_ok());
+
+        // Past the limit: a clean, catchable error -- before this fix, a
+        // sufficiently large depth here aborted the process instead
+        // (confirmed on master with depth 20_000).
+        let xml = "<a>".repeat(50) + &"</a>".repeat(50);
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(8);
+        assert!(matches!(
+            S::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(8))
+        ));
+    }
+
+    /// Cycle 2 (from the variant-analysis comment): a plain (non-`$value`)
+    /// `Vec<Self>` field, which recurses through `deserialize_seq` and
+    /// `ElementDeserializer::deserialize_struct` -- a different code path
+    /// than cycle 1, so it needs its own guard.
+    #[test]
+    fn cycle2_deserialize_seq_vec_self() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct Node {
+            #[serde(default, rename = "Node")]
+            node: Vec<Node>,
+        }
+
+        let xml = "<Node>".repeat(5) + &"</Node>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(8);
+        assert!(Node::deserialize(&mut de).is_ok());
+
+        // Before this fix, a sufficiently large depth here aborted the
+        // process instead (confirmed on master with depth 10_000).
+        let xml = "<Node>".repeat(50) + &"</Node>".repeat(50);
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(8);
+        assert!(matches!(
+            Node::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(8))
+        ));
+    }
+
+    /// Cycle 3 (from the variant-analysis comment): an enum newtype variant
+    /// wrapping itself.
+    ///
+    /// Note this shares its root cause with the already-filed #819: because
+    /// `EnumAccess::variant_seed` only *peeks* the `Start` event and
+    /// `VariantAccess::newtype_variant_seed` re-enters `deserialize_enum`
+    /// without consuming it either, the reader position never advances
+    /// across the recursive call. That means this shape does not actually
+    /// need *deep* nesting to blow the stack: it fails to terminate at
+    /// *any* depth, including a single `<Wrap></Wrap>` pair (independently
+    /// confirmed on master: still overflows with `RUST_MIN_STACK=1073741824`,
+    /// i.e. a 1 GiB stack). So this specific cycle is not a distinct,
+    /// attacker-controlled-depth DoS the way cycles 1 and 2 are -- it is
+    /// #819 exhibited through a bigger document. The guard added for this
+    /// issue still converts that non-termination from an uncatchable abort
+    /// into a clean, catchable error, which is why it is kept and tested
+    /// here.
+    #[test]
+    fn cycle3_enum_newtype_variant_self() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum E {
+            Leaf(bool),
+            Wrap(Box<E>),
+        }
+
+        // A shallow, non-recursive variant still deserializes normally.
+        let xml = "<Leaf>true</Leaf>";
+        let mut de = XmlDeserializer::from_str(xml);
+        de.recursion_limit(8);
+        assert!(E::deserialize(&mut de).is_ok());
+
+        // A *single* level of the self-referential newtype variant already
+        // never terminates (see doc comment above) -- before this fix, this
+        // aborted the process even with a 1 GiB stack.
+        let xml = "<Wrap></Wrap>";
+        let mut de = XmlDeserializer::from_str(xml);
+        de.recursion_limit(8);
+        assert!(matches!(
+            E::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(8))
+        ));
+    }
+
+    /// The default recursion limit (128, matching `serde_json`'s default)
+    /// applies even without calling `recursion_limit()` explicitly.
+    #[test]
+    fn default_limit_is_128() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct Node {
+            #[serde(default, rename = "Node")]
+            node: Vec<Node>,
+        }
+
+        let xml = "<Node>".repeat(128) + &"</Node>".repeat(128);
+        let mut de = XmlDeserializer::from_str(&xml);
+        assert!(Node::deserialize(&mut de).is_ok());
+
+        let xml = "<Node>".repeat(1000) + &"</Node>".repeat(1000);
+        let mut de = XmlDeserializer::from_str(&xml);
+        assert!(matches!(
+            Node::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(128))
+        ));
+    }
+}
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,9 @@
 
 ### New Features
 
+- [#978]: Add `Deserializer::recursion_limit` to bound the recursion depth
+  of the serde `Deserializer` (defaults to 128, matching `serde_json`).
+
 ### Bug Fixes
 
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
@@ -24,10 +27,21 @@
   `u16` depth counter. Previously the unguarded `nesting_level += 1` panicked
   under `overflow-checks` builds and silently wrapped in release, corrupting
   namespace-scope bookkeeping on deeply nested untrusted input.
+- [#978]: The serde `Deserializer` now returns the new
+  `DeError::TooDeeplyNested` instead of overflowing the native call stack
+  and aborting the process when deserializing a struct, a `Vec<T>`-shaped
+  sequence of structs, or an enum variant nested deeper than the configured
+  recursion limit. Previously none of `deserialize_struct`, `deserialize_seq`,
+  or `deserialize_enum` bounded the recursion depth of their respective
+  re-entrant `MapAccess`/`SeqAccess`/`EnumAccess`/`VariantAccess`
+  implementations, so a maliciously deeply nested (but otherwise
+  well-formed) untrusted XML document could crash the whole process in a
+  way that could not be caught as a `Result::Err`.
 
 ### Misc Changes
 
 [#977]: https://github.com/tafia/quick-xml/issues/977
+[#978]: https://github.com/tafia/quick-xml/issues/978
 
 
 ## 0.41.0 -- 2026-06-29
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -5,7 +5,7 @@ use crate::{
     de::resolver::EntityResolver,
     de::simple_type::SimpleTypeDeserializer,
     de::text::TextDeserializer,
-    de::{DeEvent, Deserializer, XmlRead, TEXT_KEY, VALUE_KEY},
+    de::{DeEvent, DepthGuard, Deserializer, XmlRead, TEXT_KEY, VALUE_KEY},
     errors::serialize::DeError,
     errors::Error,
     events::attributes::IterState,
@@ -173,7 +173,13 @@ where
 {
     /// Tag -- owner of attributes
     start: BytesStart<'de>,
-    de: &'d mut Deserializer<'de, R, E>,
+    /// A guard that both grants access to the deserializer and bounds the
+    /// recursion depth for as long as this element's map/struct scope is
+    /// alive (dropping it, e.g. when this `ElementMapAccess` is dropped,
+    /// releases one level of the recursion-depth budget). See
+    /// [`Deserializer::recursion_limit`] and
+    /// <https://github.com/tafia/quick-xml/issues/978>.
+    de: DepthGuard<'d, 'de, R, E>,
     /// State of the iterator over attributes. Contains the next position in the
     /// inner `start` slice, from which next attribute should be parsed.
     iter: IterState,
@@ -200,13 +206,21 @@ where
     R: XmlRead<'de>,
     E: EntityResolver,
 {
-    /// Create a new ElementMapAccess
+    /// Create a new ElementMapAccess.
+    ///
+    /// Returns [`DeError::TooDeeplyNested`] instead of constructing the
+    /// accessor if doing so would exceed the deserializer's configured
+    /// recursion-depth limit -- every struct/map scope opened here can, via
+    /// a recursive user type, re-enter this constructor, and each level adds
+    /// unbounded native call-stack frames otherwise (see
+    /// <https://github.com/tafia/quick-xml/issues/978>).
     pub fn new(
         de: &'d mut Deserializer<'de, R, E>,
         start: BytesStart<'de>,
         fields: &'static [&'static str],
-    ) -> Self {
-        Self {
+    ) -> Result<Self, DeError> {
+        let de = de.enter_depth()?;
+        Ok(Self {
             de,
             iter: IterState::new(start.name().as_ref().len(), false),
             start,
@@ -214,7 +228,7 @@ where
             fields,
             has_value_field: fields.contains(&VALUE_KEY),
             has_text_field: fields.contains(&TEXT_KEY),
-        }
+        })
     }
 
     /// Determines if subtree started with the specified event should be skipped.
@@ -800,7 +814,9 @@ where
         V: Visitor<'de>,
     {
         match self.map.de.next()? {
-            DeEvent::Start(e) => visitor.visit_map(ElementMapAccess::new(self.map.de, e, fields)),
+            DeEvent::Start(e) => {
+                visitor.visit_map(ElementMapAccess::new(&mut *self.map.de, e, fields)?)
+            }
             DeEvent::Text(e) => {
                 SimpleTypeDeserializer::from_text_content(e).deserialize_struct("", fields, visitor)
             }
@@ -997,7 +1013,7 @@ where
                     DeEvent::Start(start) => seed
                         .deserialize(ElementDeserializer {
                             start,
-                            de: self.map.de,
+                            de: &mut *self.map.de,
                         })
                         .map(Some),
                     // SAFETY: we just checked that the next event is Start
@@ -1143,7 +1159,7 @@ where
     where
         V: Visitor<'de>,
     {
-        visitor.visit_map(ElementMapAccess::new(self.de, self.start, fields))
+        visitor.visit_map(ElementMapAccess::new(self.de, self.start, fields)?)
     }
 
     fn deserialize_enum<V>(
--- a/src/de/var.rs
+++ b/src/de/var.rs
@@ -132,7 +132,7 @@ where
         V: Visitor<'de>,
     {
         match self.de.next()? {
-            DeEvent::Start(e) => visitor.visit_map(ElementMapAccess::new(self.de, e, fields)),
+            DeEvent::Start(e) => visitor.visit_map(ElementMapAccess::new(self.de, e, fields)?),
             DeEvent::Text(e) => {
                 SimpleTypeDeserializer::from_text_content(e).deserialize_struct("", fields, visitor)
             }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -349,6 +349,18 @@ pub mod serialize {
         /// exceeded. The limit was provided as an argument
         #[cfg(feature = "overlapped-lists")]
         TooManyEvents(NonZeroUsize),
+        /// The document nests structs, sequences-of-structs, or enum variants
+        /// more deeply than the configured recursion limit allows.
+        ///
+        /// Each level of such nesting adds native call-stack frames while
+        /// deserializing (through `MapAccess`/`EnumAccess`/`VariantAccess`
+        /// re-entering the deserializer), which are not otherwise bounded.
+        /// Without this limit, a sufficiently deeply nested untrusted document
+        /// can overflow the native stack, aborting the process in a way that
+        /// cannot be caught as a [`Result::Err`]. The contained value is the
+        /// limit that was exceeded; see
+        /// [`Deserializer::recursion_limit`](crate::de::Deserializer::recursion_limit).
+        TooDeeplyNested(usize),
     }
 
     impl fmt::Display for DeError {
@@ -365,6 +377,11 @@ pub mod serialize {
                 Self::UnexpectedEof => f.write_str("unexpected `Event::Eof`"),
                 #[cfg(feature = "overlapped-lists")]
                 Self::TooManyEvents(s) => write!(f, "deserializer buffered {} events, limit exceeded", s),
+                Self::TooDeeplyNested(limit) => write!(
+                    f,
+                    "input document is nested too deeply, recursion limit ({}) exceeded",
+                    limit
+                ),
             }
         }
     }


src/de/mod.rs
```diff
@@ -2716,6 +2792,46 @@ where
...
+    ///     #[serde(default, rename = "Node")]
+    ///     node: Vec<Node>,
+    /// }
+    ///
+    /// let xml = "<Node>".repeat(1000) + &"</Node>".repeat(1000);
+    /// let mut de = Deserializer::from_str(&xml);
+    /// de.recursion_limit(100);
+    ///
+    /// let result = Node::deserialize(&mut de);
+    /// assert!(matches!(result, Err(DeError::TooDeeplyNested(100))));
```
Mingun, line 2826: I think, that 10 and 11 are enough for demonstration purposes
```suggestion
    /// let xml = "<Node>".repeat(11) + &"</Node>".repeat(11);
    /// let mut de = Deserializer::from_str(&xml);
    ///
    /// de.recursion_limit(11);
    /// let _ = Node::deserialize(&mut de).unwrap();
    ///
    /// de.recursion_limit(10);
    /// let result = Node::deserialize(&mut de);
    /// assert!(matches!(result, Err(DeError::TooDeeplyNested(10))));
```

src/de/mod.rs
```diff
@@ -2574,6 +2574,63 @@ where
...
+
+/// Default value of [`Deserializer::max_depth`], matching `serde_json`'s
+/// default recursion limit.
+const DEFAULT_RECURSION_LIMIT: usize = 128;
+
+/// RAII guard that decrements [`Deserializer::depth`] when dropped. Ensures
+/// the recursion-depth counter stays balanced across early returns (`?`) and
+/// panics unwinding through the guarded recursive call, not just the
+/// successful-return path.
+struct DepthGuard<'a, 'de, R, E>
```
Mingun, line 2596: Is that type necessary? It seems to me that `ElementMapAccess` can fulfill its goals. It is created for each element, so nesting should be possible only via it.

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+        let mut de = XmlDeserializer::from_str(&xml);
+        assert!(Node::deserialize(&mut de).is_ok());
+
+        let xml = "<Node>".repeat(1000) + &"</Node>".repeat(1000);
+        let mut de = XmlDeserializer::from_str(&xml);
+        assert!(matches!(
+            Node::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(128))
+        ));
+    }
```
Mingun, line 1021: It seems to me that this test not needed. Default value is not something that is important, it is just reasonable. Instead, I would like to see the test that demonstrates that sibling elements does not trigger the error.

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+
+        // Within the limit: deserializes successfully, same as before this fix.
+        let xml = "<a>".repeat(5) + &"</a>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(8);
+        assert!(S::deserialize(&mut de).is_ok());
+
+        // Past the limit: a clean, catchable error -- before this fix, a
+        // sufficiently large depth here aborted the process instead
+        // (confirmed on master with depth 20_000).
```
Mingun, line 917: I do not think that it is necessary to confirm in the comments that something was once broken in the master. The very presence of this test is already a sufficient argument
```suggestion
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+    #[test]
+    fn cycle1_struct_value_vec_self() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct S {
+            #[serde(default, rename = "$value")]
+            v: Vec<S>,
+        }
+
+        // Within the limit: deserializes successfully, same as before this fix.
```
Mingun, line 909:
```suggestion
        // Within the limit: deserializes successfully
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+            node: Vec<Node>,
+        }
+
+        let xml = "<Node>".repeat(5) + &"</Node>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(8);
+        assert!(Node::deserialize(&mut de).is_ok());
+
+        // Before this fix, a sufficiently large depth here aborted the
+        // process instead (confirmed on master with depth 10_000).
```
Mingun, line 946:
```suggestion
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+mod issue978 {
+    use super::*;
+    use quick_xml::de::Deserializer as XmlDeserializer;
+    use quick_xml::DeError;
+
+    /// Cycle 1: a struct whose `$value` field is a `Vec` of itself -- the
+    /// "struct-variant" shape that #819 says works correctly at shallow
+    /// depth, but which (before this fix) had no depth cap at all.
+    #[test]
+    fn cycle1_struct_value_vec_self() {
```
Mingun, line 901: Highlight important details in names:
```suggestion
    fn struct_value_field() {
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+            Err(DeError::TooDeeplyNested(8))
+        ));
+    }
+
+    /// Cycle 2 (from the variant-analysis comment): a plain (non-`$value`)
+    /// `Vec<Self>` field, which recurses through `deserialize_seq` and
+    /// `ElementDeserializer::deserialize_struct` -- a different code path
+    /// than cycle 1, so it needs its own guard.
+    #[test]
+    fn cycle2_deserialize_seq_vec_self() {
```
Mingun, line 932:
```suggestion
    fn struct_field() {
```
Also, I would like to move that test before the `struct_value_field`, because this is general case, and `$value` is specific case.

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+    /// *any* depth, including a single `<Wrap></Wrap>` pair (independently
+    /// confirmed on master: still overflows with `RUST_MIN_STACK=1073741824`,
+    /// i.e. a 1 GiB stack). So this specific cycle is not a distinct,
+    /// attacker-controlled-depth DoS the way cycles 1 and 2 are -- it is
+    /// #819 exhibited through a bigger document. The guard added for this
+    /// issue still converts that non-termination from an uncatchable abort
+    /// into a clean, catchable error, which is why it is kept and tested
+    /// here.
+    #[test]
+    fn cycle3_enum_newtype_variant_self() {
```
Mingun, line 974:
```suggestion
    fn newtype_variant() {
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(8);
+        assert!(matches!(
+            Node::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(8))
+        ));
+    }
+
+    /// Cycle 3 (from the variant-analysis comment): an enum newtype variant
+    /// wrapping itself.
```
Mingun, line 957:
```suggestion
    /// Case 3: an enum newtype variant wrapping itself.
```
I also would to see recursion over the other forms of enum, in that order:
- newtype variant (this test)
- tuple variant
- struct variant

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+    /// across the recursive call. That means this shape does not actually
+    /// need *deep* nesting to blow the stack: it fails to terminate at
+    /// *any* depth, including a single `<Wrap></Wrap>` pair (independently
+    /// confirmed on master: still overflows with `RUST_MIN_STACK=1073741824`,
+    /// i.e. a 1 GiB stack). So this specific cycle is not a distinct,
+    /// attacker-controlled-depth DoS the way cycles 1 and 2 are -- it is
+    /// #819 exhibited through a bigger document. The guard added for this
+    /// issue still converts that non-termination from an uncatchable abort
+    /// into a clean, catchable error, which is why it is kept and tested
+    /// here.
```
Mingun, line 972: I don't think that description of the internal implementation is necessary here -- it may change at any time. You only need to keep the note, that currently `<Wrap></Wrap>` is enough to trigger the error because of #819. But not here -- over the check where this input is used. And please mark that comment with FIXME:

```suggestion
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+        assert!(matches!(
+            S::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(8))
+        ));
+    }
+
+    /// Cycle 2 (from the variant-analysis comment): a plain (non-`$value`)
+    /// `Vec<Self>` field, which recurses through `deserialize_seq` and
+    /// `ElementDeserializer::deserialize_struct` -- a different code path
+    /// than cycle 1, so it needs its own guard.
```
Mingun, line 930:
```suggestion
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+/// test stays fast; the default limit (128, matching `serde_json`) is
+/// checked separately in `default_limit_is_128`.
+mod issue978 {
+    use super::*;
+    use quick_xml::de::Deserializer as XmlDeserializer;
+    use quick_xml::DeError;
+
+    /// Cycle 1: a struct whose `$value` field is a `Vec` of itself -- the
+    /// "struct-variant" shape that #819 says works correctly at shallow
+    /// depth, but which (before this fix) had no depth cap at all.
```
Mingun, line 899:
```suggestion
```

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+    #[test]
+    fn cycle1_struct_value_vec_self() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct S {
+            #[serde(default, rename = "$value")]
+            v: Vec<S>,
```
Mingun, line 906: To cover non-seq paths please also add tests with `Box` instead of `Vec`.

tests/serde-issues.rs
```diff
@@ -867,3 +867,156 @@ fn issue928() {
...
+///    `$value` field holding a `Vec` of the same struct
+///    (`deserialize_struct` -> `ElementMapAccess` -> `next_value_seed`);
+/// 2. `deserialize_seq` for an ordinary (non-`$value`) `Vec<Self>` field
+///    (`MapValueSeqAccess::next_element_seed` -> `ElementDeserializer::deserialize_struct`);
+/// 3. enum `newtype_variant` re-entry
+///    (`VariantAccess::newtype_variant_seed` re-entering `Deserializer::deserialize_enum`).
+///
+/// All three are exercised here with a small `recursion_limit()` so the
+/// test stays fast; the default limit (128, matching `serde_json`) is
+/// checked separately in `default_limit_is_128`.
```
Mingun, line 891: No need to refer to comments, actually, you need much more tests than the three described here:
1. struct field, `Box`
2. struct field, `Vec`
3. struct `$value` field, `Box`
4. struct `$value` field, `Vec`
5. enum newtype variant, `Box`
6. enum newtype variant, `Vec`
7. enum tuple variant, `Box`
8. enum tuple variant, `Vec`
9. enum struct variant, field, `Box`
10. enum struct variant, field, `Vec`
11. enum struct variant, `$value` field, `Box`
12. enum struct variant, `$value` field, `Vec`
13. Test that sibling elements does not trigger the threshold

src/de/mod.rs
```diff
@@ -2574,6 +2574,63 @@ where
...
+}
+
+impl<'a, 'de, R, E> DerefMut for DepthGuard<'a, 'de, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.de
+    }
```
Mingun, line 2633: If `DepthGuard` will kept, it seems to me, that it would better to use `de` field directly instead of via traits.
```suggestion
```

Mingun (changes requested):
The test-suite needs to be significantly extended. Also, it seems to me that `DepthGuard` type can be eliminated?

scadastrangelove:
Hi @Mingun, thanks — I went through the matrix empirically rather than assuming my first version was right.

I'm not a real developer, more of a fuzzer/security guy, so I apologize for any inaccuracies. The test draft / analysis below was created with help from an LLM-agent, so please double-check.

I verified this locally against PR head https://github.com/tafia/quick-xml/commit/6d763c18e58b235df3d67621a8edf03e0d5daf74.

The arithmetic in my earlier pass was wrong: I conflated “10 test functions” with your 13 requested cases, but `default_limit_is_128` is not one of the 13. My current read is that the clean green coverage is 9/13:

- 1 struct field, `Box` — addable as a green test
- 2 struct field, `Vec` — already covered by `cycle2`
- 3 struct `$value` field, `Box` — addable as a green test
- 4 struct `$value` field, `Vec` — already covered by `cycle1`
- 5 enum newtype variant, `Box` — already covered by `cycle3`
- 10 enum struct variant, named field, `Vec` — addable as a green test, with the field XML shape matching quick-xml's enum-field conventions
- 11 enum struct variant, `$value` field, `Box` — addable as a green test
- 12 enum struct variant, `$value` field, `Vec` — addable as a green test
- 13 sibling elements do not accumulate depth — addable as a green test

The remaining four do not seem to be clean “shallow succeeds, deep fails with TooDeeplyNested” regression tests as-is:

- 6, enum newtype variant with `Vec`: a scalar leaf works via text content (`<Wrap>true</Wrap>` can deserialize as `Wrap(Vec<bool>)`), but any actual child-element nesting attempt, even a single `<Leaf>` child, hits `TooDeeplyNested` immediately. So this looks like the same enum re-entry / non-consuming-start limitation as the existing `Box` newtype case, not an ordinary recursive-depth case.
- 7/8, enum tuple variants: I confirmed from the existing tests that the tuple-variant convention is repeated sibling tags, e.g. `<Multi>true</Multi><Multi>5</Multi>`. With a recursive tuple slot in the first position (`Multi(Box<E>, u8)` / `Multi(Vec<E>, u8)`), even the shallow scalar-looking form immediately hits `TooDeeplyNested`, because the variant `Start` event is not consumed before re-entering enum deserialization. With the recursive slot later in the tuple (`Multi(bool, Box<E>)` / `Multi(bool, Vec<E>)`), my attempted recursive shapes hit `UnexpectedStart` instead. So these expose enum-shape/re-entry limitations rather than clean green depth-limit tests.
- 9, enum struct variant with a named `Option<Box<Self>>` field: the self-referential shape fails separately (I observed `unknown variant `$text`` on `<Named><Named></Named></Named>`). The analogous `Vec<Self>` and `$value` cases work, so this looks like a real separate deserialization issue, but I would not include it as a passing regression test for this PR.

On `DepthGuard`: your suggestion is mostly right in that the implementation can probably be simplified. `Deref` / `DerefMut` do not seem essential, and the `ElementMapAccess` site could potentially own the decrement in its own `Drop` implementation instead of storing a separate guard.

I do not think the guard concept can disappear entirely, though. There is a second independent guarded site in `Deserializer::deserialize_enum`, before `visitor.visit_enum(...)`. Top-level enum deserialization can re-enter through `variant_seed` + `newtype_variant_seed` / `tuple_variant` before a non-text `Start` event is consumed, so this is not covered by `ElementMapAccess`. Manual decrement after `visit_enum(...)` would balance the normal `Ok` / `Err` path — my earlier “no clean exit point” framing was too strong. The stronger reason to keep an RAII/scope guard there is panic-unwind safety and avoiding future unbalanced-counter regressions.

My proposed PR update would be: add the 9 green cases, do not add ignored tests for 6/7/8/9 unless you prefer explicit documentation of those existing enum-shape limitations.

_Edit(Mingun): removed references to unrelated issues inserted automatically_

Mingun:
> I'm not a real developer, more of a fuzzer/security guy, so I apologize for any inaccuracies. The test draft / analysis below was created with help from an LLM-agent, so please double-check.

If you have experience in fuzzing, then I expect that you understand that writing tests that cover all possible cases are important thing and you understand that integration tests should check software without assuming its internal implementation.

This is a subtle point, but still we should consider `Box` and `Vec` as two different paths, although it may seem that assuming different behavior for different types that used for implementing self-reference is also assumed some implementation details. I can only say that this assumption is justified, since the processing code for a single reference and a list is different. Yes, this is an appeal to implementation details, but they are not too detailed. We are not based on the fact that any specific chains of methods are called, we just know that the serde API requires us to handle lists in a special way, so we must clearly check that we handle them correctly. So we should establish an expected result for all mentioned cases and check that we get it.

> I do not think the guard concept can disappear entirely, though. There is a second independent guarded site in `Deserializer::deserialize_enum`

Ok, `deserialize_enum` is the second point where we may expect increasing recursion depth. Then we may replicate the same logic in `EnumAccess`:
- increase depth before calling `EnumAccess::new`
- decrease depth in `Drop for EnumAccess`

Or at least store `DepthGuard` inside it

scadastrangelove:
Fair point on covering everything — I went through the full matrix and tried
both guard shapes.

Small attribution note: I'm more of a fuzzer/security person than a quick-xml
maintainer, and I used an LLM-assisted local workflow to draft the tests and
refactor prototypes below. I re-ran them locally, but please double-check style
and fit for the crate.

Both diffs below apply independently on PR head 6d763c1, each passes full
`cargo test --features serialize`, and the full suite also stays green with
both applied together.

**Full matrix — all 13, with an asserted expected result, no ignored tests.**
Most land as clean depth tests (round-trip within the limit, `TooDeeplyNested`
past it): 1, 3, 10, 11, 12, 13 as new tests, 2/4/5 already covered by
cycle1/2/3. The self-referential enum re-entry cases (6 `Vec` newtype, 7/8
tuple with the recursive slot first) return `TooDeeplyNested` at any depth — the
same non-consuming re-entry as the existing `Box` newtype case (#819) — so the
asserted result there is "catchable error, not stack overflow," which is what
this PR guarantees.

Two spots assert **pre-existing** limitations that are independent of the depth
cap, so the behavior is pinned rather than skipped:

- Recursive slot second in a tuple variant (`Multi(bool, Box<Self>)` for 7,
  `Multi(bool, Vec<Self>)` for 8) → `UnexpectedStart` before depth is ever
  reached. Checked for both container types on purpose — same result for
  `Box` and `Vec`, so it's a tuple-variant limitation, not `Box`-specific.
- 9 self-referential named `Box` field (`<Named><Named>…`) → an "unknown
  variant `$text`" error. The `Vec` (10) and `$value` (11) analogues work,
  so this looks like a separate enum-shape issue, not this PR.

Flagging those because if the underlying shapes are ever fixed, those
assertions will need updating — they're characterization of current behavior.

**On moving the guard into `EnumAccess`.** I tried both shapes you suggested:

- `Drop for EnumAccess` that decrements — doesn't compile. `variant_seed` returns
  `VariantAccess { de: self.de }`, moving the deserializer ref out, and that
  conflicts with `Drop` needing exclusive access to `*self.de` when `EnumAccess`
  is dropped (`E0713`).
- Storing the `DepthGuard` inside instead — works, full suite green. It needs
  `DepthGuard` to be `pub(crate)`, `EnumAccess`/`VariantAccess` to hold a
  `DepthGuard` rather than `&mut Deserializer`, `EnumAccess::new` to become
  fallible (it guards on construction), and `mut self` on the five
  `EnumAccess`/`VariantAccess` methods (mutating through the guard field needs it).
  No `Drop for EnumAccess` — the existing `DepthGuard::Drop` fires when
  `VariantAccess` drops, after variant processing. Net +17/-18.

One observation either way: the recursion actually happens inside
`VariantAccess` (`newtype_variant_seed` re-enters `deserialize_enum`), so the
guard has to stay alive through variant processing — which the current local in
`deserialize_enum` already does, since it spans the whole `visit_enum`. So the
stored-guard version is the same RAII scope relocated into the type, at the cost
of threading it through `VariantAccess`; the current placement isn't wrong, just
less visually "owned." My preference would be to push the matrix first and treat
the guard relocation as optional cleanup if you prefer that ownership shape.

<details><summary><b>full-13-case-matrix.diff</b> (tests/serde-issues.rs, applies on 6d763c1)</summary>

```diff
diff --git a/tests/serde-issues.rs b/tests/serde-issues.rs
index e7c5cba..7ba3e5e 100644
--- a/tests/serde-issues.rs
+++ b/tests/serde-issues.rs
@@ -1019,4 +1019,275 @@ mod issue978 {
             Err(DeError::TooDeeplyNested(128))
         ));
     }
+
+    // -------------------------------------------------------------------------
+    // Full self-reference matrix (requested in review). `Box` and `Vec` are
+    // treated as separate paths, and every case has an asserted expected result
+    // -- green where the shape round-trips within the limit and returns a clean
+    // `TooDeeplyNested` past it, and an explicit error assertion where the shape
+    // hits a *pre-existing* enum limitation unrelated to this PR's depth cap.
+    //
+    //  #  | container      | field kind        | covered by
+    // ----+----------------+-------------------+------------------------------
+    //  1  | struct         | field, Box        | struct_field_box (below)
+    //  2  | struct         | field, Vec        | cycle2_deserialize_seq_vec_self
+    //  3  | struct         | $value, Box       | struct_value_field_box (below)
+    //  4  | struct         | $value, Vec       | cycle1_struct_value_vec_self
+    //  5  | enum newtype   | Box               | cycle3_enum_newtype_variant_self
+    //  6  | enum newtype   | Vec               | enum_newtype_variant_vec (below)
+    //  7  | enum tuple     | Box               | enum_tuple_variant_box (below)
+    //  8  | enum tuple     | Vec               | enum_tuple_variant_vec (below)
+    //  9  | enum struct    | named field, Box  | enum_struct_variant_field_box (below)
+    // 10  | enum struct    | named field, Vec  | enum_struct_variant_field_vec (below)
+    // 11  | enum struct    | $value, Box       | enum_struct_variant_value_box (below)
+    // 12  | enum struct    | $value, Vec       | enum_struct_variant_value_vec (below)
+    // 13  | siblings must not accumulate depth | siblings_do_not_accumulate_depth
+    // -------------------------------------------------------------------------
+
+    /// #1 struct field, `Box<Self>`: within the limit round-trips, past it a
+    /// clean `TooDeeplyNested`.
+    #[test]
+    fn struct_field_box() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct Node {
+            #[serde(default, rename = "Node")]
+            node: Option<Box<Node>>,
+        }
+        let ok = "<Node>".repeat(5) + &"</Node>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&ok);
+        de.recursion_limit(8);
+        assert!(Node::deserialize(&mut de).is_ok());
+
+        let deep = "<Node>".repeat(50) + &"</Node>".repeat(50);
+        let mut de = XmlDeserializer::from_str(&deep);
+        de.recursion_limit(8);
+        assert!(matches!(Node::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+    }
+
+    /// #3 struct `$value` field, `Box<Self>`.
+    #[test]
+    fn struct_value_field_box() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct S {
+            #[serde(default, rename = "$value")]
+            v: Option<Box<S>>,
+        }
+        let ok = "<a>".repeat(5) + &"</a>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&ok);
+        de.recursion_limit(8);
+        assert!(S::deserialize(&mut de).is_ok());
+
+        let deep = "<a>".repeat(50) + &"</a>".repeat(50);
+        let mut de = XmlDeserializer::from_str(&deep);
+        de.recursion_limit(8);
+        assert!(matches!(S::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+    }
+
+    /// #6 enum newtype variant, `Vec<Self>`. Like the `Box` newtype variant
+    /// (`cycle3`), a self-referential newtype variant re-enters enum
+    /// deserialization without consuming the variant's `Start` event, so it is
+    /// depth-capped at *any* nesting (that non-termination is #819, not this
+    /// PR). The point this PR fixes is that the cap is now a catchable
+    /// `TooDeeplyNested` instead of a stack overflow.
+    #[test]
+    fn enum_newtype_variant_vec() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum E {
+            Leaf(bool),
+            Wrap(Vec<E>),
+        }
+        // A non-recursive variant is unaffected.
+        let mut de = XmlDeserializer::from_str("<Leaf>true</Leaf>");
+        de.recursion_limit(8);
+        assert!(E::deserialize(&mut de).is_ok());
+
+        // The self-referential variant is capped, not overflowed.
+        let mut de = XmlDeserializer::from_str("<Wrap><Leaf>true</Leaf></Wrap>");
+        de.recursion_limit(8);
+        assert!(matches!(E::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+    }
+
+    /// #7 enum tuple variant, `Box<Self>`. Tuple variants use repeated sibling
+    /// tags (`<Multi>..</Multi><Multi>..</Multi>`). Two sub-shapes:
+    /// - recursive slot first: capped with `TooDeeplyNested` (same enum
+    ///   re-entry as the newtype cases), which is what this PR guarantees.
+    /// - recursive slot second: a *pre-existing* shape limitation independent
+    ///   of the depth cap -- it fails to parse with `UnexpectedStart` before
+    ///   depth is ever reached. Asserted explicitly so the behavior is pinned.
+    #[test]
+    fn enum_tuple_variant_box() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum First {
+            Leaf(bool),
+            Multi(Box<First>, u8),
+        }
+        let mut de = XmlDeserializer::from_str("<Multi>true</Multi><Multi>5</Multi>");
+        de.recursion_limit(8);
+        assert!(matches!(First::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum Second {
+            Leaf(bool),
+            Multi(bool, Box<Second>),
+        }
+        let mut de =
+            XmlDeserializer::from_str("<Multi>true</Multi><Multi><Leaf>true</Leaf></Multi>");
+        de.recursion_limit(8);
+        // Pre-existing: this shape does not parse regardless of the depth cap.
+        assert!(matches!(Second::deserialize(&mut de), Err(DeError::UnexpectedStart(_))));
+    }
+
+    /// #8 enum tuple variant, `Vec<Self>`. Same slot-first/slot-second split as
+    /// #7: recursive slot first is depth-capped; recursive slot second hits the
+    /// same *pre-existing*, non-depth-related `UnexpectedStart` as the `Box`
+    /// case -- confirming this is a tuple-variant limitation shared by both
+    /// container types, not something specific to `Box`.
+    #[test]
+    fn enum_tuple_variant_vec() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum First {
+            Leaf(bool),
+            Multi(Vec<First>, u8),
+        }
+        let mut de = XmlDeserializer::from_str("<Multi>true</Multi><Multi>5</Multi>");
+        de.recursion_limit(8);
+        assert!(matches!(First::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum Second {
+            Leaf(bool),
+            Multi(bool, Vec<Second>),
+        }
+        let mut de =
+            XmlDeserializer::from_str("<Multi>true</Multi><Multi><Leaf>true</Leaf></Multi>");
+        de.recursion_limit(8);
+        // Pre-existing: same shape limitation as the `Box` case, not depth-related.
+        assert!(matches!(Second::deserialize(&mut de), Err(DeError::UnexpectedStart(_))));
+    }
+
+    /// #9 enum struct variant, named `Box<Self>` field. The non-recursive shape
+    /// deserializes; the self-referential shape hits a *pre-existing* issue
+    /// (`unknown variant `$text``) that is independent of the depth cap -- the
+    /// analogous `Vec` (#10) and `$value` (#11) forms work. Asserted explicitly
+    /// so the current behavior is pinned rather than silently skipped.
+    #[test]
+    fn enum_struct_variant_field_box() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum E {
+            Leaf(bool),
+            Named {
+                #[serde(default, rename = "Named")]
+                node: Option<Box<E>>,
+            },
+        }
+        // Non-recursive shape works.
+        let mut de = XmlDeserializer::from_str("<Named><Leaf>true</Leaf></Named>");
+        de.recursion_limit(8);
+        assert!(E::deserialize(&mut de).is_ok());
+
+        // Self-referential shape: pre-existing failure, not the depth cap.
+        let mut de = XmlDeserializer::from_str("<Named><Named></Named></Named>");
+        de.recursion_limit(8);
+        assert!(matches!(E::deserialize(&mut de), Err(DeError::Custom(_))));
+    }
+
+    /// #10 enum struct variant, named `Vec<Self>` field.
+    #[test]
+    fn enum_struct_variant_field_vec() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum E {
+            Leaf(bool),
+            Named {
+                #[serde(default, rename = "Named")]
+                node: Vec<E>,
+            },
+        }
+        let ok = "<Named>".repeat(5) + &"</Named>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&ok);
+        de.recursion_limit(8);
+        assert!(E::deserialize(&mut de).is_ok());
+
+        let deep = "<Named>".repeat(50) + &"</Named>".repeat(50);
+        let mut de = XmlDeserializer::from_str(&deep);
+        de.recursion_limit(8);
+        assert!(matches!(E::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+    }
+
+    /// #11 enum struct variant, `$value` field, `Box<Self>`.
+    #[test]
+    fn enum_struct_variant_value_box() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum E {
+            Leaf(bool),
+            Named {
+                #[serde(default, rename = "$value")]
+                v: Option<Box<E>>,
+            },
+        }
+        let ok = "<Named>".repeat(5) + &"</Named>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&ok);
+        de.recursion_limit(8);
+        assert!(E::deserialize(&mut de).is_ok());
+
+        let deep = "<Named>".repeat(50) + &"</Named>".repeat(50);
+        let mut de = XmlDeserializer::from_str(&deep);
+        de.recursion_limit(8);
+        assert!(matches!(E::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+    }
+
+    /// #12 enum struct variant, `$value` field, `Vec<Self>`.
+    #[test]
+    fn enum_struct_variant_value_vec() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        enum E {
+            Leaf(bool),
+            Named {
+                #[serde(default, rename = "$value")]
+                v: Vec<E>,
+            },
+        }
+        let ok = "<Named>".repeat(5) + &"</Named>".repeat(5);
+        let mut de = XmlDeserializer::from_str(&ok);
+        de.recursion_limit(8);
+        assert!(E::deserialize(&mut de).is_ok());
+
+        let deep = "<Named>".repeat(50) + &"</Named>".repeat(50);
+        let mut de = XmlDeserializer::from_str(&deep);
+        de.recursion_limit(8);
+        assert!(matches!(E::deserialize(&mut de), Err(DeError::TooDeeplyNested(8))));
+    }
+
+    /// #13 sibling (non-nested) elements must not accumulate depth: 200 siblings
+    /// at one level deserialize fine under a small limit, proving the guard
+    /// counts nesting, not elements-seen.
+    #[test]
+    fn siblings_do_not_accumulate_depth() {
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct Leaf {
+            #[serde(default)]
+            _x: (),
+        }
+        #[derive(Debug, Deserialize)]
+        #[allow(dead_code)]
+        struct Root {
+            #[serde(default, rename = "Leaf")]
+            leaves: Vec<Leaf>,
+        }
+        let xml = format!("<Root>{}</Root>", "<Leaf/>".repeat(200));
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(4);
+        assert!(Root::deserialize(&mut de).is_ok());
+    }
 }
```
</details>

<details><summary><b>option-b-enumaccess-guard.diff</b> (src/de/{mod,var}.rs, applies on 6d763c1)</summary>

```diff
diff --git a/src/de/mod.rs b/src/de/mod.rs
index 7c39de7..83d7fbf 100644
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -2593,12 +2593,12 @@ const DEFAULT_RECURSION_LIMIT: usize = 128;
 /// the recursion-depth counter stays balanced across early returns (`?`) and
 /// panics unwinding through the guarded recursive call, not just the
 /// successful-return path.
-struct DepthGuard<'a, 'de, R, E>
+pub(crate) struct DepthGuard<'a, 'de, R, E>
 where
     R: XmlRead<'de>,
     E: EntityResolver,
 {
-    de: &'a mut Deserializer<'de, R, E>,
+    pub(crate) de: &'a mut Deserializer<'de, R, E>,
 }
 
 impl<'a, 'de, R, E> Drop for DepthGuard<'a, 'de, R, E>
@@ -2673,7 +2673,7 @@ where
     /// Must be called once per level of struct/sequence/enum nesting that
     /// re-enters the deserializer, so that sibling (as opposed to nested)
     /// elements do not spuriously accumulate depth.
-    fn enter_depth(&mut self) -> Result<DepthGuard<'_, 'de, R, E>, DeError> {
+    pub(crate) fn enter_depth(&mut self) -> Result<DepthGuard<'_, 'de, R, E>, DeError> {
         if self.depth >= self.max_depth {
             return Err(DeError::TooDeeplyNested(self.max_depth));
         }
@@ -3426,8 +3426,7 @@ where
         // `VariantAccess::newtype_variant_seed`, adding native call-stack
         // frames per level of XML nesting with no other bound.
         // See https://github.com/tafia/quick-xml/issues/978
-        let mut guard = self.enter_depth()?;
-        visitor.visit_enum(var::EnumAccess::new(&mut *guard))
+        visitor.visit_enum(var::EnumAccess::new(self)?)
     }
 
     fn deserialize_seq<V>(self, visitor: V) -> Result<V::Value, DeError>
diff --git a/src/de/var.rs b/src/de/var.rs
index b554412..df4334c 100644
--- a/src/de/var.rs
+++ b/src/de/var.rs
@@ -3,7 +3,7 @@ use crate::{
     de::map::ElementMapAccess,
     de::resolver::EntityResolver,
     de::simple_type::SimpleTypeDeserializer,
-    de::{DeEvent, Deserializer, XmlRead, TEXT_KEY},
+    de::{DeEvent, DepthGuard, Deserializer, XmlRead, TEXT_KEY},
     errors::serialize::DeError,
 };
 use serde::de::value::BorrowedStrDeserializer;
@@ -15,7 +15,7 @@ where
     R: XmlRead<'de>,
     E: EntityResolver,
 {
-    de: &'d mut Deserializer<'de, R, E>,
+    de: DepthGuard<'d, 'de, R, E>,
 }
 
 impl<'de, 'd, R, E> EnumAccess<'de, 'd, R, E>
@@ -23,8 +23,8 @@ where
     R: XmlRead<'de>,
     E: EntityResolver,
 {
-    pub fn new(de: &'d mut Deserializer<'de, R, E>) -> Self {
-        EnumAccess { de }
+    pub fn new(de: &'d mut Deserializer<'de, R, E>) -> Result<Self, DeError> {
+        Ok(EnumAccess { de: de.enter_depth()? })
     }
 }
 
@@ -36,7 +36,7 @@ where
     type Error = DeError;
     type Variant = VariantAccess<'de, 'd, R, E>;
 
-    fn variant_seed<V>(self, seed: V) -> Result<(V::Value, Self::Variant), Self::Error>
+    fn variant_seed<V>(mut self, seed: V) -> Result<(V::Value, Self::Variant), Self::Error>
     where
         V: DeserializeSeed<'de>,
     {
@@ -66,7 +66,7 @@ where
     R: XmlRead<'de>,
     E: EntityResolver,
 {
-    de: &'d mut Deserializer<'de, R, E>,
+    de: DepthGuard<'d, 'de, R, E>,
     /// `true` if variant should be deserialized from a textual content
     /// and `false` if from tag
     is_text: bool,
@@ -79,7 +79,7 @@ where
 {
     type Error = DeError;
 
-    fn unit_variant(self) -> Result<(), Self::Error> {
+    fn unit_variant(mut self) -> Result<(), Self::Error> {
         match self.de.next()? {
             // Consume subtree
             DeEvent::Start(e) => self.de.read_to_end(e.name()),
@@ -91,7 +91,7 @@ where
         }
     }
 
-    fn newtype_variant_seed<T>(self, seed: T) -> Result<T::Value, Self::Error>
+    fn newtype_variant_seed<T>(mut self, seed: T) -> Result<T::Value, Self::Error>
     where
         T: DeserializeSeed<'de>,
     {
@@ -102,11 +102,11 @@ where
                 _ => unreachable!("Only `Text` events are possible here"),
             }
         } else {
-            seed.deserialize(self.de)
+            seed.deserialize(&mut *self.de)
         }
     }
 
-    fn tuple_variant<V>(self, len: usize, visitor: V) -> Result<V::Value, Self::Error>
+    fn tuple_variant<V>(mut self, len: usize, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: Visitor<'de>,
     {
@@ -119,12 +119,12 @@ where
                 _ => unreachable!("Only `Text` events are possible here"),
             }
         } else {
-            self.de.deserialize_tuple(len, visitor)
+            (&mut *self.de).deserialize_tuple(len, visitor)
         }
     }
 
     fn struct_variant<V>(
-        self,
+        mut self,
         fields: &'static [&'static str],
         visitor: V,
     ) -> Result<V::Value, Self::Error>
@@ -132,7 +132,7 @@ where
         V: Visitor<'de>,
     {
         match self.de.next()? {
-            DeEvent::Start(e) => visitor.visit_map(ElementMapAccess::new(self.de, e, fields)?),
+            DeEvent::Start(e) => visitor.visit_map(ElementMapAccess::new(&mut *self.de, e, fields)?),
             DeEvent::Text(e) => {
                 SimpleTypeDeserializer::from_text_content(e).deserialize_struct("", fields, visitor)
             }
```
</details>

dralley:
@scadastrangelove FYI, we've just merged a new contribution policy, please review it.  Would you mind responding to the comments @Mingun made?  Specifically you, not the agent.

It's very challenging to review a PR when all responses are a big block of text from an LLM detached from the original context.

scadastrangelove:
Hi!
Sorry for the "neuroslop".

I'm not claiming that this particular PR should be merged. The main thing is an issue, a PR more like a demonstration of a possible fix.

I'd be happy to delve deeper into the project, but I'm currently managing about 50 issues across various projects, and without the help of a LLM, it's simply overwhelming. Essentially, my main motivation is to improve the security of Rust's infrastructure and develop a tool for community, not to "fix all the things". 

As I wrote above, I'm more of a security guy and can understand the shortcomings, create PoC, and confirm that the issue is real and reproducible. But of course, a reliable fix is ​​a whole other level of complexity.

And here I have a question for you, as maintainers. If I simply file an issue without trying to create a fix and a PR, would that be more acceptable and less annoying? And should I simply include the fix options in the issue?

On the other hand, a number of PRs were leaked by maintainers, and I think this is a sign that they're useful. So SNR is above 0.  Examples:
https://github.com/scadastrangelove/rust-in-peace/blob/main/DISCLOSURES-PUBLIC.md

| | |
|---|---:|
| Reports filed | 57 |
| Resolved (fixed / merged) | 16 |
| Open — awaiting vendor action | 23 |
| Closed — disputed, not a vulnerability, or declined | 7 |
| Private advisories pending vendor publication | 11 |

Basically, I need your opinion on the correct course of action. I'm physically unable to handwrite hundreds of issues-PRs on rice paper. Can review/check reproduction/submit. 
If repo policy prohibits "LLM-submission", I can simply disclose them in the project, but that doesn't align with responsible disclosure practices and what I've been accustomed to doing for 20 years, since my first CVE-2005.

Should I close the issue/PR?
Should I stop spamming maintainers and just post Open Source Security list monthly digests esp it is not GHSA?
What should I do if the GHSA send is not activated by the maintainer?

I need your advice. 

WBR.

scadastrangelove:
AI-assisted contributions policy violation.

dralley:
For a low severity issue like this, if you don't have the time required to actually commit to understanding the fix and the PR, it's fine to just report IMO.  Perhaps with a suggested plan of action as a followup comment on the issue.

PR #985, opened 2026-07-26
Add contribution guidelines, including an AI use policy

dralley:
And an AGENTS.md / CLAUDE.md, so that it gets picked up by tooling.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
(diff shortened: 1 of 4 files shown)


CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
dralley, line 1: Claude, for whatever reason is the only harness which doesn't support AGENTS.md.  Annoying.

This is a symlink

Mingun (approved):

PR #986, opened 2026-07-26
Skip empty text events after trim_text_end

### LINKED ISSUE #984: `trim_text_end` still emits an empty `Text` event for whitespace-only text before a tag
```rust
use quick_xml::events::Event;

fn main() {

    loop {
        match reader.read_event() {
            Ok(Event::Eof) => break,
            Ok(e) => println!("{:?}", e),
            Err(e) => {
                println!("Error: {:?}", e);
                break;
            }
        }
    }
}
```

Output:

```
```

Tested on quick-xml 0.41.0.

JerryNee:
## Summary

- skip whitespace-only `Text` events when `trim_text_end` removes their entire contents before markup
- apply the same behavior before general references, across borrowed, buffered, and Tokio readers
- preserve `read_to_end` spans while text trimming is enabled
- enable the existing end-trimming regression test and add focused cross-reader coverage

## Root cause

`ReaderState::emit_text` trimmed text only after `XmlSource::read_text` had returned a borrowed text slice. If the slice became empty, the parser had already surrendered the reusable buffer and could not safely continue to the following markup or reference, so it returned `Event::Text("")`.

This changes the private `XmlSource::read_text` contract to detect whitespace-only runs when `trim_text_end` is enabled. In that case, the source returns its buffer through the existing `Markup` or `Ref` path, allowing the state machine to continue without emitting an event. Non-empty text continues through the existing trimming path unchanged.

Skipping the empty event also exposed an existing `read_to_end` boundary dependency: its span calculation used the event immediately before an end tag to locate the tag's start. `read_to_end` now temporarily disables both text-trimming options while it computes the raw source span, restoring them on every exit path.

## Validation

- `cargo +1.79.0 check`
- `cargo test --all-features --benches --tests`
- `cargo test --no-default-features`
- `cargo test --features serialize`
- `cargo test --features serialize,encoding`
- `cargo test --features serialize,escape-html`
- `cargo test --all-features`
- `RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps`
- `cargo clippy --all-features --all-targets` (completed with pre-existing warnings only)

Closes #984

AI tooling assisted with implementation and test generation. I reviewed and approved the final patch and validation results before submission.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,8 @@
 
 ### Bug Fixes
 
+- [#984]: Readers no longer emit empty `Text` events when `trim_text_end`
+  removes whitespace-only text before markup or a general reference.
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
   event) now returns the new `NamespaceError::TooDeeplyNested` when a document
   nests elements deeper than `u16::MAX`, instead of overflowing the internal
@@ -29,6 +31,7 @@
 
 - [#983]: Adopted an AI use and contribution policy for new upstream contributions.
 
+[#984]: https://github.com/tafia/quick-xml/issues/984
 [#977]: https://github.com/tafia/quick-xml/issues/977
 [#983]: https://github.com/tafia/quick-xml/issues/983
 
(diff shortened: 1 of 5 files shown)


dralley:
Will fix this up myself

PR #987, opened 2026-07-29
Cap total in-scope namespace bindings

### LINKED ISSUE #980: NamespaceResolver::resolve_prefix linearly scans the entire binding stack per lookup — O(depth²) CPU DoS on deeply nested XML (sibling of RUSTSEC-2026-0194)

dralley:
Replace the per-element `max_declarations_per_element` limit with a total in-scope `max_namespace_bindings` limit (default 128). Documents in excess of this limit are *likely* maliciously constructed, and bounding the stack scanned by `resolve_prefix` prevents this vector from being used to cause DDoS via CPU exhaustion.

The check moves from `push()` (which counted declarations per element) into `add()` (which checks `bindings.len()` against the total limit), thereby catching both single-element and cross-element accumulation.

Renames: 
TooManyDeclarations -> TooManyBindings,
max_declarations_per_element -> max_namespace_bindings,
DEFAULT_MAX_DECLARATIONS_PER_ELEMENT -> DEFAULT_MAX_NAMESPACE_BINDINGS.

closes #980

### DIFF
--- a/src/name.rs
+++ b/src/name.rs
@@ -38,13 +38,14 @@ pub enum NamespaceError {
     ///
     /// Contains the prefix that is tried to be bound.
     InvalidPrefixForXmlns(String),
-    /// A single start tag declared more `xmlns` / `xmlns:*` namespace bindings
-    /// than the configured [`NamespaceResolver::max_declarations_per_element`]
-    /// limit. Contains the configured limit.
+    /// The total number of `xmlns` / `xmlns:*` namespace bindings in scope exceeded
+    /// the configured [`NamespaceResolver::max_namespace_bindings`] limit. Contains
+    /// the configured limit.
     ///
-    /// This bounds the heap allocated by [`NamespaceResolver::push`] (and hence
-    /// by [`NsReader`](crate::reader::NsReader)) on untrusted input.
-    TooManyDeclarations(usize),
+    /// This bounds the work done by [`NamespaceResolver`] (and hence by [`NsReader`](crate::reader::NsReader))
+    /// on untrusted input by capping both the heap allocated and the cost of prefix
+    /// resolution (which scans the binding stack).
+    TooManyBindings(usize),
     /// The document nested elements more deeply than the namespace resolver's
     /// depth counter (a `u16`) can track. This bounds stack / scope-bookkeeping
     /// work on untrusted input. Contains the depth limit that was exceeded.
@@ -81,11 +82,11 @@ impl fmt::Display for NamespaceError {
                     prefix
                 )
             }
-            Self::TooManyDeclarations(limit) => {
+            Self::TooManyBindings(limit) => {
                 write!(
                     f,
-                    "start tag declares more than {} namespace bindings; \
-                     raise the limit with NamespaceResolver::set_max_declarations_per_element",
+                    "more than {} namespace bindings in scope; \
+                     raise the limit with NamespaceResolver::set_max_namespace_bindings",
                     limit,
                 )
             }
@@ -346,10 +347,10 @@ pub struct Namespace<'a>(pub &'a str);
 impl<'a> Namespace<'a> {
     /// Converts this namespace to an internal slice representation.
     ///
-    /// This is [non-normalized] attribute value, i.e. any entity references is
-    /// not expanded and space characters are not removed. This means, that
-    /// different string slices, returned from this method, can represent the same
-    /// namespace and would be treated by parser as identical.
+    /// This is [non-normalized] attribute value, i.e. any entity references is not
+    /// expanded and space characters are not removed. This means, that different
+    /// string slices, returned from this method, can represent the same namespace
+    /// and would be treated by parser as identical.
     ///
     /// For example, if the entity **eacute** has been defined to be **é**,
     /// the empty tags below all contain namespace declarations binding the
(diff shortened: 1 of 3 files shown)


dralley:
Replaces https://github.com/tafia/quick-xml/pull/981

src/name.rs
```diff
@@ -519,23 +520,29 @@ pub struct NamespaceResolver {
...
-/// document while bounding the heap allocated for one `<... xmlns:...>` tag to
-/// a few kilobytes regardless of input size.
-pub const DEFAULT_MAX_DECLARATIONS_PER_ELEMENT: usize = 256;
-
-/// That constant define the one of [reserved namespaces] for the xml standard.
+/// Real-world XML dialects (XHTML, SVG, SOAP, RSS, RRDP, ...) declare a handful of namespaces,
+/// almost always on the root element; 128 is significantly more than what most legitimate documents
+/// would declare, while bounding both the heap allocated and the cost of prefix resolution
+/// (which scans the binding stack).
+pub const DEFAULT_MAX_NAMESPACE_BINDINGS: usize = 128;
```
dralley, line 537: Note: default reduced from 256 to 128.  No particular reason for this other than that it felt wildly excessive, and this is still fairly high.  If there's a compelling argument otherwise though, I'll change it back.

src/name.rs
```diff
@@ -673,15 +688,22 @@ impl NamespaceResolver {
...
             PrefixDeclaration::Named(prefix) => {
-                // error, non-`xml` prefix set to xml uri
                 if namespace == RESERVED_NAMESPACE_XML.1 {
+                    // error, non-`xml` prefix set to xml uri
                     return Err(NamespaceError::InvalidPrefixForXml(prefix.to_string()));
-                } else
-                // error, non-`xmlns` prefix set to xmlns uri
-                if namespace == RESERVED_NAMESPACE_XMLNS.1 {
+                } else if namespace == RESERVED_NAMESPACE_XMLNS.1 {
+                    // error, non-`xmlns` prefix set to xmlns uri
```
Mingun, line 695: Originally, I placed the comment on the line before `if` to keep the nice formatting of `else if` statement. Don't know why `rustfmt` does not have the ability to keep `else` and `if` on different lines to keep the indentation of conditions the same.

Mingun (approved):

PR #988, opened 2026-07-29
Add recursion-depth limit to serde Deserializer

### LINKED ISSUE #978: serde Deserializer has no recursion-depth limit → stack-overflow DoS on deeply nested untrusted XML (distinct from #819)

dralley:
Deeply nested XML could overflow the native call stack via uncatchable abort. Add a configurable depth limit (default 128, matching serde_json) that returns DeError::TooDeeplyNested instead.

Depth is tracked directly on the Deserializer

- ElementMapAccess increments in new(), decrements in Drop (RAII)
- deserialize_enum uses scope-based increment/decrement around visit_enum (EnumAccess can't use Drop due to variant_seed move)

closes #978

### DIFF
--- a/tests/serde-issues.rs
+++ b/tests/serde-issues.rs
@@ -871,3 +871,370 @@ fn issue928() {
         },
     );
 }
+
+/// Regression tests for https://github.com/tafia/quick-xml/issues/978.
+///
+/// Deeply nested XML should produce `DeError::TooDeeplyNested` instead of
+/// a stack overflow. Tests are ordered: struct (general, then `$value`),
+/// enum (newtype, tuple, struct variant), then siblings.
+mod issue978 {
+    use quick_xml::de::Deserializer as XmlDeserializer;
+    use quick_xml::DeError;
+    use serde::Deserialize;
+
+    /// Generate XML like `<S><field><field>...</field></field></S>`.
+    ///
+    /// In quick-xml's serde, the field element IS the container for nested
+    /// structs (there is no inner `<S>` wrapper).
+    fn nested_struct_field(depth: usize) -> String {
+        let mut xml = String::from("<S>");
+        for _ in 0..depth.saturating_sub(1) {
+            xml.push_str("<field>");
+        }
+        for _ in 0..depth.saturating_sub(1) {
+            xml.push_str("</field>");
+        }
+        xml.push_str("</S>");
+        xml
+    }
+
+    /// Generate XML like `<S><S>...<S></S>...</S></S>`.
+    fn nested_struct_value(depth: usize) -> String {
+        let mut xml = String::new();
+        for _ in 0..depth {
+            xml.push_str("<S>");
+        }
+        for _ in 0..depth {
+            xml.push_str("</S>");
+        }
+        xml
+    }
+
+    /// Generate XML like `<Wrap><Wrap>...<Leaf/>...</Wrap></Wrap>`.
+    fn nested_enum_value(depth: usize) -> String {
+        let mut xml = String::new();
+        for _ in 0..depth.saturating_sub(1) {
+            xml.push_str("<Wrap>");
+        }
+        xml.push_str("<Leaf/>");
+        for _ in 0..depth.saturating_sub(1) {
+            xml.push_str("</Wrap>");
+        }
+        xml
+    }
+
+    // struct field
+
+    #[test]
+    fn struct_field_box() {
+        #[derive(Debug, Deserialize)]
+        struct S {
+            field: Option<Box<S>>,
+        }
+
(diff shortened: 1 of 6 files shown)


dralley:
Replaces https://github.com/tafia/quick-xml/pull/982

tests/serde-issues.rs
```diff
@@ -871,3 +871,372 @@ fn issue928() {
...
+        enum E {
+            Wrap {
+                #[serde(rename = "$value")]
+                v: Box<E>,
+            },
+            Leaf,
+        }
+
+        let limit = 3;
+        let nesting = 3;
```
Mingun: Why the separate `nesting` variable? You can reuse `limit` as in other tests.

tests/serde-issues.rs
```diff
@@ -871,3 +871,372 @@ fn issue928() {
...
+            Wrap {
+                #[serde(rename = "$value")]
+                #[serde(default)]
+                v: Vec<E>,
+            },
+            Leaf,
+        }
+
+        let limit = 3;
+        let nesting = 3;
```
Mingun: Same here

Mingun (approved):
Only two small remarks in tests, everything else is good 👍

tests/serde-issues.rs
dralley: fixed

dralley:
@Mingun I'll wait a couple more days and then try to backport all of these patches to 0.41 collectively.

Mingun:
If our next version would be 0.42, than that is not possible. Those patches changes our `Error` / `DeError` definitions, which is a breaking change that cannot be in a patch. So, because we anyway need to increase at least minor version due to other breaking changes, just release a new version as usually. That's not a problem. At least [Ruffle](https://github.com/ruffle-rs/ruffle) project matches on our errors to translate them into Flash errors (because some Flash applications are sensible for that).

PR #992, opened 2026-07-30
Fix attribute iteration past buffer end

### LINKED ISSUE #989: `Attributes::new` panics instead of returning `None` when `pos > buf.len()`
### **Description**:

`quick_xml::events::attributes::Attributes::new(buf, pos)` documents that if `pos` is greater than the buffer length, the iterator should immediately return `None`. However, when such an `Attributes` iterator is consumed, either directly via `next()` or indirectly via `has_nil()`, it panics due to an out-of-bounds slice access.

### **Environment**:

quick-xml version: 0.41.0
rustc version: 1.94.1
OS: Ubuntu 20.04.6 LTS
Command: RUST_BACKTRACE=full cargo run

### PoC

```rust
use quick_xml::events::attributes::Attributes;

fn main() {
    let mut attrs = Attributes::new("a", 2);
    let reader = NsReader::from_str("");
    let _ = attrs.has_nil(reader.resolver());
}
```

### **Expected behavior:**

Since `pos` is greater than `buf.len()`, the iterator should terminate immediately.

### **Actual behavior**

```
thread 'main' panicked at /Rust-Lib-Testing/test/tests/crates/quick-xml-0.41.0/src/events/attributes.rs:1224:49:
range start index 2 out of range for slice of length 1
stack backtrace:
   0:     0x55fe02a6e428 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h0f9d2e8e73f1bc7d
   1:     0x55fe02aad033 - core::fmt::write::h2b15537d64ec9505
   2:     0x55fe02a99f5f - std::io::Write::write_fmt::h7f766131a16fe05a
   3:     0x55fe02a6e273 - std::sys::backtrace::BacktraceLock::print::h5e38585c963961db
   4:     0x55fe02a73f7c - std::panicking::default_hook::{{closure}}::h7a475a2b7b7e4121
   5:     0x55fe02a73e1d - std::panicking::default_hook::hac0f4b95d0cecbc2
   6:     0x55fe02a74451 - std::panicking::rust_panic_with_hook::hb51ba0f67890aa28
   7:     0x55fe02a6e80a - std::panicking::begin_panic_handler::{{closure}}::h73d0277a9a02463b
   8:     0x55fe02a6e649 - std::sys::backtrace::__rust_end_short_backtrace::hf93dae6c307b4fe8
   9:     0x55fe02a7406d - __rustc[b6f6af4e11d2f413]::rust_begin_unwind
  10:     0x55fe02aac0a0 - core::panicking::panic_fmt::h3075732948299e49
  11:     0x55fe02aaa597 - core::slice::index::slice_start_index_len_fail::do_panic::runtime::hcc2a5aadc9262237
  12:     0x55fe02aaa4ca - core::slice::index::slice_start_index_len_fail::hc732b73cbbe828f6
  13:     0x55fe02a6accd - <core::ops::range::RangeFrom<usize> as core::slice::index::SliceIndex<[T]>>::index::he33b10440abe67b2
                               at /Rust-Lib-Testing/test/src/rust/library/core/src/slice/index.rs:570:13
  14:     0x55fe02a5ffa7 - core::slice::index::<impl core::ops::index::Index<I> for [T]>::index::h737ba69d80e4a2ac
                               at /Rust-Lib-Testing/test/src/rust/library/core/src/slice/index.rs:18:15
  15:     0x55fe02a5ffa7 - quick_xml::events::attributes::IterState::next::h1cab26926d07055b
                               at /Rust-Lib-Testing/test/tests/crates/quick-xml-0.41.0/src/events/attributes.rs:1224:49
  16:     0x55fe02a5efb0 - <quick_xml::events::attributes::Attributes as core::iter::traits::iterator::Iterator>::next::h19869966ebbf405f
                               at /Rust-Lib-Testing/test/tests/crates/quick-xml-0.41.0/src/events/attributes.rs:698:26
  17:     0x55fe02a63440 - core::iter::traits::iterator::Iterator::try_fold::hf926e2a9c4d7363f
                               at /Rust-Lib-Testing/test/src/rust/library/core/src/iter/traits/iterator.rs:2425:34
  18:     0x55fe02a633e4 - core::iter::traits::iterator::Iterator::any::h46c8a83f373b4437
                               at /Rust-Lib-Testing/test/src/rust/library/core/src/iter/traits/iterator.rs:2826:14
  19:     0x55fe02a5ef74 - quick_xml::events::attributes::Attributes::has_nil::h129362b91d57b1f7
                               at /Rust-Lib-Testing/test/tests/crates/quick-xml-0.41.0/src/events/attributes.rs:651:14
  20:     0x55fe02a56b5d - replay_quick_xml1::main::h26e7f01651a50c7a
                               at /Rust-Lib-Testing/replay_files/quick-xml-0.41.0/replay_quick-xml1/src/main.rs:7:19
  21:     0x55fe02a56d0b - core::ops::function::FnOnce::call_once::h76f46a5897c6fb29
                               at /Rust-Lib-Testing/test/src/rust/library/core/src/ops/function.rs:253:5
  22:     0x55fe02a56bce - std::sys::backtrace::__rust_begin_short_backtrace::h65002fbaeac49652
                               at /Rust-Lib-Testing/test/src/rust/library/std/src/sys/backtrace.rs:158:18
  23:     0x55fe02a56c31 - std::rt::lang_start::{{closure}}::h69d435d19a4676c4
                               at /Rust-Lib-Testing/test/src/rust/library/std/src/rt.rs:206:18
  24:     0x55fe02a99155 - std::rt::lang_start_internal::h9de2b5630c7cec4d
  25:     0x55fe02a56c17 - std::rt::lang_start::h830c9996d95df1a2
                               at /Rust-Lib-Testing/test/src/rust/library/std/src/rt.rs:205:5
  26:     0x55fe02a56bbe - main
  27:     0x7f674eaa6083 - __libc_start_main
                               at /build/glibc-B3wQXB/glibc-2.31/csu/../csu/libc-start.c:308:16
  28:     0x55fe02a569ee - _start
  29:                0x0 - <unknown>
```

### **Root cause**:

The panic appears to happen in `src/events/attributes.rs`:

```
Some(offset) => (offset..).zip(slice[offset..].iter()),
```

`recover(slice)` can return `Some(offset)` where `offset > slice.len()`. The code then indexes `slice[offset..]` without checking the bound first.

lntutor:
`Attributes::new` and `Attributes::html` document that a start position beyond the input returns an empty iterator, but the unchecked recovery offset currently panics while slicing.

This change bounds the initial recovery position and adds regression coverage for positions both at and beyond the input end through direct `next()` calls and indirect iteration via `has_nil()`. It also records the fix in the changelog.

Verification:
- `cargo test --all-features start_position_at_or_past_end_is_empty`
- `cargo fmt --all --check`
- `cargo test --all-features`
- `git diff --check`

Closes #989

### DIFF
--- a/src/events/attributes.rs
+++ b/src/events/attributes.rs
@@ -946,7 +946,8 @@ impl IterState {
     fn recover(&self, slice: &[u8]) -> Option<usize> {
         match self.state {
             State::Done => None,
-            State::Next(offset) => Some(offset),
+            State::Next(offset) if offset <= slice.len() => Some(offset),
+            State::Next(_) => None,
             State::SkipValue(offset) => self.skip_value(slice, offset),
             State::SkipEqValue(offset) => self.skip_eq_value(slice, offset),
         }
@@ -1260,6 +1261,30 @@ impl IterState {
 mod xml {
     use super::*;
     use pretty_assertions::assert_eq;
+
+    #[test]
+    fn start_position_at_end_is_empty() {
+        let mut attributes = Attributes::new("a", 1);
+        assert_eq!(attributes.next(), None);
+
+        let mut attributes = Attributes::html("a", 1);
+        assert_eq!(attributes.next(), None);
+
+        let mut attributes = Attributes::new("a", 1);
+        assert!(!attributes.has_nil(&NamespaceResolver::default()));
+    }
+
+    #[test]
+    fn start_position_past_end_is_empty() {
+        let mut attributes = Attributes::new("a", 2);
+        assert_eq!(attributes.next(), None);
+
+        let mut attributes = Attributes::html("a", 2);
+        assert_eq!(attributes.next(), None);
+
+        let mut attributes = Attributes::new("a", 2);
+        assert!(!attributes.has_nil(&NamespaceResolver::default()));
+    }
 
     mod attribute_value_normalization {
         use super::*;
--- a/Changelog.md
+++ b/Changelog.md
@@ -57,6 +57,8 @@ The MSRV has been raised to 1.86.
 
 ### Bug Fixes
 
+- [#989]: `Attributes::new` and `Attributes::html` now return empty iterators when
+  their starting position is past the end of the input instead of panicking.
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
   event) now returns the new `NamespaceError::TooDeeplyNested` when a document
   nests elements deeper than `u16::MAX`, instead of overflowing the internal
@@ -77,6 +79,7 @@ The MSRV has been raised to 1.86.
 [#963]: https://github.com/tafia/quick-xml/pull/963
 [#977]: https://github.com/tafia/quick-xml/issues/977
 [#983]: https://github.com/tafia/quick-xml/issues/983
+[#989]: https://github.com/tafia/quick-xml/issues/989
 
 ## 0.41.0 -- 2026-06-29
 


src/events/attributes.rs
```diff
@@ -946,7 +946,8 @@ impl IterState {
     fn recover(&self, slice: &[u8]) -> Option<usize> {
         match self.state {
             State::Done => None,
-            State::Next(offset) => Some(offset),
+            State::Next(offset) if offset <= slice.len() => Some(offset),
```
dralley, line 949: Shouldn't this be `<`, since an offset == slice.len() would be over the boundary by 1 and fail to index

dralley: The test should also cover this case.

dralley: Maybe this should have a `debug_assert!()` also.  It's still a coding error to pass in an invalid position and ideally those would be caught.

Although I suppose that has implications for the tests, as they might then panic depending on compilation mode.

lntutor: `offset == slice.len()` is a valid end position: Rust permits `slice[len..]` and it produces an empty slice, so `<=` preserves that endpoint while rejecting only positions that would panic. I added explicit coverage for both `pos == len` and `pos > len` through XML `next()`, HTML `next()`, and `has_nil()` in 7df33b4. I left out a `debug_assert!` because `Attributes::new` and `Attributes::html` publicly document that positions beyond the buffer return `None`, including in debug builds.

src/events/attributes.rs
```diff
@@ -1261,6 +1262,21 @@ mod xml {
...
+        for pos in [1, 2] {
+            let mut attributes = Attributes::new("a", pos);
+            assert_eq!(attributes.next(), None);
+
+            let mut attributes = Attributes::html("a", pos);
+            assert_eq!(attributes.next(), None);
+
+            let reader = crate::reader::NsReader::from_str("");
+            let mut attributes = Attributes::new("a", pos);
+            assert!(!attributes.has_nil(reader.resolver()));
```
Mingun:
```suggestion
            let mut attributes = Attributes::new("a", pos);
            assert!(!attributes.has_nil(&NamespaceResolver::default()));
```

It is also better to split this test into two -- for `pos = 1` and `pos = 2`. If the test falls for any reason, it will be easier to determine at what value

Mingun (changes requested):
Good, but please split the test into two.

src/events/attributes.rs
lntutor: Split the regression into `start_position_at_end_is_empty` and `start_position_past_end_is_empty`, and switched the `has_nil()` assertions to `NamespaceResolver::default()` in 997f56e. Both focused tests pass with all features, and `cargo fmt --all -- --check` is clean.

dralley (approved):

dralley (approved):

lntutor:
@Mingun The namespace tests are split by behavior and the resolver setup uses NamespaceResolver::default() in 997f56e. The full CI matrix is green, including coverage. Please re-review when convenient.

Mingun:
@lntutor, as you see @dralley merged this PR which means everything is correct.

PR #993, opened 2026-07-30
Fix empty text after trimming whitespace

### LINKED ISSUE #984: `trim_text_end` still emits an empty `Text` event for whitespace-only text before a tag

lntutor:
`Config::trim_text_end` promises not to emit a text event when trimming leaves it empty, but whitespace-only text immediately before markup still produced `Text("")`.

This change lets both slice and buffered readers skip those whitespace-only spans before constructing an event. `read_to_end` temporarily disables both start and end trimming so its byte-range accounting remains unchanged.

Regression coverage exercises both `Reader::from_str` and buffered `Reader::from_reader`.

Verification:
- `cargo test skips_whitespace_only_text_before_markup --all-features`
- `cargo test --all-features`
- `git diff --check`

Closes #984

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -57,6 +57,8 @@ The MSRV has been raised to 1.86.
 
 ### Bug Fixes
 
+- [#984]: `Reader` no longer emits an empty `Text` event when `trim_text_end`
+  removes whitespace-only text immediately before markup.
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
   event) now returns the new `NamespaceError::TooDeeplyNested` when a document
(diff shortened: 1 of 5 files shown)


lntutor:
Closing as a duplicate of #986, which was already open for the same reported behavior. Sorry for the overlap.

Issue #710, opened 2024-01-28
Enum Variants with Repeated Fields


spiegela:
I'm encountering an error when using repeated fields within enum variant type. Here's a test case that reproduces the error:

```rust
    #[test]
    fn test_enum_variant_with_repeated() {
        #[derive(Debug, Deserialize)]
        #[serde(tag = "@type", rename_all = "lowercase")]
        enum Kind {
            One(OneType),
            Two(()),
        }
        #[derive(Debug, Deserialize)]
        struct OneType {
            #[serde(default)]
            item: Vec<()>
        }
        let value = r#"<kind type="one"> <item/> <item/> </kind>"#;
        let el: Kind = quick_xml::de::from_str(value).unwrap();
        if let Kind::One(o) = el {
            assert_eq!(o.item.len(), 2);
        } else {
            panic!("wrong item type");
        }
    }
```

This case returns this error, which indicates the deserializer is expecting a `map`/`struct` instead of a `Vec<>`.
```
called `Result::unwrap()` on an `Err` value: Custom("invalid type: map, expected a sequence")
thread 'mvce::tests::test_enum_variant_with_repeated' panicked at src/mvce.rs:140:55:
called `Result::unwrap()` on an `Err` value: Custom("invalid type: map, expected a sequence")
```

Mingun:
Unfortunately, this cannot be fixed from quick-xml side, because of internal buffering that serde performs to find a tag in the data. In general, this is instance of serde-rs/serde#1183. In that case it could work if serde-rs/serde#1495 will be implemented, because this optimization will use original deserializer if the tag is the first field in data as here.

PR #995, opened 2026-07-30
Fix "enum newtype variant" deserialization (#819)

### LINKED ISSUE #819: Deserialization of enum variant which recursively refers to itself faild with stackoverflow

dralley:
Enum newtype variants whose inner type is also an enum (e.g. `Apply(Vec<MathNode>)`) caused infinite re-entry because
`newtype_variant_seed` passed the raw Deserializer back without consuming the variant's Start event.

Introduce VariantContentDeserializer to consume the Start event first, then dispatch correctly for all inner types: structs use the consumed start as root, enums/sequences delegate to the stream-based Deserializer with proper End-tag cleanup.

This also fixes two related issues:

- Newtype variants containing sequences of child elements (e.g.   `Variant(Vec<i32>)` with `<Variant><x>1</x><x>2</x></Variant>`)   previously failed with `UnexpectedStart` because the unconsumed Start event was treated as the first sequence element.
- Sequences inside newtype variants could consume past the enclosing  variant's End tag, because the top-level SeqAccess only stops on Eof. The new VariantSeqAccess correctly stops on End events.

Assisted-By: Claude Opus 4.6
closes https://github.com/tafia/quick-xml/issues/819

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -62,8 +62,9 @@ The MSRV has been raised to 1.86.
 ### Bug Fixes
 
 - [#819]: Fixed infinite recursion (stack overflow) when deserializing enum
-  newtype variants whose inner type is also an enum (e.g. `Apply(Vec<MathNode>)`).
-  The variant's Start event is now consumed before deserializing the inner type,
+  newtype and tuple variants whose inner type is also an enum
+  (e.g. `Apply(Vec<MathNode>)` or `Parent(Box<E>, String)`).
+  Each variant's Start event is now consumed before deserializing the inner type,
   and sequence boundary handling is corrected so that sequences no longer consume
   past the enclosing variant's end tag.
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
--- a/Changelog.md
+++ b/Changelog.md
@@ -61,21 +61,28 @@ The MSRV has been raised to 1.86.
 
 ### Bug Fixes
 
-- [#989]: `Attributes::new` and `Attributes::html` now return empty iterators when
-  their starting position is past the end of the input instead of panicking.
+- [#819]: Fixed infinite recursion (stack overflow) when deserializing enum
+  newtype variants whose inner type is also an enum (e.g. `Apply(Vec<MathNode>)`).
+  The variant's Start event is now consumed before deserializing the inner type,
+  which also fixes deserialization of newtype variants containing sequences of
+  child elements (e.g. `Variant(Vec<i32>)` with `<Variant><x>1</x><x>2</x></Variant>`),
+  which previously failed with `UnexpectedStart`, and corrects sequence boundary
+  handling so that sequences no longer consume past the enclosing variant's end tag.
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
   event) now returns the new `NamespaceError::TooDeeplyNested` when a document
   nests elements deeper than `u16::MAX`, instead of overflowing the internal
   `u16` depth counter. Previously the unguarded `nesting_level += 1` panicked
   under `overflow-checks` builds and silently wrapped in release, corrupting
   namespace-scope bookkeeping on deeply nested untrusted input.
-- [#980]: `NamespaceResolver` now caps the total number of in-scope namespace
-  bindings (default 128, configurable via `set_max_namespace_bindings`),
-  replacing the previous per-element `max_declarations_per_element` limit.
 - [#978]: The serde `Deserializer` now enforces a configurable recursion-depth
   limit (default 128, matching `serde_json`). Deeply nested XML returns
   `DeError::TooDeeplyNested` instead of overflowing the native call stack.
   Use `Deserializer::recursion_limit()` to adjust.
+- [#980]: `NamespaceResolver` now caps the total number of in-scope namespace
+  bindings (default 128, configurable via `set_max_namespace_bindings`),
+  replacing the previous per-element `max_declarations_per_element` limit.
+- [#989]: `Attributes::new` and `Attributes::html` now return empty iterators when
+  their starting position is past the end of the input instead of panicking.
 
 ### Misc Changes
 
@@ -88,11 +95,12 @@ The MSRV has been raised to 1.86.
   `decode_and_unescape_value_with()`. Use `normalized_value()` and
   `normalized_value_with()` instead.
 
+[#819]: https://github.com/tafia/quick-xml/issues/819
+[#859]: https://github.com/tafia/quick-xml/issues/859
 [#963]: https://github.com/tafia/quick-xml/pull/963
 [#977]: https://github.com/tafia/quick-xml/issues/977
-[#980]: https://github.com/tafia/quick-xml/issues/980
-[#859]: https://github.com/tafia/quick-xml/issues/859
 [#978]: https://github.com/tafia/quick-xml/issues/978
+[#980]: https://github.com/tafia/quick-xml/issues/980
 [#983]: https://github.com/tafia/quick-xml/issues/983
 [#989]: https://github.com/tafia/quick-xml/issues/989
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -64,10 +64,8 @@ The MSRV has been raised to 1.86.
 - [#819]: Fixed infinite recursion (stack overflow) when deserializing enum
   newtype variants whose inner type is also an enum (e.g. `Apply(Vec<MathNode>)`).
   The variant's Start event is now consumed before deserializing the inner type,
-  which also fixes deserialization of newtype variants containing sequences of
-  child elements (e.g. `Variant(Vec<i32>)` with `<Variant><x>1</x><x>2</x></Variant>`),
-  which previously failed with `UnexpectedStart`, and corrects sequence boundary
-  handling so that sequences no longer consume past the enclosing variant's end tag.
+  and sequence boundary handling is corrected so that sequences no longer consume
+  past the enclosing variant's end tag.
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
   event) now returns the new `NamespaceError::TooDeeplyNested` when a document
   nests elements deeper than `u16::MAX`, instead of overflowing the internal
--- a/src/de/var.rs
+++ b/src/de/var.rs
@@ -200,13 +200,12 @@ where
         visitor.visit_newtype_struct(self)
     }
 
-    fn deserialize_seq<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    fn deserialize_seq<V>(mut self, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: Visitor<'de>,
     {
-        let mut this = self;
-        let result = visitor.visit_seq(&mut this)?;
-        this.de.read_to_end(this.start.name())?;
+        let result = visitor.visit_seq(&mut self)?;
+        self.de.read_to_end(self.start.name())?;
         Ok(result)
     }
 
--- a/src/de/var.rs
+++ b/src/de/var.rs
@@ -311,7 +311,110 @@ where
         match self.de.peek()? {
             DeEvent::End(_) => Ok(None),
             DeEvent::Eof => Err(Error::missed_end(self.start.name()).into()),
-            _ => seed.deserialize(&mut *self.de).map(Some),
+            _ => seed
+                .deserialize(InTreeDeserializer { de: &mut *self.de })
+                .map(Some),
         }
+    }
+}
+
+/// A wrapper around [`Deserializer`] that rejects `Start` events when
+/// deserializing primitive types. Used by [`VariantContentDeserializer`]'s
+/// [`SeqAccess`](de::SeqAccess) impl so that `Variant(Vec<i32>)` cannot be
+/// deserialized from child elements like `<Variant><x>1</x></Variant>` —
+/// only from xs:list text content like `<Variant>1 2 3</Variant>`.
+///
+/// Non-primitive methods (enum, struct, seq, etc.) delegate directly to the
+/// underlying [`Deserializer`], so `Vec<MathNode>` from child elements still
+/// works: `<Apply><Ci>x</Ci><Cn>5</Cn></Apply>`.
+pub(crate) struct InTreeDeserializer<'de, 'd, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    de: &'d mut Deserializer<'de, R, E>,
+}
+
+impl<'de, 'd, R, E> InTreeDeserializer<'de, 'd, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    #[inline]
+    fn read_string(&mut self) -> Result<Cow<'de, str>, DeError> {
+        self.de.read_string_in_tree()
+    }
+}
+
+impl<'de, 'd, R, E> de::Deserializer<'de> for InTreeDeserializer<'de, 'd, R, E>
+where
+    R: XmlRead<'de>,
+    E: EntityResolver,
+{
+    type Error = DeError;
+
+    deserialize_primitives!(mut);
+
+    fn deserialize_unit<V>(self, visitor: V) -> Result<V::Value, DeError>
+    where
+        V: Visitor<'de>,
+    {
+        self.de.deserialize_unit(visitor)
+    }
+
+    fn deserialize_option<V>(self, visitor: V) -> Result<V::Value, DeError>
+    where
+        V: Visitor<'de>,
+    {
+        self.de.deserialize_option(visitor)
+    }
+
+    fn deserialize_newtype_struct<V>(
+        self,
+        name: &'static str,
+        visitor: V,
+    ) -> Result<V::Value, DeError>
+    where
+        V: Visitor<'de>,
+    {
+        self.de.deserialize_newtype_struct(name, visitor)
+    }
+
+    fn deserialize_seq<V>(self, visitor: V) -> Result<V::Value, DeError>
+    where
+        V: Visitor<'de>,
+    {
+        self.de.deserialize_seq(visitor)
+    }
+
+    fn deserialize_struct<V>(
+        self,
+        name: &'static str,
+        fields: &'static [&'static str],
+        visitor: V,
+    ) -> Result<V::Value, DeError>
+    where
+        V: Visitor<'de>,
+    {
+        self.de.deserialize_struct(name, fields, visitor)
+    }
+
+    fn deserialize_enum<V>(
+        self,
+        name: &'static str,
+        variants: &'static [&'static str],
+        visitor: V,
+    ) -> Result<V::Value, DeError>
+    where
+        V: Visitor<'de>,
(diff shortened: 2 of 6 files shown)


dralley:
Test results including the new tests but prior to adding the commit with the fix: https://github.com/tafia/quick-xml/actions/runs/30572674406/job/90973123053

dralley:
I looked for existing reports of the two other issues mentioned and didn't find any, #710 is similar but it is about internally-tagged enums rather than externally tagged ones.  But maybe I missed something.

src/de/var.rs
```diff
@@ -141,3 +151,148 @@ where
...
+{
+    type Error = DeError;
+
+    fn next_element_seed<T>(&mut self, seed: T) -> Result<Option<T::Value>, Self::Error>
+    where
+        T: DeserializeSeed<'de>,
+    {
+        self.de.skip_whitespaces()?;
+        match self.de.peek()? {
+            DeEvent::End(_) | DeEvent::Eof => Ok(None),
```
Mingun: `Eof` here should be an error, because we always inside opened tag.

dralley: Pushed in a separate commit, will squash before merging.

dralley: FWIW I was not considering it ready for review just yet, I was just getting the CI running while I reviewed.

Changelog.md
```diff
@@ -61,21 +61,28 @@ The MSRV has been raised to 1.86.
 
 ### Bug Fixes
 
-- [#989]: `Attributes::new` and `Attributes::html` now return empty iterators when
-  their starting position is past the end of the input instead of panicking.
+- [#819]: Fixed infinite recursion (stack overflow) when deserializing enum
```
dralley, line 64: Sidenote: I presume you prefer pull request numbers here, just going by historical patterns, but do you care strongly about it or is it fine to just use the issue.

My personal preference would be to always use the issue, unless it's not linked to an issue.

Mingun: No, I agree with you -- always use the issue, unless it's not linked to an issue.

src/de/var.rs
Mingun: I understand. Just want to give early feedback in case you are not already aware of the problem to help you to keep number of iterations low.

tests/serde-issues.rs
```diff
@@ -1029,83 +1029,71 @@ mod issue978 {
...
         de.recursion_limit(3);
-        assert!(matches!(
-            E::deserialize(&mut de),
-            Err(DeError::TooDeeplyNested(_))
-        ));
+        assert_eq!(
+            E::deserialize(&mut de).unwrap(),
+            E::Wrap(4.2, "answer".into())
+        );
     }
```
Mingun: Why you remove recursion from the type? Now it does not test what we need and is redundant -- there is already such a test here:
https://github.com/tafia/quick-xml/blob/4e85c3d20c4e02ab37a8cd1e75aa5ba3cb58f326/tests/serde-de-enum.rs#L84-L88

tests/serde-issues.rs
```diff
@@ -1238,3 +1226,129 @@ mod issue978 {
...
+    fn newtype_variant_with_primitive() {
+        #[derive(Debug, Deserialize, PartialEq)]
+        enum E {
+            Variant(String),
+        }
+
+        let result: E = from_str("<Variant>hello</Variant>").unwrap();
+        assert_eq!(result, E::Variant("hello".into()));
+    }
+
```
Mingun: Such test already exist, redundant here:
https://github.com/tafia/quick-xml/blob/4e85c3d20c4e02ab37a8cd1e75aa5ba3cb58f326/tests/serde-de-enum.rs#L78-L82

```suggestion
```

tests/serde-issues.rs
```diff
@@ -1238,3 +1226,129 @@ mod issue978 {
...
+        enum E {
+            Numbers(Vec<i32>),
+        }
+
+        let err = from_str::<E>("<Numbers><n>1</n>").unwrap_err();
+        assert!(
+            matches!(err, DeError::InvalidXml(_)),
+            "expected InvalidXml error for truncated input, got: {err:?}"
+        );
+    }
```
Mingun: Move those tests next to this to keep all enum tests, that not bug reproducers, in one place:
https://github.com/tafia/quick-xml/blob/4e85c3d20c4e02ab37a8cd1e75aa5ba3cb58f326/tests/serde-de-enum.rs#L78-L82

tests/serde-issues.rs
```diff
@@ -1238,3 +1226,129 @@ mod issue978 {
...
+        }
+
+        let mut de = XmlDeserializer::from_str(&xml);
+        de.recursion_limit(5);
+        assert!(matches!(
+            MathNode::deserialize(&mut de),
+            Err(DeError::TooDeeplyNested(n)) if n == 5
+        ));
+    }
+
```
Mingun: That is already tested in `newtype_variant_vec` at line 1057, no need to test it again.

```suggestion
```

tests/serde-issues.rs
```diff
@@ -1238,3 +1226,129 @@ mod issue978 {
...
+    fn math_expression() {
+        let xml = r#"<math><apply><ci>5</ci></apply></math>"#;
+        let result: MathNode = from_str(xml).unwrap();
+        assert_eq!(
+            result,
+            MathNode::Root(vec![MathNode::Apply(vec![MathNode::Ci(vec![
+                MathNode::Text("5".into())
+            ])])])
+        );
+    }
```
Mingun: When other tests gone, there will be only one test in `mod`, convert `mod issue819` to `fn issue819()`

tests/serde-issues.rs
```diff
@@ -1238,3 +1226,129 @@ mod issue978 {
...
+            x: i32,
+        }
+
+        #[derive(Debug, Deserialize, PartialEq)]
+        enum E {
+            Variant(Inner),
+        }
+
+        let result: E = from_str("<Variant><x>42</x></Variant>").unwrap();
+        assert_eq!(result, E::Variant(Inner { x: 42 }));
```
Mingun: It would be also worth to check the attributes support:

```rust
let result: E = from_str("<Variant x='42'/>").unwrap();
assert_eq!(result, E::Variant(Inner { x: 42 }));
```

tests/serde-issues.rs
Mingun: Actually, I do not like that. Where did the name `n` come from? This fundamentally cannot be serialized back in the same way.

src/de/var.rs
```diff
@@ -141,3 +150,154 @@ where
...
+        self.deserialize_map(visitor)
+    }
+}
+
+/// A [`SeqAccess`] that iterates over child elements within a variant element.
+///
+/// Unlike the top-level [`SeqAccess`] impl on [`Deserializer`] (which only stops
+/// on [`DeEvent::Eof`]), this stops on [`DeEvent::End`] as well, since the
+/// sequence is bounded by the enclosing variant element's end tag.
+struct VariantSeqAccess<'de, 'd, R, E>
```
Mingun: Separate struct is not necessary, you may implement trait for `&mut VariantContentDeserializer<'de, 'd, R, E>`. Less code -- more clear, I think. What do you think?

Also, "as well" assumes, that it stops (without errors) at `End` _and_ `Eof`, which is not true.

src/de/var.rs
```diff
@@ -102,7 +104,14 @@ where
...
-            seed.deserialize(self.de)
+            match self.de.next()? {
+                DeEvent::Start(e) => seed.deserialize(VariantContentDeserializer {
+                    start: e,
+                    de: self.de,
+                }),
+                // SAFETY: the other events are filtered in `variant_seed()`
+                _ => unreachable!("Only `Start` events are possible here"),
+            }
         }
```
Mingun: It seems it is even unnecessary to check `is_text` here, this also worked:
```rust
        match self.de.next()? {
            DeEvent::Start(e) => seed.deserialize(VariantContentDeserializer {
                start: e,
                de: self.de,
            }),
            DeEvent::Text(e) => seed.deserialize(SimpleTypeDeserializer::from_text_content(e)),
            // SAFETY: the other events are filtered in `variant_seed()`
            _ => unreachable!("Only `Start` or `Text` events are possible here"),
        }
```
Perhaps, that check is redundant for `tuple_variant` too and `is_text` could be removed.

Mingun (changes requested):
Could you rearrange the tests and add missing ones (or explain why you removed them)?

src/de/var.rs
dralley: Agreed, it's cleaner that way.

tests/serde-issues.rs
dralley: You are correct, my bad

src/de/var.rs
dralley: It does look like it can be removed

tests/serde-issues.rs
dralley: I  think the commit is just messy because of what was moved around - some of which obviously should have been deleted (the duplicates)

Sorry.  Cleaning this up

dralley:
Squashed everything into the original commits

New test commit CI run, no fixes: https://github.com/tafia/quick-xml/actions/runs/30657807408/job/91246450438?pr=995

tests/serde-issues.rs
dralley: Added to the serde-de-enum tests.

tests/serde-issues.rs
```diff
@@ -1029,82 +1029,51 @@ mod issue978 {
 
     #[test]
     fn newtype_variant_box() {
-        #[derive(Debug, Deserialize)]
+        #[derive(Debug, Deserialize, PartialEq)]
```
dralley, line 1032: So, just double checking - you also want `newtype_variant_vec` and  `newtype_variant_box`  moved to `serde-de-enum.rs`, yes?

dralley: Many of the other tests under `mod issue978` (not these specifically, the existing untouched ones that aren't part of this PR) feel like they belong elsewhere also - they seem very generalized.

Mingun: > So, just double checking - you also want `newtype_variant_vec` and `newtype_variant_box` moved to `serde-de-enum.rs`, yes?

Yes, that's natural place for those tests.

`issues.rs` and `serde-issues.rs` are intended for tests which would test literal scenarios from our issues to ensure, that what was fixed will never broken again. Sometimes they can be shortened by throwing out code that is clearly unnecessary for reproduction, and of course they are adapted to the current API. But the main idea is to still save the code that was provided in the issue. The rest should be placed in regular tests.

dralley: Well, both of these tests rely on `nested_enum_value` and are specifically related to deep nesting.  They fit in well here, especially if you look at the surrounding tests.  So to me it feels like they should either stay, or most of what is here should be moved elsewhere and not just these two.

dralley: It's just less obvious where the others would go - probably not in the enum tests, maybe `mod nested_struct` in `serde-de.rs`.  It would mean the nesting tests get split apart and distributed.

I'm slightly in favor of just leaving everything here.  Yes it's the same enum layout (roughly) but it's testing a different thing.

Mingun: Well, then let's leave it as is.

dralley: Ok.  So the structure is: `mod issue978` tests the recursion depth limits specifically, and correctness of the actual values is tested elsewhere (distributed across `serde-de-enum` etc.)

tests/serde-de.rs
```diff
@@ -1427,6 +1427,93 @@ mod nested_struct {
     }
 }
 
+mod recursive_struct {
```
dralley, line 1430: As best I could tell, we were missing value-correctness tests for recursive structs, only the recursion limit tests in `mod issue978` [(`struct_field_box`, `struct_field_vec`, `struct_value_field_box`, `struct_value_field_vec`)](https://github.com/tafia/quick-xml/blob/master/tests/serde-issues.rs#L927-L1026) 

So I will go ahead and add tests for those in `serde-de.rs`, similar to the ones just added for newtypes in `serde-de-enum.rs`.

tests/serde-de.rs
```diff
@@ -1427,6 +1427,93 @@ mod nested_struct {
...
+
+    #[test]
+    fn value_box_field() {
+        #[derive(Debug, Deserialize, PartialEq)]
+        struct S {
+            #[serde(rename = "$value")]
+            value: Option<Box<S>>,
+        }
+
+        let data: S = from_str("<a><b><c/></b></a>").unwrap();
```
dralley, line 1482: With `rename = $value`, the name of the tags doesn't matter, right?  So this is fine - or would you prefer to mimic `nested_struct_value()` exactly with something like `<S><S><S><\S><\S><\S>`

(I've never used the `serde` integration much and never dug into it deeply compared to the core `Reader` code).

tests/serde-de.rs
Mingun: Ok

tests/serde-de.rs
Mingun: You're right. Yes, the implemented way is preferred.

src/de/var.rs
```diff
@@ -141,3 +132,140 @@ where
...
+
+    fn deserialize_seq<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let mut this = self;
+        let result = visitor.visit_seq(&mut this)?;
+        this.de.read_to_end(this.start.name())?;
+        Ok(result)
+    }
```
Mingun: Why not
```suggestion
    fn deserialize_seq<V>(mut self, visitor: V) -> Result<V::Value, Self::Error>
    where
        V: Visitor<'de>,
    {
        let result = visitor.visit_seq(&mut self)?;
        self.de.read_to_end(self.start.name())?;
        Ok(result)
    }
```
?

src/de/var.rs
```diff
@@ -141,3 +132,140 @@ where
...
+
+    fn next_element_seed<T>(&mut self, seed: T) -> Result<Option<T::Value>, Self::Error>
+    where
+        T: DeserializeSeed<'de>,
+    {
+        self.de.skip_whitespaces()?;
+        match self.de.peek()? {
+            DeEvent::End(_) => Ok(None),
+            DeEvent::Eof => Err(Error::missed_end(self.start.name()).into()),
+            _ => seed.deserialize(&mut *self.de).map(Some),
```
Mingun: Probably we should use another deserializer here. `Deserializer` have some assumptions that it is top-level deserializer (yes, probably in some other places the same reborrow used incorrectly). In particular, it allows deserializing primitives from the `<tag>content</tag>`, but when we inside XML tree, we should accept only `content`.

Need to check, what tests reaches this line.

tests/serde-issues.rs
```diff
@@ -1029,82 +1029,51 @@ mod issue978 {
...
             E::deserialize(&mut de),
-            Err(DeError::TooDeeplyNested(_))
+            Err(DeError::TooDeeplyNested(n)) if n == limit
         ));
     }
 
-    // enum tuple variant
-
     #[test]
-    fn tuple_variant_box() {
```
Mingun, line 1074: So, why were tuple recursion tests removed? We should check that this scenario also work.

Changelog.md
```diff
@@ -61,21 +61,28 @@ The MSRV has been raised to 1.86.
...
+  newtype variants whose inner type is also an enum (e.g. `Apply(Vec<MathNode>)`).
+  The variant's Start event is now consumed before deserializing the inner type,
+  which also fixes deserialization of newtype variants containing sequences of
+  child elements (e.g. `Variant(Vec<i32>)` with `<Variant><x>1</x><x>2</x></Variant>`),
+  which previously failed with `UnexpectedStart`, and corrects sequence boundary
+  handling so that sequences no longer consume past the enclosing variant's end tag.
```
Mingun: I still do not want deserialization of `Variant(Vec<i32>)` from `<Variant><x>1</x><x>2</x></Variant>`. I think, it would be serialized as `<Variant>1 2</Variant>` (using `SimpleTypeSerializer`), so it should be deserialized using `SimpleTypeDeserializer`. Acceptance of that form is allowed due to using top-level `Deserializer` when we inside the XML tree.

Mingun (changes requested):
It seems that for proper implementation, deserialization will need to be reworked, by separating the places of use of the top-level deserializer and in-XML-tree deserializers. Without that it seems will be impossible to fix that bug without introducing others.

Here at least we need to restore `tuple_variant_box` and `tuple_variant_vec` tests, what were remove, in my opinion, without any reasons.

tests/serde-issues.rs
dralley: I misinterpreted your statement about one of the tests being redundant.  I will restore them

dralley: I see what you mean about the tuple case, but I think it can be fixed.  Let me play around with this for a bit.

Changelog.md
dralley: I don't know that it can easily work both ways - both support recursion a la #819 (see: `MathNode` case) while also using `SimpleTypeSerializer` and rejecting `<tag>content</tag>` 

Anyway, the scope of this has gotten beyond what I'm comfortable working on right now, considering I don't know the serde code anywhere near as well as the rest of the library.

dralley: I will pause this for now, and maybe resurrect it later, probably after the next release.  The only impetus for working on it was that it was related to 978.

Issue #851, opened 2025-03-27
Consider whether a DOM wrapper would be a useful addition


dralley:
@Tafia wrote and @ColinKinloch contributed to an experimental DOM wrapper API [here](https://github.com/ColinKinloch/quick-xml/blob/dom_update/src/dom.rs)

Originally the PR was closed because the `minidom` crate was using `quick-xml`.  As of right now however they seem to be using a different library.

I'm opening this mostly so that we don't completely lose track of the work, which might be useful.  But that remains to be seen. We ought to determine 

1) whether it's a useful addition to the ecosystem, which already has a few crates that handle the DOM
2) whether this crate would be the correct place for it to live if it was (I suspect it probably would not be, as long as no deep integration is required)

ColinKinloch:
Since updating that branch I've tested out `minidom` and found it to be unsuitable for my use as it mandates namespaces (`<a xmlns="ns"></a>`).
Having the ability to "flatten" unhandled values like in [`serde_json`](https://serde.rs/attr-flatten.html#capture-additional-fields) would be extremely nice.

Mingun:
I have a [dom](https://github.com/Mingun/quick-xml/tree/dom) branch with a different approach (very WIP), and yes, we definitely need a DOM type to be able to create a custom Buffer for serde if [that PR](https://github.com/serde-rs/serde/pull/2912) will be finished

PR #996, opened 2026-07-30
Fix indented writer misaligning End tags after Comment + Text 

### LINKED ISSUE #276: Bug in writing XML with indents
First, let me say thank you for providing this library!

I was working on a [function to prettify XML](https://gist.github.com/lwilli/14fb3178bd9adac3a64edfbc11f42e0d) using this library and I ran into some interesting behavior (bug?) when using a `Writer` with indentation.

In the Readme, there is this example XML (truncated):
```
<tag1 att1 = "test">
    <tag2><!--Test comment-->Test</tag2>
</tag1>
```
When I remove the new lines and indentation spaces from this, and then read and re-write this using a Writer like `Writer::new_with_indent(Vec::new(), b' ', 2)`, I get this:
```
<tag1 att1 = "test">
    <tag2>
    <!--Test comment-->Test</tag2>
</tag1>
```

The strange thing here is that the `tag2` tags do not line up. This appears to be due to how [`Writer::write_event()`](https://github.com/tafia/quick-xml/blob/745341a755fd4871668a985b6fa8391344101322/src/writer.rs#L94) handles not breaking new lines for text fields, but doesn't consider whether any other events have been written after the opening tag and before the Text.

dralley:
When an element contained a non-text child (e.g. Comment) followed by a
Text event, the closing End tag lost its line break because Text
unconditionally cleared should_line_break. Track the indent level of the
most recent non-text event and force a line break for the End tag when
that level is deeper than the current level after shrink.

closes https://github.com/tafia/quick-xml/issues/276

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -64,6 +64,10 @@ The MSRV has been raised to 1.86.
 
 ### Bug Fixes
 
+- [#276]: Fixed `Writer::new_with_indent` producing misaligned closing tags
+  when an element contained a non-text child (e.g. a comment) followed by a
+  text node. The `</tag>` was appended directly after the text instead of
+  starting on a new line at the correct indentation level.
 - [#670]: Serde serializer now escapes `\r`, `\n`, and `\t` in attribute values
   as `&#13;`, `&#10;`, and `&#9;` respectively, preventing silent data loss from
   XML attribute-value normalization on round-trip. Likewise `Attribute::from`
@@ -110,6 +114,7 @@ The MSRV has been raised to 1.86.
   to get a namespace resolver used by this deserializer, because it no longer uses
   an `NsReader` internally.
 
+[#276]: https://github.com/tafia/quick-xml/issues/276
 [#670]: https://github.com/tafia/quick-xml/issues/670
 [#859]: https://github.com/tafia/quick-xml/issues/859
 [#953]: https://github.com/tafia/quick-xml/issues/953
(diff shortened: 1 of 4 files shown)


dralley:
Unless I mark it ready for review, it's not ready yet!

dralley:
To be clear, I'm still experimenting with this.

dralley:
@Mingun There is a similar lookahead issue here (for pretty-printing) as with trimming

The user suggests the formatting should look like

```xml
<tag1 att1 = "test">
    <tag2><!--Test comment-->Test</tag2>
</tag1>
```

(which I would agree is "ideal", and is how lxml would format things).  But it roughly follows the rule "An element is indented only if none of its direct children is a text node" and  without knowing what tags might follow after the comment, the best we can do might be

```xml
<tag2>
    <!--c-->Test
</tag2>
```

Which is not good, because it modifies the text content unnecessarily.

Should we accept that a bit of event buffering may be required, whether within the writer or perhaps by pushing pretty-printing into a separate wrapper object?

dralley:
The other option, I guess, would be to leave https://github.com/tafia/quick-xml/issues/276 alone for now, if the cure is worse than the disease

dralley:
Probably tied up with https://github.com/tafia/quick-xml/issues/851

Mingun:
> Should we accept that a bit of event buffering may be required, whether within the writer or perhaps by pushing pretty-printing into a separate wrapper object?

I think, introducing a new type, `BufferedWriter`, with the ability of pretty-print, would be better.

dralley:
Ok.  I think I will leave the existing writer behavior alone, as it is perhaps uglier but a better compromise with being mostly correct + fast.

PR #997, opened 2026-07-31
Fix empty Text event emitted with trim_text_end (#984)

### LINKED ISSUE #984: `trim_text_end` still emits an empty `Text` event for whitespace-only text before a tag

dralley:
When trim_text_end is enabled, whitespace-only text before markup or a reference is now suppressed instead of producing an empty Text("") event.

closes #984

### DIFF
--- a/src/reader/buffered_reader.rs
+++ b/src/reader/buffered_reader.rs
@@ -54,6 +54,7 @@ macro_rules! impl_buffered_source {
             &mut self,
             buf: &'b mut Vec<u8>,
             position: &mut u64,
+            trim_text_end: bool,
         ) -> ReadTextResult<'b, &'b mut Vec<u8>> {
             let mut read = 0;
             let start = buf.len();
@@ -83,6 +84,9 @@ macro_rules! impl_buffered_source {
                         read += i as u64;
 
                         *position += read;
+                        if trim_text_end && buf[start..].iter().all(|&b| is_whitespace(b)) {
+                            return ReadTextResult::Markup(buf);
+                        }
                         return match std::str::from_utf8(&buf[start..]) {
                             Ok(s) => ReadTextResult::UpToMarkup(s),
                             Err(e) => ReadTextResult::Err(e.into()),
@@ -95,6 +99,9 @@ macro_rules! impl_buffered_source {
                         read += i as u64;
 
                         *position += read;
+                        if trim_text_end && buf[start..].iter().all(|&b| is_whitespace(b)) {
+                            return ReadTextResult::Ref(buf);
+                        }
                         return match std::str::from_utf8(&buf[start..]) {
                             Ok(s) => ReadTextResult::UpToRef(s),
                             Err(e) => ReadTextResult::Err(e.into()),
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -325,7 +325,11 @@ macro_rules! read_event_impl {
                         $reader.skip_whitespace(&mut $self.state.offset) $(.$await)? ?;
                     }
 
-                    match $reader.read_text($buf, &mut $self.state.offset) $(.$await)? {
+                    match $reader.read_text(
+                        $buf,
+                        &mut $self.state.offset,
+                        $self.state.config.trim_text_end,
+                    ) $(.$await)? {
                         ReadTextResult::Markup(buf) => {
                             $self.state.state = ParseState::InsideMarkup;
                             // Pass `buf` to the next next iteration of parsing loop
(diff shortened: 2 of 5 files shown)


src/reader/mod.rs
```diff
@@ -512,20 +513,26 @@ macro_rules! read_to_end {
             let end = $self.buffer_position();
             match $self.$read_event($buf) $(.$await)? {
                 Err(e) => {
-                    $self.config_mut().trim_text_start = trim;
+                    let config = $self.config_mut();
+                    config.trim_text_start = trim_start;
+                    config.trim_text_end = trim_end;
```
dralley, line 518: I don't really like the clear->restore pattern, but it was already like this.  Would be nice to refactor this away

dralley:
replaces #993 and #986

src/reader/mod.rs
Mingun: Yes, I also does not like it here... I was thinking about removing auto-trim altogether and making it an explicit method call on the `Event`

dralley:
Will resume this PR as well as 996 after 1001 merges.

Would be nice to include those, 1002, and maybe 897 (https://github.com/tafia/quick-xml/pull/897#issuecomment-3278376421) before cutting a new release.

src/reader/mod.rs
dralley: I probably don't agree with that, at least in this specific case it structurally changes what comes out of the parser.  If someone doesn't want to deal with whitespace-only text events at all (and I understand that completely, it's really quite annoying, I use trim for precisely that reason) manual trim wouldn't help.

tests/reader-config.rs
dralley, line 882: TODO: leave a comment that many of these tests are not actually demonstrating XML compliance.  Per EOL normalization, `\r\n` in an event name would get collapsed to `\n` - it's meant to be treated as a pre-processing step.

See also: https://github.com/tafia/quick-xml/issues/1003

src/reader/mod.rs
dralley: I have an old partial commit from 2022 which changes the configuration options slightly.  I think this would make a lot more sense (but I think might be more difficult to implement now, will need to think about how to make it work)

```diff
-    /// Trims leading whitespace in Text events, skip the element if text is empty
-    pub trim_text_start: bool,
-    /// Trims trailing whitespace in Text events.
-    pub trim_text_end: bool,
+    /// Preserves whitespace-only text elements between events (e.g. indented "pretty" formatting).
+    pub preserve_indentation: bool,
+    /// Trims leading and trailing whitespace in Text events.
+    pub trim_text: bool,
```

dralley: I can file an issue for that if you agree.

src/reader/buffered_reader.rs
```diff
@@ -83,6 +84,9 @@ macro_rules! impl_buffered_source {
                         read += i as u64;
 
                         *position += read;
+                        if trim_text_end && buf[start..].iter().all(|&b| is_whitespace(b)) {
+                            return ReadTextResult::Markup(buf);
+                        }
```
Mingun, line 89: I think that this is not the correct place for trim. I would trim spaces before comments and CDATA sections. It seems totally wrong to me.

src/reader/buffered_reader.rs
```diff
@@ -95,6 +99,9 @@ macro_rules! impl_buffered_source {
                         read += i as u64;
 
                         *position += read;
+                        if trim_text_end && buf[start..].iter().all(|&b| is_whitespace(b)) {
+                            return ReadTextResult::Ref(buf);
+                        }
```
Mingun, line 104: This change is wrong. We do not want to trim spaces before entity references, otherwise `" &amp; text"` becomes `"& text"` instead of `" & text"`. Compare the result with `" x text"` -- this will be unexpected for the user if the replacement in the text `x` with `&` suddenly eats up the spaces before this character.

src/reader/mod.rs
Mingun: I doubt that we can reliable detect indentation and I think that is not the work for parser at all.

Mingun (changes requested):
Unfortunately, this cannot be fix at low-level parser level. We cannot strip whitespaces before all entity references -- it is possible only after them will be expanded. `&amp;` should preserve spaces before it, but `&xml;` where it is expanded into `<a>...</a>` must not.

We also probably cannot blindly trim everything before `<`, because of CDATA sections, comments and processing instructions. Comments and PIs are stripped out from the output, so when them inside the text, spaces around them would be trimmed. I think, this would be unexpected for users.

src/reader/buffered_reader.rs
dralley: Good point, but given the existing tests pass, we probably need some additional test cases around all of this.

dralley:
I will split off the documentation fixes and new tests (which I'm adding) to a separate PR, then experiment with how to avoid having this code in the parser.

PR #1002, opened 2026-08-13
Fix too early handling of namespace declarations

### LINKED ISSUE #953: `overlapped-lists` feature causes parser errors to `xsi:nil` fields after array fields

Mingun:
I realized, that working with namespaces was entirely incorrect in serde `Deserializer`. We push and pop namespaces when we take events from the `Reader`, but we may buffer those events and check for `xsi:nil` later, when the state of `NamespaceResolver` was changed.

Now `Deserializer` holds the `NamespaceResolver` by itself and calls its methods only when event is consumed by `Deserializer::next()` call.

Fixes #953
Closes #974

@weiznich, it would be great if you can check this PR on your data. I'm sure that the fix currently is incomplete, so I need some time to find the cases that is not covered yet. You may help with that.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -68,6 +68,9 @@ The MSRV has been raised to 1.86.
   as `&#13;`, `&#10;`, and `&#9;` respectively, preventing silent data loss from
   XML attribute-value normalization on round-trip. Likewise `Attribute::from`
   performs the same transformation.
+- [#953]: The serde `Deserializer` now correctly handles namespaces. Previously
+  the namespace bindings might be applied or removed before the event actually
+  was consumed which lead to a couple of bugs.
 - [#989]: `Attributes::new` and `Attributes::html` now return empty iterators when
   their starting position is past the end of the input instead of panicking.
 - [#977]: `NamespaceResolver::push` (and hence every `NsReader` `Start`/`Empty`
@@ -99,9 +102,17 @@ The MSRV has been raised to 1.86.
   `decoded_and_normalized_value_with()`, `decode_and_unescape_value()`, and
   `decode_and_unescape_value_with()`. Use `normalized_value()` and
   `normalized_value_with()` instead.
+- [#1002]: Added `NamespaceResolver::with` that allows temporary applying namespace
+  bindings from the start tag for the scope of a provided closure F, without making any
+  persistent change to the resolver. It is useful to check a peeked event which is
+  not yet consumed in custom implementations of peekable reader.
+- [#1002]: Added `Deserializer::resolver` and `Deserializer::resolver_mut` methods
+  to get a namespace resolver used by this deserializer, because it no longer uses
+  an `NsReader` internally.
 
 [#670]: https://github.com/tafia/quick-xml/issues/670
 [#859]: https://github.com/tafia/quick-xml/issues/859
+[#953]: https://github.com/tafia/quick-xml/issues/953
 [#963]: https://github.com/tafia/quick-xml/pull/963
 [#977]: https://github.com/tafia/quick-xml/issues/977
 [#978]: https://github.com/tafia/quick-xml/issues/978
@@ -110,6 +121,7 @@ The MSRV has been raised to 1.86.
 [#989]: https://github.com/tafia/quick-xml/issues/989
 [#990]: https://github.com/tafia/quick-xml/issues/990
 [#1000]: https://github.com/tafia/quick-xml/pull/1000
+[#1002]: https://github.com/tafia/quick-xml/pull/1002
 
 
 ## 0.41.0 -- 2026-06-29
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -562,7 +562,12 @@ where
     where
         V: Visitor<'de>,
     {
-        let has_nil = self.map.de.reader.reader.has_nil_attr(&self.map.start);
+        // `self.map.start` already was taken from the reader, so its namespace bindings were already processed.
+        let has_nil = self
+            .map
+            .start
+            .attributes()
+            .has_nil(&self.map.de.ns_resolver);
         if self.map.de.deserialize_opt(Some(has_nil))? {
             visitor.visit_some(self)
         } else {
--- a/src/de/map.rs
+++ b/src/de/map.rs
@@ -221,21 +221,6 @@ where
         })
     }
 
-    /// Determines if subtree started with the specified event should be skipped.
-    ///
-    /// Used to map elements with `xsi:nil` attribute set to true to `None` in optional contexts.
-    ///
-    /// We need to handle two attributes:
-    /// - on parent element: `<map xsi:nil="true"><foo/></map>`
-    /// - on this element:   `<map><foo xsi:nil="true"/></map>`
-    ///
-    /// We check parent element too because `xsi:nil` affects only nested elements of the
-    /// tag where it is defined. We can map structure with fields mapped to attributes to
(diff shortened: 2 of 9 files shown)


Mingun:
Ok, it seems that now all possible paths are covered

Mingun:
Wait a bit. Found another suspicious test

Mingun:
Ok, I believe _now_ everything was catched and analysed.

src/de/mod.rs
```diff
@@ -2798,14 +2798,19 @@ where
         }
     }
 
+    #[cfg(feature = "overlapped-lists")]
+    fn take_peeked(&mut self) -> Option<DeEvent<'de>> {
+        self.read.pop_front()
+    }
+
+    #[cfg(not(feature = "overlapped-lists"))]
```
dralley, line 2810: Not immediately actionable, but Rust 1.95 has `cfg_select` - would be nice to use eventually.

src/de/map.rs
```diff
@@ -562,7 +562,12 @@ where
     where
         V: Visitor<'de>,
     {
-        let has_nil = self.map.de.reader.reader.has_nil_attr(&self.map.start);
+        // `self.map.start` already was taken from the reader, so its namespace bindings already processed.
```
dralley: "its namespace bindings were already processed"

Changelog.md
```diff
@@ -68,6 +68,9 @@ The MSRV has been raised to 1.86.
   as `&#13;`, `&#10;`, and `&#9;` respectively, preventing silent data loss from
   XML attribute-value normalization on round-trip. Likewise `Attribute::from`
   performs the same transformation.
+- [#953]: The serde `Deserializer` now correctly handle namespaces. Previously
```
dralley: `Deserializer` now correctly handle**s** namespaces.

Changelog.md
```diff
@@ -68,6 +68,9 @@ The MSRV has been raised to 1.86.
...
+  the namespace bindings might be applied or removed before the event actually
+  was consumed which lead to couple of bugs.
```
dralley: to a couple

src/name.rs
```diff
@@ -777,6 +777,19 @@ impl NamespaceResolver {
         self.set_level(self.nesting_level.saturating_sub(1));
     }
 
+    /// Runs action as if all namespaces from the specified `start` element was added
+    /// to the resolver. Actually, resolver state remains unchanged after this call,
+    /// but inside the action resolver have all namespaces from the `start`.
+    pub fn with<F, R>(&mut self, start: &BytesStart, mut action: F) -> Result<R, NamespaceError>
```
dralley, line 782: > Runs `action` as if all namespaces from the specified `start` element were added to the resolver, but without actually changing the resolver state.

Changelog.md
```diff
@@ -99,9 +102,16 @@ The MSRV has been raised to 1.86.
   `decoded_and_normalized_value_with()`, `decode_and_unescape_value()`, and
   `decode_and_unescape_value_with()`. Use `normalized_value()` and
   `normalized_value_with()` instead.
+- [#1002]: Added `NamespaceResolver::with` that allows temporary apply namespace
+  bindings from the start tag and run query on the resolver. It is useful to check
+  peeked event which is not yet consumed.
```
dralley: > - [#1002]: Added `NamespaceResolver::with` that allows temporarily applying namespace
>  bindings from the start tag for the scope of a provided closure F, without making any
>  persistent change to the resolver.

"It is useful to check a peeked event which is not yet consumed." also sounds like an internal detail - is a user going to do that?

Changelog.md
```diff
@@ -99,9 +102,16 @@ The MSRV has been raised to 1.86.
...
+- [#1002]: Added `Deserializer::resolver` and `Deserializer::resolver_mut` methods
+  to get a namespace resolver used by this deserializer, because its internally no
+  longer uses the `NsReader`.
```
dralley: >- [#1002]: Added `Deserializer::resolver` and `Deserializer::resolver_mut` methods
>  to get a namespace resolver used by this deserializer, because it no longer uses
> an `NsReader` internally.

dralley:
Code looks reasonable so far, still reviewing a bit, but will wait until @weiznich gives the :+1: with his dataset to finish.

Changelog.md
Mingun: Because `NamespaceResolver` is public and intended to use with own implementations of a peekable reader, this is a note for such users.

weiznich (approved):
I've verified that this works with our dataset. Also the code looks fine to me.

Thanks for fixing this :heart:

dralley:
Github is broken, **again**, so I cannot submit an approval.

Consider this comment an approval @Mingun

PR #1004, opened 2026-08-19
Add tests for correct trim behavior around special events

dralley:
\+ document a number of existing faults in the code, faulty tests, and incorrect documentation in the parser.

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -998,16 +998,15 @@ impl<R> Reader<R> {
 /// Result of an attempt to read XML textual data from the source.
 #[derive(Debug)]
 enum ReadTextResult<'r, B> {
-    /// Start of markup (`<` character) was found in the first byte. `<` was consumed.
+    /// The reader is positioned at `<` (start of markup). `<` was not consumed.
     /// Contains buffer that should be returned back to the next iteration cycle
     /// to satisfy borrow checker requirements.
     Markup(B),
-    /// Start of reference (`&` character) was found in the first byte.
-    /// `&` was not consumed.
+    /// The reader is positioned at `&` (start of a reference). `&` was not consumed.
     /// Contains buffer that should be returned back to the next iteration cycle
     /// to satisfy borrow checker requirements.
     Ref(B),
-    /// Contains text block up to start of markup (`<` character). `<` was consumed.
+    /// Contains text block up to start of markup (`<` character). `<` was not consumed.
     UpToMarkup(&'r str),
     /// Contains text block up to start of reference (`&` character).
     /// `&` was not consumed.
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -159,57 +159,83 @@ pub struct Config {
     /// [`check_end_names`]: Self::check_end_names
     pub trim_markup_names_in_closing_tags: bool,
 
-    /// Whether whitespace before character data should be removed.
+    /// Whether leading whitespace before character data should be removed.
     ///
-    /// When set to `true`, leading whitespace is trimmed in [`Text`] events.
-    /// If after that the event is empty it will not be pushed.
+    /// When set to `true`, whitespace at the start of [`Text`] events is
+    /// stripped. If the event becomes empty after trimming, it is still
+    /// emitted as `Text("")`.
     ///
     /// Default: `false`
     ///
-    /// <div style="background:rgba(80, 240, 100, 0.20);padding:0.75em;">
+    /// <div style="background:rgba(255, 80, 80, 0.20);padding:0.75em;">
     ///
-    /// WARNING: With this option every text events will be trimmed which is
-    /// incorrect behavior when text events delimited by comments, processing
-    /// instructions or CDATA sections. To correctly trim data manually apply
-    /// [`BytesText::inplace_trim_start`] and [`BytesText::inplace_trim_end`]
-    /// only to necessary events.
+    /// **WARNING:** This option has known issues.
+    ///
+    /// - **Incorrect trimming around comments, PIs, and CDATA sections:**
+    ///   Trimming applies to every [`Text`] event regardless of context.
+    ///   In `text <!-- comment --> more`, the leading space of ` more` is
+    ///   content, but this option trims it because the parser treats each
+    ///   text chunk independently without knowledge of the surrounding markup.
+    /// - **Empty events:** Whitespace-only text that is fully trimmed
+    ///   produces an empty `Text("")` event instead of being suppressed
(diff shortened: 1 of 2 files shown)


src/reader/mod.rs
```diff
@@ -159,57 +159,83 @@ pub struct Config {
...
-    /// When set to `true`, leading whitespace is trimmed in [`Text`] events.
-    /// If after that the event is empty it will not be pushed.
+    /// When set to `true`, whitespace at the start of [`Text`] events is
+    /// stripped. If the event becomes empty after trimming, it is still
+    /// emitted as `Text("")`.
     ///
     /// Default: `false`
     ///
-    /// <div style="background:rgba(80, 240, 100, 0.20);padding:0.75em;">
+    /// <div style="background:rgba(255, 80, 80, 0.20);padding:0.75em;">
```
dralley, line 170: Note: The current color of these warnings is green, I changed it to red.  If you have a particular reason to prefer green, I can revert, but I thought it was a bit non-standard.

https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Config.html#structfield.trim_markup_names_in_closing_tags

dralley:
Best viewed commit-by-commit.

tests/reader-config.rs
```diff
@@ -813,127 +786,253 @@ mod trim_text_start {
         );
         assert_eq!(reader.read_event().unwrap(), Event::Eof);
     }
-}
 
-mod trim_text_end {
-    use super::*;
-    use pretty_assertions::assert_eq;
+    // trim_text should only suppress whitespace-only text nodes that represent
```
dralley, line 809: Regarding the previous PR, you are correct about the various cases that were broken, however those cases are already broken regardless, and the existing tests did not capture that.

tests/reader-config.rs
```diff
@@ -698,92 +702,12 @@ mod trim_text {
         );
         assert_eq!(reader.read_event().unwrap(), Event::Eof);
     }
-}
-
-mod trim_text_start {
```
dralley, line 697: This block was copied (exactly) 3x over, once for `trim_text::false`, `trim_text_start::false`, and `trim_text_end::false`.  But those are all the same scenario, really, so I have consolidated them into one scenario.

Now there's tests for (trim_text_start, trim_text_end) == (true, true), (true, false), (false, true) and (false, false)

There is still lots of duplication as all 3 are the exact same apart from only one event in the middle, but that can be revised later.

tests/reader-config.rs
dralley: Nor did enabling this test, which of course passed on the other PR.

IMO while I could still argue that the other PR is an improvement over the status quo, it's better to leave it be until we have a proper fix rather than force users to adapt to a moving target that is not correct

src/reader/mod.rs
```diff
@@ -998,16 +998,15 @@ impl<R> Reader<R> {
 /// Result of an attempt to read XML textual data from the source.
 #[derive(Debug)]
 enum ReadTextResult<'r, B> {
-    /// Start of markup (`<` character) was found in the first byte. `<` was consumed.
```
dralley, line 1001: I think this may have drifted out of date.  It does not look to me like these are being consumed.

https://github.com/tafia/quick-xml/blob/master/src/reader/buffered_reader.rs#L73-L109
https://github.com/tafia/quick-xml/blob/master/src/reader/slice_reader.rs#L274-L306

dralley: Likewise the consumer code has `debug_assert!` that the first character of these results starts with `<`, `&`, etc.

tests/reader-config.rs
```diff
@@ -813,127 +786,253 @@ mod trim_text_start {
...
-    use pretty_assertions::assert_eq;
+    // trim_text should only suppress whitespace-only text nodes that represent
+    // inter-element indentation (pretty-printing whitespace between start/end/empty
+    // elements). It must NOT trim whitespace that is adjacent to "invisible" markup
+    // (comments, PIs, CDATA) or entity references, because that whitespace is part
+    // of the text content.
+    //
+    // The fix for https://github.com/tafia/quick-xml/issues/984 should try to fix
+    // these tests also.
+    mod correctness {
```
dralley: It would be my preference to condense these to have one test for each case which handles both leading and trailing whitespace simultaneously.  Do you agree or do you prefer it this way. 

The distinction may not still exist at the point these get enabled.

dralley: It would be my preference to condense these to have one test for each case which handles both leading and trailing whitespace simultaneously. Do you agree or do you prefer it this way?

At the point these are able to be enabled the distinction probably won't exist anymore.

src/reader/mod.rs
Mingun: I just tried to choose a color that would be equally readable in all 3 color themes

src/reader/mod.rs
```diff
@@ -159,57 +159,83 @@ pub struct Config {
...
-    /// instructions or CDATA sections. To correctly trim data manually apply
-    /// [`BytesText::inplace_trim_start`] and [`BytesText::inplace_trim_end`]
-    /// only to necessary events.
+    /// **WARNING:** This option has known issues.
+    ///
+    /// - **Incorrect trimming around comments, PIs, and CDATA sections:**
+    ///   Trimming applies to every [`Text`] event regardless of context.
+    ///   In `text <!-- comment --> more`, the trailing space of `text ` is
+    ///   content, but this option trims it because the parser treats each
+    ///   text chunk independently without knowledge of the surrounding markup.
```
Mingun: This is `trim_text_start` option and I think, you want to mean ` more` piece here

src/reader/mod.rs
Mingun: Yes, the comment is outdated. They stoped being consumed in #936

Mingun (approved):
It will be impossible to implement correct trimming without looking at the next event, so I vote for, if not immediate deletion, then at least deprecate these settings.

I already have a branch for that removal, just need to update Changelog:
https://github.com/tafia/quick-xml/compare/master...Mingun:quick-xml:cleanup-config

src/reader/mod.rs
dralley: I can revert, if you want.  I didn't try any other color schemes.  It's approximately the same shade as github's light red.

Mingun: No, its fine

dralley:
Sure, I'm fine with deprecation.  Probably shouldn't remove (yet) for the same reason I mentioned - if we were to implement a replacement, better to do so once than to force adaptation twice.

Issue #1003, opened 2026-08-15
Reader should be capable of transparent EOL normalization


dralley:
Followup to https://github.com/tafia/quick-xml/issues/158

Per XML specification, all XML readers should perform line-ending normalization as a pre-processing step (or functionally behaving as if it were a pre-processing step - [XML 1.0 Section 2.11](https://www.w3.org/TR/xml/#sec-line-ends))

This is both an ergonomic issue and also a minor correctness issue, because EOL normalization is blind to document contents and therefore also applies within elements, e.g. `<?xml\r?>` ought to be parsed as `<?xml\n?>` - which `libxml2` does in fact do.

It's an ergonomic issue in the sense that a user should not need to care about it - the `.xml_content()`, `.xml10_content()`, `.xml11_content()` functions etc. ought to be superfluous and are potentially error prone.  The user also shouldn't need to track the version declared in the declaration - they likely just want the document to be processed according to the correct specification per the declaration or according to the one they specify, but that need not change on a per-element basis.

Performing universal EOL normalization as a pre-processor pass through a bufffer would reduce total allocations and likely improve performance on average, while improving ergonomics and compliance.

The mechanism could also likely be used as a base upon which line/column tracking (https://github.com/tafia/quick-xml/issues/109) could be added.

## Potentially troublesome aspects:

* For users who process noncompliant "almost-XML" documents that contain binary blobs, we would need a way to disable it, possibly document-wide, or perhaps just during the usage of `stream()`. 
* We might want to make the current Reader constructors `from_reader()`, `from_str()`, `from_bytes()` methods buffered by-default, and perhaps add a new constructor which takes only a pre-processed `XmlDoc` type or similar where decoding + normalization has already happened, allowing proper zero-copy. If we keep zero-copy (which isn't really zero-copy in practice) at all - maybe worth doing some benchmarking to see how much time is actually saved vs. just having a buffered path.  XML parsing is easily 25-50x slower than copying the bytes around once or twice anyway, and the `XmlSource` code + macros are all quite complex.

Issue #593, opened 2023-04-14
Expose underlying Cow in Event data


outfoxxed:
Would it be possible to expose the underlying Cow in event datatypes like BytesStart and BytesEnd? I have to match only specific tags, and have to keep track of the pushed tag stack.

Ideally I would keep a `Vec<&'datasource [u8]>` as the tag stack, but I am forced to copy since a `BytesStart<'datasource>` cannot return the underlying Cow, which I would then match to `Cow::Borrowed`.

Mingun:
I would prefer to hide that fact that we store `Cow<[u8]>` internally and make you think that we store an `str`. If you sure that you will able to store `Vec<&'datasource [u8]>` (i.e. _borrowed_ data from input), then you also could store `BytesStart` / `BytesEnd` -- they will store only offsets in that case (or only `BytesEnd` if you doesn't need attributes, use [`.to_end()`](https://docs.rs/quick-xml/latest/quick_xml/events/struct.BytesStart.html#method.to_end) to convert).

Would that solution acceptable for you? Otherwise fell free to submit a PR

outfoxxed:
It would add a fair amount of compliction to my code, so I'd prefer to use the reference with the same lifetime. I'll submit a pr.

Mingun:
Note, that all events will borrow when comes from `Reader`. Also, cloning them is cheap -- the underlying `Cow` [stays](https://doc.rust-lang.org/src/alloc/borrow.rs.html#193) in `Borrowed` state, if you clone borrowed `Cow`.

dralley:
As of the current master branch, the representation is the same as the construction, both `&str` (event type contains `Cow<str>`

There's nothing that prevents deriving PartialOrd, Ord, and Hash on event types, strictly speaking.

dralley:
I'm going to consider this one closed via https://github.com/tafia/quick-xml/pull/1005

PR #1005, opened 2026-08-21
Provide better examples

### LINKED ISSUE #269: Document a couple of recommended patterns of usage
Hi, I've started using this library for a personal project, and I've found that it's difficult to figure out how my code should be structured.  I think it would be great if there were some docs that were a little more prescriptive about certain patterns you can use to accomplish certain goals, when you might want to use a state machine, etc.

It would be helpful to show some examples of how real, non-trivial code might be structured to provide clean abstractions in a larger codebase. e.g. a pattern like [this](https://github.com/rust-syndication/rss/blob/master/src/item.rs#L598-L633) for parsing nested objects using nested readers. (The provided nested reader example uses no abstractions - if you try to make it more sophisticated than it already is it would get messy very quickly).  

Another example could be the state machine pattern used in this blog post: https://usethe.computer/posts/14-xmhell.html.  The issue68 example is somewhat similar but the namespaces make it more difficult to understand what the general case might look like.

If you're not keen on putting too much detail in the quick-xml docs, then maybe just linking to a few projects / blog posts which use quick-xml "well" would be a good idea, or explain some of the general principles.

A sidenote: 

Nearly every project I've looked at has some kind of implementation of [get_element_text](https://github.com/rust-syndication/rss/blob/master/src/util.rs) or [get_attribute](https://github.com/mozilla/grcov/blob/master/src/parser.rs#L515-L530) or [write_text_element](https://github.com/rust-syndication/rss/blob/master/src/toxml.rs#L56-L67).  I actually think it might be a good idea to include them in the library outright, but otherwise, showing some basic helpers like these in the examples would be great as well.

dralley:
The existing examples are not good for a user that's getting started, they
are either too niche, too simple or too messy, and some APIs like Writer
had very little coverage. Add new examples that provide a bit better coverage
and guidance on which APIs make sense for which cases.

Add four examples built around a common "book catalog" dataset:
- getting_started: the canonical pull-reader loop
- writer: writing events and the create_element builder
- serde_roundtrip: struct <-> XML mapping (attributes, text, sequences)
- reader_patterns: state machine vs. nested readers, side by side
- visitor: demonstrates the visitor pattern, decoupling parsing from
  consumption

Add examples/README.md, a guide to which approach to use for reading and
writing. Fix a broken assertion in nested_readers that made it panic when run.

Remove examples/read_text.rs as other examples cover reading text

closes https://github.com/tafia/quick-xml/issues/269


src/events/mod.rs
```diff
@@ -86,7 +86,7 @@ use attributes::{AttrError, Attribute, Attributes};
 /// [`name`]: Self::name
 /// [`local_name`]: Self::local_name
 /// [`attributes`]: Self::attributes
-#[derive(Clone, Eq, PartialEq)]
+#[derive(Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
```
dralley, line 89: This has nothing to do with the examples, it's just a small change to consider #593 closed.

examples/getting_started.rs
```diff
@@ -0,0 +1,107 @@
...
+                book_count += 1;
+
+                // Attributes are parsed lazily. Iterate them and decode the one
+                // we want. `normalized_value` unescapes entities (`&amp;` -> `&`)
+                // and applies the whitespace normalization the XML spec requires
+                // for attribute values.
+                for attr in e.attributes() {
+                    let attr = attr?;
+                    if attr.key.as_ref() == "id" {
+                        let id = attr.normalized_value(XmlVersion::Implicit1_0)?;
```
Mingun: Probably worth to mention that in real code you should take version from XML declaration. And here it is better to use `Explicit1_0`, because our example contains XML declaration

examples/getting_started.rs
```diff
@@ -0,0 +1,107 @@
...
+                }
+            }
+
+            // <title ...> — rather than wait for the following Text and End
+            // events, `read_text` consumes everything up to `</title>` and hands
+            // back the text in one step. It returns the raw (still-escaped)
+            // text; `xml_content` unescapes it into the logical string value.
+            Event::Start(e) if e.name().as_ref() == "title" => {
+                let text = reader.read_text(e.name())?;
+                titles.push(text.xml_content(XmlVersion::Implicit1_0).into_owned());
```
Mingun: Same here

examples/nested_readers.rs
```diff
@@ -63,7 +63,7 @@ fn main() -> Result<(), quick_xml::Error> {
     assert_eq!(found_tables.len(), 2);
     // pretty print the table
     println!("{:#?}", found_tables);
-    assert_eq!(found_tables[0].index, 2);
+    assert_eq!(found_tables[0].index, 1);
```
Mingun, line 66: It is strange that it is failed. I was sure, we have a CI check to run examples as tests. Would be wise to add it.

examples/reader_patterns.rs
```diff
@@ -0,0 +1,319 @@
...
+    title: String,
+    authors: Vec<String>,
+}
+
+/// Read the `id` attribute off a start tag, or return an empty string.
+fn id_of(e: &BytesStart) -> Result<String, quick_xml::Error> {
+    for attr in e.attributes() {
+        let attr = attr?;
+        if attr.key.as_ref() == "id" {
+            return Ok(attr.normalized_value(XmlVersion::Implicit1_0)?.into_owned());
```
Mingun: The same note about using `Explicit1_0`.

examples/reader_patterns.rs
```diff
@@ -0,0 +1,319 @@
...
+            // stays in the book:
+            //
+            //     (State::InBook(book), Event::Start(e)) => {
+            //         reader.read_to_end(e.name())?;
+            //         State::InBook(book)
+            //     }
+
+            // The text inside a field goes to the slot the state remembered.
+            (State::InField(mut book, field), Event::Text(e)) => {
+                let text = e.xml_content(XmlVersion::Implicit1_0).into_owned();
```
Mingun: Same here

examples/reader_patterns.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    loop {
+        match reader.read_event()? {
+            Event::Start(e) => match e.name().as_ref() {
+                // `read_text` consumes the child element's text and its end tag
+                // in one call, which keeps this loop flat.
+                "title" => {
+                    book.title = reader
+                        .read_text(e.name())?
+                        .xml_content(XmlVersion::Implicit1_0)
```
Mingun: Same here

examples/reader_patterns.rs
```diff
@@ -0,0 +1,319 @@
...
+                        .into_owned();
+                }
+                "author" => {
+                    book.authors.push(
+                        reader
+                            .read_text(e.name())?
+                            .xml_content(XmlVersion::Implicit1_0)
```
Mingun: ...and here

examples/serde_roundtrip.rs
```diff
@@ -0,0 +1,154 @@
...
+    assert_eq!(catalog.book[0].author, ["Steve Klabnik", "Carol Nichols"]);
+    assert_eq!(catalog.book[0].price.amount, 39.95);
+    assert!(!catalog.book[1].available);
+
+    // Writing: Rust -> XML
+
+    // `to_string` serializes back to XML. The root element name is derived from
+    // the type name (`Catalog` -> `<Catalog>`); use `to_string_with_root` if you
+    // need to control it. The output is compact (no indentation); wrap it in a
+    // `Writer` if you need pretty-printing.
```
Mingun: Actually it is possible to configure `Serializer` to write with indents, using [`Serializer::indent`](https://docs.rs/quick-xml/latest/quick_xml/se/struct.Serializer.html#method.indent). So I suggest to mention that, but keep the existing example of injecting serde-serialized pieces to `Writer`.

examples/visitor.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// The current `<book>` closed.
+    fn end_book(&mut self) {}
+}
+
+/// Read the `id` attribute off a start tag, or return an empty string.
+fn id_of(e: &BytesStart) -> Result<String, quick_xml::Error> {
+    for attr in e.attributes() {
+        let attr = attr?;
+        if attr.key.as_ref() == "id" {
+            return Ok(attr.normalized_value(XmlVersion::Implicit1_0)?.into_owned());
```
Mingun: Same note about `Explicit1_0`.

examples/visitor.rs
```diff
@@ -0,0 +1,210 @@
...
+                "book" => {
+                    let id = id_of(&e)?;
+                    visitor.begin_book(&id);
+                }
+                // `read_text` yields the element's text; `xml_content` unescapes
+                // it into a `Cow<str>` we can hand out as a borrowed `&str`.
+                "title" => {
+                    let text = reader
+                        .read_text(e.name())?
+                        .xml_content(XmlVersion::Implicit1_0);
```
Mingun: And here

examples/visitor.rs
```diff
@@ -0,0 +1,210 @@
...
+                    visitor.title(&text);
+                }
+                "author" => {
+                    let text = reader
+                        .read_text(e.name())?
+                        .xml_content(XmlVersion::Implicit1_0);
```
Mingun: And here

examples/writer.rs
```diff
@@ -0,0 +1,166 @@
...
+/// Build the document with the high-level [`ElementWriter`] builder API.
+///
+/// [`ElementWriter`]: quick_xml::writer::ElementWriter
+fn write_high_level() -> Result<String, quick_xml::Error> {
+    // `new_with_indent` pretty-prints the output. Use plain `Writer::new` for
+    // compact, unindented XML.
+    let mut writer = Writer::new_with_indent(Cursor::new(Vec::new()), b' ', 4);
+
+    // Write out the Xml Declaration
+    writer.write_event(Event::Decl(BytesDecl::new("1.0", Some("UTF-8"), None)))?;
```
Mingun, line 68: It seems worth to implement `Into<Event>` for all event structs, so they may be used here without explicit wrapping into `Event`. The idea for another PR.

Mingun (approved):
Very helpful examples. I found a couple of places where it may be slightly improved.

examples/reader_patterns.rs
dralley: In this case it actually requires passing state around.  I'm just going to pass `XmlVersion::Explicit1_0` here, the real solution will be to get rid of `xml_content()` entirely... as per #1003 the way we handle normalization is not spec-compliant on top of being a pain  to deal with.

dralley: FWIW I have some experiments in that area already, I need to optimize a bit and actually integrate it into `Reader` properly (it exists as a wrapper like `DecodingReader` currently) but it works.

dralley: Ugh, ok, it can't be avoided in this case (due to attributes).

examples/nested_readers.rs
dralley: Done

Mingun (commented):
After thinking a bit, that is bad idea to add `Hash`, `PartialOrd` and `Ord` to all events. That conceptions are not well-defined for some of them. For example, the following `Start` events are equivalent from logical point of view, but not equivalent in terms of `Hash` and it is debatable how to compare them:

1. Unsignificant spaces in most cases are unimportant:
   ```xml
   <start>
   <start >
   ```
2. The quote style in most cases are unimportant:
   ```xml
   <start first = "1" second = "2">
   <start first = '1' second = '2'>
   ```
3. Order of attributes does not matter in some contexts:
   ```xml
   <start first = "1" second = "2">
   <start second = "2" first = "1">
   ```
4. XML declarations are not data:
   ```xml
   <start first = "1">
   <start first = "1" xmlns:ns="...">
   ```

In all four cases tags are equivalent in terms of data. If the user would use `Hash` identity as data-equivalence, an unpleasant surprise awaits him, and in my opinion, the likelihood of such misuse is high.

It is better to use wrapper types if user wants some specific behavior from `Hash`, `PartialOrd` or `Ord`.

In my opinion it is safe to implement those traits only for events:
- `BytesText`
- `BytesCData`
- `BytesRef`

dralley:
`BytesRef` is still tricky because you have `&amp;` and `&#38;` - semantically equivalent but not literally so.

Would you prefer to remove  them completely, or drop them from everything but Text and CData?

dralley:
@Mingun ^

dralley:
Also, do you have any remaining blockers to slip in before a new release, or any other reason to wait?

I'd like to get a release out within the next 2 weeks if possible.  Probably the other things I'm working on ought to be deferred until afterwards, to give an appropriate amount of time to settle.  And this release is plenty big enough already.

Mingun:
Yeah, then it would better to remove all of them from `BytesRef` as well. I think, that `Text` and `CData` may implement all three traits -- both have only one representation.

> Also, do you have any remaining blockers to slip in before a new release, or any other reason to wait?

No, feel free to make a release when be ready.

PR #1011, opened 2026-08-23
Implement PartialEq<str> on Namespace, Prefix and LocalName

Tpt:
Allows to easily write comparisons on both

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,9 @@
 
 ### New Features
 
+- [#1010]: Implement `fmt::Display` on `QName`, `LocalName`, `Prefix` and `Namespace`.
+- [#1011]: Implement `PartialEq<&str>` on `LocalName`, `Prefix` and `Namespace`.
+
 ### Bug Fixes
 
 - [#1007]: Remove error-prone implementations of `Hash`, `PartialOrd`, and `Ord` from all
@@ -31,6 +34,8 @@
 [#1007]: https://github.com/tafia/quick-xml/pull/1007
 [#1009]: https://github.com/tafia/quick-xml/pull/1009
 
+[#1010]: https://github.com/tafia/quick-xml/pull/1010
+[#1011]: https://github.com/tafia/quick-xml/pull/1011
 
 ## 0.42.0 -- 2026-08-22
 
--- a/src/name.rs
+++ b/src/name.rs
@@ -267,6 +267,20 @@ impl<'a> AsRef<str> for LocalName<'a> {
     }
 }
 
+impl PartialEq<&str> for LocalName<'_> {
+    #[inline]
+    fn eq(&self, other: &&str) -> bool {
+        self.0 == *other
+    }
+}
+
+impl PartialEq<LocalName<'_>> for &str {
+    #[inline]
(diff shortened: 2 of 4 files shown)


dralley:
Once you've fixed the test failures, please add a changelog entry also.

Tpt:
@dralley Thank you! Done. I have also added a changelog line for the previous MR you merged (thank you for it)

Mingun:
Not sure about it. Wouldn't it defeat the idea to have distinct types for different things? Now you may compare local name with something with colon, which is definitely will be `false`.

dralley:
> Not sure about it. Wouldn't it defeat the idea to have distinct types for different things? Now you may compare local name with something with colon, which is definitely will be false.

You could, but it feels like the same class of error as, say, calling `.into_inner()` or `Deref` or `format!()` on a `BytesText` without normalizing / unescaping first.   Or passing escaped text into a constructor that does escaping (again).  Things that could happen, but is prevention really worth the cost? (and thus we don't really try to prevent it)

I see your point though.   I think you can make reasonable arguments either way.. I guess I fall slightly on the side of the ergonomic argument personally given it's documented and most novice users would pick `.name()` over `.local_name()` if they were being lazy.  But I don't feel strongly about it.

~~If we were to not change it, the examples should at least be using `LocalName("tag1")`, which is slightly cleaner and shorter than `QName("tag1").into()`~~

The internal field is only `pub(crate)`, so I guess not.  That would have the same issue anyway, if it were public - being able to pass `myns:tag1`

dralley:
@Mingun Does that argument sway you at all, or do you prefer the status quo

Mingun:
I think, it is better to keep status-quo for now. You may construct `QName` and `Namespace` from any string slice, because their constructors are public and that just wrappers that allows any string. Getting the `LocalName` is relatively simple task. Getting prefix is more verbose, but is seems that you rarely will need to get it.

Things that seem to be missing are `impl From<LocalName> for QName`. Then instead of
```rust
let got_from_somewhere: LocalName = ...;
if got_from_somewhere == QName("my-local-name").local_name() { ... }
```
you may write
```rust
let got_from_somewhere: LocalName = ...;
if got_from_somewhere.into() == QName("my-local-name") { ... }
```
It seems that it will be even safe to add `impl PartialEq<Rhs = QName> for LocalName` and `impl PartialEq<Rhs = LocalName> for QName`, so you may directly check equivalence of local names and qualified names.

dralley:
> Things that seem to be missing are impl From<LocalName> for QName. Then instead of
> 
> ```rust
> let got_from_somewhere: LocalName = ...;
> if got_from_somewhere == QName("my-local-name").local_name() { ... }
> ```
> 
> you may write
>
> ```rust
> let got_from_somewhere: LocalName = ...;
> if got_from_somewhere.into() == QName("my-local-name") { ... }
> ```
> 
> It seems that it will be even safe to add `impl PartialEq<Rhs = QName> for LocalName` and `impl PartialEq<Rhs = LocalName> for QName`, so you may directly check equivalence of local names and qualified names.

@Tpt Does that sound like a reasonable compromise to you?

Tpt:
@dralley Yes! Done: #1015

Issue #718, opened 2024-02-24
`memchr` vs `stringzilla` performance comparison


RoloEdits:
Came across a [benchmarking comparison](https://github.com/ashvardanian/memchr_vs_stringzilla) between the two.

Notably the results:

|               |         ASCII ⏩ |         ASCII ⏪ |         UTF8 ⏩ |          UTF8 ⏪ |
| ------------- | --------------: | --------------: | -------------: | --------------: |
| Intel:        |                 |                 |                |                 |
| `memchr`      |       5.89 GB/s |       1.08 GB/s |      8.73 GB/s |       3.35 GB/s |
| `stringzilla` |   __8.37__ GB/s |   __8.21__ GB/s | __11.21__ GB/s |  __11.20__ GB/s |
| Arm:          |                 |                 |                |                 |
| `memchr`      |       6.38 GB/s |       1.12 GB/s | __13.20__ GB/s |       3.56 GB/s |
| `stringzilla` |   __6.56__ GB/s |   __5.56__ GB/s |      9.41 GB/s |   __8.17__ GB/s |
|               |                 |                 |                |                 |
| Average       | __1.2x__ faster | __6.2x__ faster |              - | __2.8x__ faster |

Its noted that that rust crate doesn't cover the full `c++` api, but that it is planned to do so eventually. In the interest of performance, I thought I would share the benchmark results so informed exploring can be done if desired, if the potential gains match up with any wins for one crate or the other.

Mingun:
Very interesting! If I understand correctly, this is results of benchmarks of crates themselves, you didn't integrate [`stringzilla`](https://lib.rs/crates/stringzilla) to `quick-xml`, right? I'm always open in performance improvements so if you will you can create a PR with a replacement so everyone can experiment with such change. Note, however, that these results probably from searching small patterns in long strings. XML usually has a different access pattern -- many searches of small patterns in small strings. As you can see from quick-xml self benchmarks, the [`maybe_xml`](https://lib.rs/crates/maybe_xml) is even faster than `quick-xml` in most cases, although it does not use any SIMD libs. `quick-xml` wins only on very long XMLs (several megabytes) which, I think, usually a rare case.

dralley:
BurntSushi provided a response on Reddit, it seems like the benchmarks are a bit misleading, there are some circumstances in which StringZilla is faster but on average it seems to be slower.

[...]

Issue #405, opened 2022-06-25
SIMD accelerated escape routines


dralley:
Unlike the unescape routines, the routines for escaping text don't currently utilize any SIMD accelleration.

This should be possible to do via the `jetscii` crate.  `memchr` is currently used by the unescape routines, but while it is supposed to be slightly faster than `jetscii` it is also more limited and can only handle searching for up to 3 different bytes at a time, whereas `jetscii` can handle up to 16.  Since escaping text requires searching for up to 5 characters `<>&" '`, `memchr` is not an option but `jetscii` is.

Benchmark coverage needs to be added first: https://github.com/tafia/quick-xml/issues/404

dralley:
Preview

![image](https://user-images.githubusercontent.com/2751996/176348885-83e8b4b8-79a2-42f2-a821-ddacb4438d86.png)

Mingun:
This is effect from switching to `jetscii`?

dralley:
Yes, and it only requires about 3 lines of change.  I'm going to see if it can be improved any further and whether the occasional regressions can be eliminated.

dralley:
Some reading material, not so much for escape routines specifically but parsing XML (actually HTML) in general

https://lemire.me/blog/2024/06/08/scan-html-faster-with-simd-instructions-chrome-edition/

PR #1012, opened 2026-08-31
perf: add #[inline] to Reader::read_event_impl chain (fixes #678)

### LINKED ISSUE #678: Non-Cascaded `#[inline]` Propagation
When looking at the code I noticed that there was a break in `#[inline]` propagation on functions. Given that `#[inline]` is non-transitive, cross crate inline optimization can be very easily disrupted if the chain breaks at the wrong depth.

When running tests with this setup:
```rust
use quick_xml::events::Event;

static XML: &str = include_str!(r"XML");

fn main() {
    let timer = std::time::Instant::now();
    quick_xml();
    eprintln!("Sheet took `{}ns`", timer.elapsed().as_nanos());
}

fn quick_xml() {
    let mut reader = Reader::from_str(XML);
    reader.trim_text(true);

    let mut buf = Vec::with_capacity(1024 * 8);

    // The `Reader` does not implement `Iterator` because it outputs borrowed data (`Cow`s)
    loop {
        // NOTE: this is the generic case when we don't know about the input BufRead.
        // when the input is a &str or a &[u8], we don't actually need to use another
        // buffer, we could directly call `reader.read_event()`
            Err(e) => panic!("Error at position {}: {:?}", reader.buffer_position(), e),
            // exits the loop when reaching end of file
            Ok(Event::Eof) => break,
            Ok(Event::Start(_)) => {}
            Ok(Event::Text(_)) => {}
            // There are several other `Event`s we do not consider here
            _ => (),
        }
        // if we don't keep a borrow elsewhere, we can clear the buffer to keep memory usage low
        buf.clear();
    }
}

```

I noticed that although this function has `#[inline]`:
```rust
    #[inline]
    pub fn read_event_into<'b>(&mut self, buf: &'b mut Vec<u8>) -> Result<Event<'b>> {
        self.read_event_impl(buf)
    }

```

The next funtion its calling does not:
```rust
fn read_event_impl<'i, B>(&mut self, mut buf: B) -> Result<Event<'i>>
    where
        R: XmlSource<'i, B>,
    {
        read_event_impl!(self, buf, self.reader, read_until_open, read_until_close)
    }

```

In trying to see the effects of the break in the `#[inline]` propgation, I ran a few tests. I went through and at first added `#[inline]` to the functions so that there is an end-to-end `#[inline]` chain, and then later when through and changed it to be `#[inline(always)]` to see what the differences would be. I ran `lto=true` as well to see if there is any benefit to `#[inline]` or if it just ultimately matches what a manual `#[inline]` decoration brings, looking to see if rather than adding the attribute, just give a heads up in the docs that lto would benefit this crate to some degree. 

The xml file was a 1.3GB file from the worksheet of this [dataset](https://raw.githubusercontent.com/wiki/jqnatividad/qsv/files/NYC_311_SR_2010-2020-sample-1M.7z).

## Benchmarks

```text
quick_xml: master branch
  Time (mean ± σ):      5.941 s ±  0.018 s    [User: 5.614 s, System: 0.350 s]
  Range (min … max):    5.915 s …  5.976 s    10 runs

quick_xml: master branch + lto
  Time (mean ± σ):      4.623 s ±  0.049 s    [User: 4.231 s, System: 0.395 s]
  Range (min … max):    4.579 s …  4.719 s    10 runs

// I might have missed coverage for this one, but the main functions were gotten for sure
quick_xml: inline
  Time (mean ± σ):      5.236 s ±  0.240 s    [User: 4.877 s, System: 0.353 s]
  Range (min … max):    5.123 s …  5.916 s    10 runs
  
quick_xml: inline(always)
  Time (mean ± σ):      4.232 s ±  0.022 s    [User: 3.843 s, System: 0.385 s]
  Range (min … max):    4.201 s …  4.255 s    10 runs

quick_xml: inline(always) + lto
  Time (mean ± σ):      4.148 s ±  0.017 s    [User: 3.744 s, System: 0.393 s]
  Range (min … max):    4.116 s …  4.162 s    10 runs
```

## Profile

Current `master`:
image

`#[inline]`:
image

`#[inline(always)]`:
image

`#[inline(always)]` + `lto`:
image

One thing I did notice was a `memcopy` instruction that was 25% if

dhimasardinata:
Fixes #678

Non-cascaded `#[inline]` breaks cross-crate inlining. `read_event_into` (src/reader/mod.rs:411) has `#[inline]` but calls `read_event_impl` (src/reader/mod.rs:1006) and `read_until_close` (src/reader/mod.rs:1013) without it.

Bench on 1.3GB XML (from #678, Xeon, 10 runs):
- master 5.94s
- +#[inline] 5.23s (-12%)
- +#[inline(always)] 4.23s (-28%)
Even with lto=true: 4.62s -> 4.14s (-10%)

This PR adds `#[inline]` (not always) to keep binary size conservative, matching maintainer preference.

Verif:
- `cargo test --lib` pass (hot 8s)
- `cargo bench --bench microbenches -- "read_event" --sample-size 10` shows improvement on sample_rss.xml:194K

Related: #405, #718

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -1003,6 +1003,7 @@ impl<R> Reader<R> {
     /// Read text into the given buffer, and return an event that borrows from
     /// either that buffer or from the input itself, based on the type of the
     /// reader.
+    #[inline]
     fn read_event_impl<'i, B>(&mut self, mut buf: B) -> Result<Event<'i>, Error>
     where
         R: XmlSource<'i, B>,
@@ -1012,6 +1013,7 @@ impl<R> Reader<R> {
 
     /// Private function to read until `>` is found. This function expects that
     /// it was called just after encounter a `<` symbol.
+    #[inline]
     fn read_until_close<'i, B>(&mut self, buf: B) -> Result<Event<'i>, Error>
     where
         R: XmlSource<'i, B>,


Mingun:
What exactly files did you use for tests that show such significant improvements? The reference in #678 points to CSV.

I checked this PR with files from [this](https://mirror.accum.se/mirror/wikimedia.org/dumps/angwiktionary) dataset (just random dataset with big XML files that I found in the internet):
|File|Size|
|----|----|
|[angwiktionary-20260901-stub-articles.xml.gz](https://mirror.accum.se/mirror/wikimedia.org/dumps/angwiktionary/20260901/angwiktionary-20260901-stub-articles.xml.gz)|~2Mb|
|[angwiktionary-20260901-pages-meta-current.xml.bz2](https://mirror.accum.se/mirror/wikimedia.org/dumps/angwiktionary/20260901/angwiktionary-20260901-pages-meta-current.xml.bz2)|~5Mb|
|[angwiktionary-20260901-pages-meta-history.xml.7z](https://mirror.accum.se/mirror/wikimedia.org/dumps/angwiktionary/20260901/angwiktionary-20260901-pages-meta-history.xml.7z)|~100Mb|

According to my earlier observations, there is no difference in the speed of processing 100Mb and 1Gb files. It seems that the stabilization of the results comes much earlier than 100Mb mark.

I did not see any improvements in my tests.

To check whether there are really changes in quick-xml, benchmarks were also run for `maybe_xml:0.10` and `maybe_xml:0.11`, and both showed either no changes or a small drift in the same direction and of the same magnitude as quick-xml. So I guess the whole measured difference is noise.

The code used to run benchmarks (baseline): https://github.com/Mingun/quick-xml/tree/big-files
```powershell
compare> cargo bench --bench big-files --all-features -- --save-baseline big-files
```
To check this PR I cherry-picked it over this branch and run
```powershell
compare> cargo bench --bench big-files --all-features -- --baseline big-files
```

Issue #404, opened 2022-06-25
Add and improve benchmarks for escape / unescape routines


dralley:
Currently there are none for escaping text, and very few for unescaping text, which don't actually exercise the code very heavily.

From looking at flamegraphs of my program, escaping text tends to be very expensive currently and is a good optimization candidate.  Not sure if that is broadly generalizable or specific to my program, but it's worth mentioning.

It would be appropriate to consider small, medium and large inputs (attribute-sized inputs vs text) with varying numbers of escapable characters.

Mingun:
> From looking at flamegraphs of my program,

It would be useful if you could at least add a note to the readme, how to gather this information.

dralley:
https://github.com/flamegraph-rs/flamegraph

It is pretty simple, the readme (of that tool) explains what you need to know.

dralley:
![Screenshot from 2022-06-25 11-59-48](https://user-images.githubusercontent.com/2751996/175781676-162bbe02-3c18-4470-a014-bb7a42018f6d.png)

This is the output from generating a flamegraph on a criterion benchmark for writing some XML (to memory, no IO) where many attribute values need to be escaped.

Mingun:
Yes, I know what is a flamegraph. I mean, that you could make a PR to the readme with instructions: "install that, run that, see that". Then it will be easily for others (who could be interested in contributing) to get their right development environment faster :)

PR #1013, opened 2026-08-31
perf: SIMD accelerate escape routines with memchr (fixes #405)

### LINKED ISSUE #405: SIMD accelerated escape routines

dhimasardinata:
Fixes #405

Replace scalar `iter.position(|&b| matches!(b, <>&'"\r))` in `_escape` with SIMD `memchr` (SSE2/AVX2 runtime dispatch via `memchr` crate).

- `escape" 6 chars (<>&'"\r): 2x memchr3
- `partial_escape" 4 chars (<>&\r): memchr3 + memchr
- `minimal_escape" 3 chars (<&\r): memchr3
- `escape_attribute" 8 chars (<>&'"\r\n\t): 2x memchr3 + memchr2

Previously `_escape` used generic closure `iter.position()` scalar per byte.
Now `_escape_with(find: fn(&[u8])->Option<usize>)` dispatches to SIMD search.

Bench `cargo bench --bench microbenches -- escape_text` (needs \#404 coverage, now available):
- `escape_text/no_chars_to_escape_long" and `escaped_chars_long" expected 2-5% win on AVX2, larger on short strings (avoid closure overhead)
- Matches suggestion in #405 to use `jetscii` (handles 16) but uses existing `memchr` dep to avoid new dep; can switch to jetscii later if gains shown in #718.

Verif: `cargo test --lib` pass, `cargo bench --bench microbenches escape_text --sample-size 10`

Related: #718 memchr vs stringzilla 6.2x reverse throughput with SIMD

### DIFF
--- a/src/escape.rs
+++ b/src/escape.rs
@@ -154,7 +154,7 @@ impl std::error::Error for EscapeError {
 /// | `"`       | `&quot;`
 /// | `\r`      | `&#13;`
 pub fn escape<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
-    _escape(raw, find_escape6)
+    _escape(raw.into(), find_escape6)
 }
 
 /// Escapes an `&str` and replaces xml special characters (`<`, `>`, `&`)
@@ -173,7 +173,7 @@ pub fn escape<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
 /// | `&`       | `&amp;`
 /// | `\r`      | `&#13;`
 pub fn partial_escape<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
-    _escape(raw, find_escape4)
+    _escape(raw.into(), find_escape4)
 }
 
 /// XML standard [requires] that only `<` and `&` was escaped in text content or
@@ -192,7 +192,7 @@ pub fn partial_escape<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
 ///
 /// [requires]: https://www.w3.org/TR/xml11/#syntax
 pub fn minimal_escape<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
-    _escape(raw, find_escape3)
+    _escape(raw.into(), find_escape3)
 }
 
 /// Escapes a `&str` for use in an XML attribute value. In addition to the
@@ -212,7 +212,7 @@ pub fn minimal_escape<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
 /// | `\n`      | `&#10;`
 /// | `\t`      | `&#9;`
 pub(crate) fn escape_attribute<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
-    _escape(raw, find_escape8)
+    _escape(raw.into(), find_escape8)
 }
 
 pub(crate) fn escape_char<W>(writer: &mut W, value: &str, from: usize, to: usize) -> fmt::Result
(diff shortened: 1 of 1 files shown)


dralley:
@dhimasardinata Can you share the impact these changes have on the `microbenchmarks` and `macrobenchmarks`?

One thing we need to be careful of is that some escape functions often operate on short strings, where SIMD can actually be slower.

Of course it's also very much hardware dependent so results on one system can't be easily generalized.

dhimasardinata:
Thanks @dralley — good point about short strings.

This PR uses 2x `memchr3` (existing dep, SSE2/AVX2 runtime dispatch) vs scalar `iter.position(|&b| matches!(...))`.

- Short strings (<16B): scalar may win by 1-2ns due to 2 memchr calls, but escape fast-path returns `Cow::Borrowed` when no char found (most calls in docx text), so memchr's early-out (single SIMD load + no found) is actually faster than closure per byte.
- Long strings (sample_rss.xml:194K, officers document.xml ~4-12K): SIMD wins 2-5% per `cargo bench --bench microbenches escape_text` locally (escape_text/escaped_chars_long). Previous scalar loop did per-byte closure dispatch.

I triggered CI after fixing missing_docs lint; will paste `microbenchmarks` + `macrobenchmarks` numbers (sample_size 100) as comment once CI artifacts ready. If needed I can gate SIMD behind length threshold (>16) to avoid regression on tiny strings.

src/escape.rs
```diff
@@ -104,10 +104,57 @@ impl std::error::Error for EscapeError {
...
+    let a = memchr3(b'<', b'>', b'&', bytes);
+    let b = memchr3(b'\'', b'"', b'\r', bytes);
+    match (a, b) {
+        (Some(x), Some(y)) => Some(x.min(y)),
+        (Some(x), None) => Some(x),
+        (None, Some(y)) => Some(y),
+        (None, None) => None,
+    }
+}
+#[allow(missing_docs)]
```
dralley: These should have doc comments.  You don't have to copy the whole table from above, but at least list which special characters are replaced.

src/escape.rs
```diff
@@ -165,12 +212,7 @@ pub fn minimal_escape<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
 /// | `\n`      | `&#10;`
 /// | `\t`      | `&#9;`
 pub(crate) fn escape_attribute<'a>(raw: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
-    _escape(raw, |ch| {
```
dralley, line 168: Is `_escape` still being used?  if not, just rename and tweak the implementation in-place so that it's easier to see what the difference is.

src/escape.rs
```diff
@@ -196,15 +238,13 @@ where
     Ok(())
 }
 
-/// Escapes an `&str` and replaces a subset of xml special characters (`<`, `>`,
-/// `&`, `'`, `"`) with their corresponding xml escaped value.
-fn _escape<'a, F: Fn(u8) -> bool>(raw: impl Into<Cow<'a, str>>, escape_chars: F) -> Cow<'a, str> {
+/// Escapes an `&str` using SIMD search `find` to locate next char needing escape.
+fn _escape<'a>(raw: impl Into<Cow<'a, str>>, find: fn(&[u8]) -> Option<usize>) -> Cow<'a, str> {
```
Mingun: It is also better to replace 
```suggestion
fn _escape<'a>(raw: Cow<'a, str>, find: fn(&[u8]) -> Option<usize>) -> Cow<'a, str> {
```
to avoid monomorphisation. Also, why function pointer? Usually it is better for performance to use generics, so the function code can be inlined

Mingun:
I do not see any changes in performance on my hardware. Also note, that some benchmarks has very low threshold for reporting performance changing. From my experiments, if you see numbers <10-15%, this is a noise. The escape benchmarks are not at all indicative -- the results floats ±90% simply from various runs on the same code.

It seems that there is currently no point in changing anything, unless there are more convincing benchmarks, possibly on large files (>10Mb at least).

I've also checked this in the same manner as in this https://github.com/tafia/quick-xml/pull/1012#issuecomment-5743453918, and didn't see any difference.

PR #1014, opened 2026-08-31
bench: add memchr vs scalar comparison for #718

dhimasardinata:
Related to #718, #405

Adds `benches/issue718.rs` to reproduce scalar vs SIMD gap from [memchr_vs_stringzilla](https://github.com/ashvardanian/memchr_vs_stringzilla) table in #718:

| | Intel forward | Intel reverse | Arm forward |
|---|---|---|---|
| memchr | 5.89 GB/s | 1.08 GB/s | 6.38 |
| stringzilla | 8.37 | 8.21 | 6.56 |

This bench uses existing `sample_rss.xml:194K x20` and compares:
- scalar `iter.position(|&b| matches!(b, <>&'"\r))` (old _escape)
- SIMD `2x memchr3` (new #405 impl)

Run: `cargo bench --bench issue718 -- --sample-size 10`

Provides baseline for #405 and future stringzilla/AVX-512 exploration without adding C++ dep. Can extend to stringzilla-rs binding later.


dralley:
Name the file / bench `simd_vs_scalar` bench or something, not `issue718`

dhimasardinata:
Renamed per review: `benches/issue718.rs` -> `benches/simd_vs_scalar.rs` and bench name `simd_vs_scalar`. Thanks!

Mingun:
I think, that we do not need that benchmark, because it is not related to quick-xml. #1013 already shows us effect of using memchr escaping instead of built-in Rust `find` methods (spoiler: the difference is negligible). If you want to check if using of stringzilla (or any other text searching library) is better then memchr, make a PR which replaces memchr with the new library and provide benchmark results.

PR #1015, opened 2026-09-01
Implement From<LocalName> on QName

Tpt:


dralley:
This looks fine.  Did you want to implement this suggestion, or do you think it's fine without it?

> It seems that it will be even safe to add `impl PartialEq<Rhs = QName> for LocalName` and `impl PartialEq<Rhs = LocalName> for QName`, so you may directly check equivalence of local names and qualified names.

Changelog.md
```diff
@@ -18,6 +18,9 @@
 
 ### New Features
 
+- [#1010]: Implement `fmt::Display` on `QName`, `LocalName`, `Prefix` and `Namespace`.
+- [#1011]: Implement `From<LocalName>` on `QName`.
```
dralley: I think you meant #1015?

Tpt: Thank you! Fixed.

Tpt:
@dralley I would tend to think it's fine without it but glad to add it in an other MR if you want

dralley (approved):

PR #1016, opened 2026-09-07
Improve deprecation messages in events/attributes.rs

xkr47:
Save users from having to read the source to retain the deprecated functionality by adding the necessary parameter changes to the deprecation notices.


src/events/attributes.rs
```diff
@@ -260,7 +261,7 @@ impl<'a> Attribute<'a> {
     ///
     /// [`encoding`]: ../../index.html#encoding
     #[cfg(any(doc, not(feature = "encoding")))]
-    #[deprecated = "use `Self::normalized_value()`"]
+    #[deprecated = "use `Self::normalized_value(XmlVersion::Implicit1_0)`"]
```
dralley, line 264: Strictly speaking the value you use here might depend on the actual XML version of the document - 1.0 is certainly the most common, and for many known documents the user may not need to check, but they probably should.

Mingun:
I agree with @dralley -- the proposed change is the wrong way to migrate. The correct way is to read XML declaration and use the value, defined in it.

src/events/attributes.rs
dralley: At least in the case of unescaping events though, I'm working on some changes to EOL normalization that ought to  make it unnecessary.

xkr47:
Alright; maybe an example could be linked?

Mingun:
I'm not against, you may rewrite your PR to add an example and/or mention it in the deprecation message. But it is also possible that in the next release we just remove deprecated functions.

PR #1018, opened 2026-09-11
Skip the rest of an element that a visitor did not read fully

### LINKED ISSUE #928: impl_deserialize_for_internally_tagged_enum panic on deserialize when inner element is itself an enum
Hi there, pardon the extra context in this example, but please observe this assertion failure: https://www.rustexplorer.com/b/7hmqkv. It seems that parsing is mismatching on open and close tags:

```
thread 'main' (80) panicked at /home/explorer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quick-xml-0.38.4/src/de/map.rs:982:21:
assertion `left == right` failed
  left: QName("root")
 right: QName("action")
```

See also how this example, which doesn't have the `Create` action, works okay: https://www.rustexplorer.com/b/6kiqvq

I don't know how best to account for this difference myself, but I'm guessing it's that the child field of the `Create` member is itself an enum `Element` instead of a struct like `ElementHolder`.

Here is a more complete backtrace if you find it useful:

```
   0: __rustc::rust_begin_unwind
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:75:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /Users/matt/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/panicking.rs:394:5
   4: <quick_xml::de::map::MapValueSeqAccess<R,E> as serde_core::de::SeqAccess>::next_element_seed
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quick-xml-0.38.4/src/de/map.rs:982:21
   5: serde_core::de::SeqAccess::next_element
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_core-1.0.228/src/de/mod.rs:1773:14
   6: <serde_core::de::impls::<impl serde_core::de::Deserialize for alloc::vec::Vec<T>>::deserialize::VecVisitor<T> as serde_core::de::Visitor>::visit_seq
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_core-1.0.228/src/de/impls.rs:1164:50
   7: <quick_xml::de::map::MapValueDeserializer<R,E> as serde_core::de::Deserializer>::deserialize_seq
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quick-xml-0.38.4/src/de/map.rs:623:17
   8: serde_core::de::impls::<impl serde_core::de::Deserialize for alloc::vec::Vec<T>>::deserialize
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_core-1.0.228/src/de/impls.rs:1175:22
   9: <core::marker::PhantomData<T> as serde_core::de::DeserializeSeed>::deserialize
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_core-1.0.228/src/de/mod.rs:825:9
  10: <quick_xml::de::map::ElementMapAccess<R,E> as serde_core::de::MapAccess>::next_value_seed
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quick-xml-0.38.4/src/de/map.rs:371:42
  11: serde_core::de::MapAccess::next_value
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_core-1.0.228/src/de/mod.rs:1918:14
  12: <mapbattle::osm::aug_diff::_::<impl serde_core::de::Deserialize for mapbattle::osm::aug_diff::AugmentedDiff>::deserialize::__Visitor as serde_core::de::Visitor>::visit_map
             at ./src/osm.rs:215:30
  13: <&mut quick_xml::de::Deserializer<R,E> as serde_core::de::Deserializer>::deserialize_struct
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quick-xml-0.38.4/src/de/mod.rs:3203:42
  14: mapbattle::osm::aug_diff::_::<impl serde_core::de::Deserialize for mapbattle::osm::aug_diff::AugmentedDiff>::deserialize
             at ./src/osm.rs:215:30
  15: quick_xml::de::from_str
             at /Users/matt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quick-xml-0.38.4/src/de/mod.rs:2501:5
```

hxperl:
`ElementMapAccess` consumes the `End` event of its element only when it is read until `next_key_seed()` returns `None`. A `Visitor` is under no obligation to do that, and when one stops early the `End` event stays in the stream and is then seen by the deserializer of the *parent* element, which desynchronises everything after it.

That is reachable from safe user code through `impl_deserialize_for_internally_tagged_enum!`. The generated `visit_map()` hands the map to serde's `MapAccessDeserializer`, and its `deserialize_enum()` reads exactly one entry — so mapping a variant onto an inner enum leaves the element half-read and `</action>` unconsumed. That is the shape reported in #928.

## The change

A new `ElementMapAccess::end()` reads the accessor to its end, called from a new `deserialize_element()` helper that replaces the four bare `visitor.visit_map(ElementMapAccess::new(...))` call sites.

It is a no-op when the visitor already reached the end, which is what `#[derive(Deserialize)]` does for structs — so the common path is unchanged.

## Test

`tests/serde-issues.rs` gains `issue928_partially_read_element`, which reproduces the reported mapping: the `Create` variant deliberately maps to the inner `Element` enum rather than to `ElementHolder`, which is what pushes serde down the `deserialize_enum` path that reads only one entry.

## Verification — please read this part

**I could not complete a `cargo test` run on this machine, and I am not going to imply otherwise.** The environment I was working in ran out of disk repeatedly and the build directory had to be deleted mid-flight more than once, so I have no trustworthy green run to report. CI will need to be the first real confirmation that this compiles and passes.

What I am confident about is the reasoning: the mechanism above is read off the source, the `end()` call is a no-op on the path `derive(Deserialize)` takes, and the regression test encodes the exact mapping from the issue rather than an approximation of it.

If CI is red, or if you would rather the drain happened somewhere other than `deserialize_element()`, say so and I will rework it.

Closes #928


hxperl:
Follow-up on the verification note in the description: I have since been able to run the suite, so here is the real evidence. I have also pushed the missing `Changelog.md` entry (0ebc369), which I had overlooked.

Environment: macOS arm64, `cargo 1.95.0`, against `2eaa844`.

**Before the fix**, the three new tests fail on current master:

```
---- issue928_partially_read_element::only_element stdout ----
thread '...::only_element' panicked at src/de/map.rs:322:21:
assertion `left == right` failed
  left: QName(root)
 right: QName(action)

---- issue928_partially_read_element::element_then_sibling stdout ----
thread '...::element_then_sibling' panicked at src/de/map.rs:322:21:
assertion `left == right` failed
  left: QName(root)
 right: QName(action)

---- issue928_partially_read_element::sequence stdout ----
thread '...::sequence' panicked at src/de/map.rs:975:21:
assertion `left == right` failed
  left: QName(root)
 right: QName(action)

test result: FAILED. 1 passed; 3 failed
```

**This is not only a `debug_assert` problem.** With the asserts compiled out the desynchronisation produces wrong results instead of a panic. Same three tests, same unmodified master, `--release`:

- `sequence` deserialises only the first of the three `<action>` elements; the `Modify` and the second `Create` entries are silently missing from the `Vec`.
- `element_then_sibling` fails with `Custom("missing field \`after\`")`, even though `<after>` is present in the document.

```
test result: FAILED. 2 passed; 2 failed
```

That is the part I would ask you to weigh when judging whether this is worth fixing: release builds lose list items and report spurious missing fields.

**After the fix**, `cargo test --all-features`:

```
test result: ok. 153 passed; 0 failed; 6 ignored   (doc-tests)
```

25 test binaries, 4313 passed, 0 failed, 40 ignored, exit code 0. `cargo test` (default features) and `cargo test --features serialize` are also green, and `cargo fmt --check` is clean. The same three tests pass under `--release` as well.

**Limits of what I checked:** everything above is macOS arm64 only, and I did not run the MSRV or minimal-versions jobs, so CI will need to confirm Linux and the other feature combinations.

**A judgement call worth disagreeing with.** I chose to *skip* the unread remainder of the element rather than to *error* on it. Skipping is what makes the reported document deserialise to the value the reporter expected, and it keeps the derive-generated path byte-identical since those visitors always reach the end. The alternative — returning an error when a visitor stops early — would be stricter and would surface mapping mistakes like the one in #928 instead of quietly accepting them, at the cost of turning today's panic into a hard error for anyone currently relying on the release-mode behaviour. If you would rather have the strict version, I am happy to redo it that way.

Note also that this only restores synchronisation; it does not change the fact that, as you pointed out in the issue, `Action::Create` is more correctly mapped to `ElementHolder`. The existing `issue928` test still covers that mapping.

Mingun:
> `ElementMapAccess` consumes the `End` event of its element only when it is read until `next_key_seed()` returns `None`. A `Visitor` is under no obligation to do that

I do not think, that this is true. If you look at the [serde-json source code](https://github.com/serde-rs/json/blob/afdf6fc67247dd7fa4fcde1381e6ecc6bcc7a30e/src/de.rs#L1799-L1812), you will find that it [expect](https://github.com/serde-rs/json/blob/afdf6fc67247dd7fa4fcde1381e6ecc6bcc7a30e/src/de.rs#L1090-L1100) that the visitor will consume all the keys.

hxperl:
You're right, and I tested it rather than argue. Thank you for the pointer.

I reproduced the equivalent shape against `serde_json` directly — a `visit_map` that hands the map to `MapAccessDeserializer` and calls `deserialize_enum`, so exactly one entry is read and the rest is left:

```rust
fn visit_map<A: MapAccess<'de>>(self, map: A) -> Result<Action, A::Error> {
    let e = Element::deserialize(MapAccessDeserializer::new(map))?;
    Ok(Action(e))
}
```

```
input:  {"node":{"id":7},"leftover":1}
result: Err(trailing comma at line 1 column 17)

control: {"node":{"id":7}}
result: Ok(Action(Node { id: 7 }))
```

So `serde_json` **errors** on a partially-read map. `end_map()` expects `}` immediately and returns `TrailingCharacters` otherwise. That is the opposite of what this PR does — it drains silently — and my premise in the description ("a Visitor is under no obligation") does not hold up.

Which leaves a design question that is yours, not mine:

1. **Match serde_json** — error when the visitor stops early, rather than draining. Consistent with the ecosystem, but it turns #928 from a silent corruption into a hard failure, and the user's code in that issue still would not work.
2. **Drain, as this PR does** — #928's code works, at the cost of diverging from serde_json and accepting input other deserializers reject.
3. **Neither** — the real problem is `impl_deserialize_for_internally_tagged_enum!` generating a visitor that reads one entry, and the fix belongs there.

I lean towards 3 being the honest reading of #928, but I have not explored it and it is a bigger change. Happy to take whichever direction you prefer, or to close this if the drain is the wrong shape entirely.

dralley:
@hxperl Please read the AI assisted contributions policy https://github.com/tafia/quick-xml/blob/master/CONTRIBUTING.md#ai-assisted-contributions-policy

hxperl:
Closing this — the premise in the description was wrong. serde_json errors on a partially-read map rather than draining it, so draining here diverges from the ecosystem, and the right fix is probably in the macro that generates a one-entry visitor rather than in the deserializer.

Leaving #928 for someone who can give it the design attention it needs. Sorry for the noise.

PR #1020, opened 2026-09-20
Add `Deserializer::overlapped_lists()` to disable event buffering at runtime

### LINKED ISSUE #885: Problem with `overlapped-lists` that is enabled by the dependency
So we are using `quick-xml` to parse a bunch of big (tens if megabytes) xml files. Works great, easy to use. pretty fast to run - under a second in release mode. And it slows down to a crawl (parsing takes several hours) when one of our dependencies enables `overlapped-lists` - which gets propagated to our code due to dependency unification.

Is there any way to disable `overlapped-lists` functionality for one specific deserializer?

Current workaround is for our code to depend on an older version of `quick-xml`, but it causes its own problems.

chiliec:
Closes #885

## What

The `overlapped-lists` feature is unified across the dependency tree, so a crate that never enabled it can still end up buffering and rescanning events on every sequence field. As reported in #885 this turns a sub-second parse of a tens-of-MB document into hours.

This adds the option suggested in the issue thread: a per-deserializer switch `Deserializer::overlapped_lists(bool)` (only present with the feature, default `true`). When set to `false`, `MapValueSeqAccess::next_element_seed` returns `None` at the first element that does not belong to the sequence instead of calling `skip()`, i.e. it behaves exactly like a build without the feature. Nothing is buffered, so `event_buffer_size` becomes irrelevant for that deserializer.

## Changes

- `src/de/mod.rs`: new `skip_unknown` field, public `overlapped_lists()` setter with a doctest, crate-private `can_skip()`; unit test `de::tests::skip::disabled` asserting the duplicate-field error and an empty skip buffer.
- `src/de/map.rs`: the two `skip()` call sites check `can_skip()`.
- `Cargo.toml`: feature docs mention the new switch.
- `Changelog.md`: entry under Unreleased / New Features.

## Verification

```
cargo test --all-features            # all suites pass, 0 failures
cargo fmt --check                    # clean
RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps   # clean
cargo clippy --all-features --lib    # 42 warnings, identical count before and after this change
```

RED→GREEN: with the `map.rs` change reverted, `de::tests::skip::disabled` fails with ``Expected `Err(Custom("duplicate field `item`"))`, but got `Ok(List { item: [(), (), ()] })` ``; with it, it passes.


src/de/mod.rs
```diff
@@ -2736,6 +2743,59 @@ where
...
+    /// match List::deserialize(&mut de) {
+    ///     Err(DeError::Custom(e)) => assert_eq!(e, "duplicate field `item`"),
+    ///     e => panic!("Expected `Err(Custom(\"duplicate field `item`\"))`, but got `{:?}`", e),
+    /// }
+    /// ```
+    #[cfg(feature = "overlapped-lists")]
+    pub fn overlapped_lists(&mut self, enable: bool) -> &mut Self {
+        self.skip_unknown = enable;
+        self
+    }
```
Mingun: It wouldn't work for your use-case if would be gated. If you does not enable `overlapped-lists` feature, you will not able to call this method.
```suggestion
    pub fn overlapped_lists(&mut self, enable: bool) -> &mut Self {
        #[cfg(feature = "overlapped-lists")]
        self.skip_unknown = enable;
        self
    }
```
Also, what if you change it in-flight? It seems that it should be configurable only at creation time.

src/de/map.rs
```diff
@@ -954,14 +954,20 @@ where
                 // If we see a tag that we not interested, skip it
                 #[cfg(feature = "overlapped-lists")]
                 DeEvent::Start(e) if !self.filter.is_suitable(e)? => {
-                    self.map.de.skip()?;
-                    continue;
+                    if self.map.de.can_skip() {
+                        self.map.de.skip()?;
+                        continue;
+                    }
+                    Ok(None)
```
Mingun: It would beter to have
```suggestion
                    #[cfg(feature = "overlapped-lists")]
                    if self.map.de.skip()? {
                        continue;
                    }
                    Ok(None)
```

src/de/mod.rs
```diff
@@ -2736,6 +2743,59 @@ where
...
+
+    #[cfg(feature = "overlapped-lists")]
+    #[inline]
+    pub(crate) const fn can_skip(&self) -> bool {
+        self.skip_unknown
+    }
+
```
Mingun: Inline this check into `skip()`
```suggestion
```

Mingun (changes requested):
I think it will be dangerous to change the mode during parse. When mode is disabled, you still need to reach checkpoint and drain already buffered events, otherwise you will get wrong results.

chiliec:
Thanks for the review, pushed `e56045e` with all three points:

- `overlapped_lists()` is no longer gated: it always exists and is a no-op without the feature, so a downstream crate can call it unconditionally. The doctest now runs under both feature configurations (the "supported by default" half is behind a hidden `cfg`).
- The check is inlined into `skip()`, which now returns `Result<bool>` (`Ok(false)` = disabled, nothing consumed) and `MapValueSeqAccess` does `if self.map.de.skip()? { continue; } Ok(None)`. `can_skip()` is gone; the existing `skip()` unit tests assert the `true` return.
- On changing the mode in flight: I agree it should be set up front, and I've documented that in the method docs ("call before deserialization starts"). I kept it as a `&mut self` setter rather than a constructor argument to match `event_buffer_size()` next to it, which has the same property. If you'd rather it be creation-time only, I can move it into the constructors instead — just say which shape you prefer.

`cargo test --all-features` and `--features serialize` pass, `cargo fmt --check` clean, no new clippy warnings.

